PR #479, opened 2024-01-30
use postgresql in kwild

jchappelow:
# PostgreSQL

This work is focused on the new `internal/sql/pg` package, and its use in the engine and other consumers like the account store.

There are also many changes to higher level designs like the atomic committer, new types of committables, and a drastically reduced dataset "registry" (`internal/sql/registry`), but these changes are more perfunctory and likely where we will need further thought and adjustment. The general architecture of kwild as it pertains to the ABCI app is generally the same.

**Review note:** This branch is also based on the [`pg-backend` branch](https://github.com/kwilteam/kwil-db/tree/pg-backend) (see [PR](https://github.com/kwilteam/kwil-db/pull/465)).  This PR begins with the `rm sqlite` commit.

## What works

- `kwild` runs
- unit tests
- acceptance test :party: (minus one lingering issue with in-line SQL expressions) :sad:

## What's broken or missing

- integration tests (actually, unknown)
- crash recovery: it's really easy to elicit "recovery not possible" errors, even through regular restarts depending on timing. This is sorta intentional since.
- many unit tests need `postgres` (the daemon process) stubbed out to become proper unit tests (there still can be complementary "integration" tests that are the ones currently tagged with `//go:build pglive`). This also prevents parallel tests.
- streamlined setup, including superuser-accessible helpers, docs, and workflows with containers
- ability to upgrade (possible at all without chain reset?)

I will revise this PR description with more information on the above shortly.

## What's changed

- New db backend: `internal/sql/pg` (see other PR)
- The dataset "registry" (`internal/sql/registry`) is updated with an approach based on postgresql "schema" (not a Kwil schema!) to put user datasets into self-contained namespaces within a single postgres database. This replaces the filesystem method used with sqlite, and makes their changes atomic.
- The dataset "registry" uses the `pg.DB`'s new commit ID mechanism to generate an ID returned by it's `Commit` method. (`DB` is a session-aware type built on top of a simpler connection pool)
- At top level, a single `pg.DB` instance is shared between each of: engine, acct store, validator store, and vote store.
- The engine continues to affect the app hash via the "registry", which is a type of committable integrated by the existing `MultiCommitter`.
- The account and validator stores use the same DB (pg database and sessioned write connection) as the engine. They still contribute to the app hash in their previous ways (streamed changes Register'd by the account store, and a commit ID func for the validator store's state). To do this within the existing `MultiCommitter` framework (we may want to change), this PR adds new committable types to provide commit IDs without actually beginning/ending a DB transaction. See the new `internal/sessions/{Aggregate,Dummy}` types, and comments in `cmd/kwild/server/build.go`.
- Voting store uses the sessioned DB connection, but does not affect the app hash (existing behavior).
- Event store uses a completely independent postgresql connection pool, but in the same database. It does not affect app hash. (existing behavior)
- Heaps of updates the the queries and code for the stores: accts, validators, events, and votes. This includes schemaifying the tables, and updating the queries for PostgreSQL.
- Engine internals. This is the bulk of it: positional query args, schema injection, in-line expressions, etc.
- `internal/engine/conv` is moved it `internal/conv` and used for more robust handling of values where type is or cannot be strict. [... more on this shortly]
- Docker compose files added and updated.
  -  `deployments/compose/postgres` contains a service for just postgres, which would be helpful if you don't want to install a system package for postgres, or you want an isolated instance
  - `deployments/compose/kwil/single` is a single kwild node container and a pg container configured to run together
  - `test/acceptance/docker-compose.yml` now includes a `pg` service, which the kwild service depends on and connects to (drives acceptances tests).
- ...

## What needs work

- Shutdown/crash robustness.

  This is hypothetically simpler on account of all user datasets now being atomically committed. There can be no partial updates to some datasets but not others. However, the existing mechanism for entering "recovery mode" has some flaws, so I disabled it so we can find a better solution for this.

- (existing) uncommitted vs committed queries from `(*engineAdapter).Call`, which supports RPC service as well as blockchain tx execution.

- The DBID -> postgresql schema convention is tentatively established in `internal/engine/types.DBIDSchema`. There might be a better place. I'm also cheating here and there with this "ds_" prefix known independently by other packages.  Not great.

- looking carefully at parser: keyword list, types, other places schema qualification is needed, etc.

- tests, in pg backend, updating registry tests, stubs vs. `postgres` live integration, etc.

- Reworking high level ABCI and multi-committer workings. We discussed having the commit ID, which becomes available *after* commit to lag (become the commit ID for the next block) so that the db's Commit can be in abci's Commit, and possibly to avoid a window for unrecoverable crashes

### types

We have types in kuneiform like `INT` and `BLOB` that are mapped to sql types in `internal/engine/metadata/ddl.columnTypeToSQLType`.

The only changes in this mapping are:

- `types.INT` (sqlite `int`) => `INT8` - this is an 8-*byte*/64-bit integer aka `BIGINT`.
- `types.BLOB` (sqlite `blob`) => `BYTEA`

The above is the Kwil schema => SQL mapping in the `table` definitions, I believe.

I don't know if these differences need to be accounted for elsewhere, or if we want to have actions with SQL queries use the new keywords or the old ones.

### keywords

There are several times more keywords: https://www.postgresql.org/docs/current/sql-keywords-appendix.html

I started updating in the sql grammar repo, but it got out of hand quickly and I stopped.

### misc

No `COLLATE NOCASE`... we can make a nocase collation though.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,7 +98,6 @@ jobs:
       # unit test
       - name: Run unit test
         run: |
-          psql -U kwild -p kwild -d kwil_test_db -h 127.0.0.1 -p 5432 -c 'CREATE PUBLICATION kwild_repl FOR ALL TABLES;'
           task test:unit
 
       # integration test
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -202,6 +202,20 @@ tasks:
     cmds:
       - go test ./acceptance -count=1 -v {{.CLI_ARGS}}
 
+  # NOTE: for now the unit tests require postgres running to support the tests
+  # with the "pglive" build tag. In CI, a postgres service is started by the
+  # github actions workflow. In dev machines, you can either use a system
+  # installation with a "kwild" (super)user and "kwil_test_db" database, or you
+  # can use the postgresql Docker compose at deployments/compose/postgres/docker-compose.yml
+  # or the custom Dockerfile at deployments/compose/postgres/Dockerfile.
+  #
+  # This should be TEMPORARY. The unit tests should work without postgres, or
+  # ideally any database engine. The various "stores" like account store can
+  # use a stubbed out Datastore interface to test the application logic, although
+  # the SQL queries in those packages should still be tested with a live DB.
+  # A mock interface should also be good for the engine. However, for the
+  # internal/sql/pg package, it is very tightly coupled to the pgx package and
+  # and the workings of an actual postgres host.
   test:unit:
     desc: Run unit tests
     cmds:
--- a/deployments/compose/kwil/init.sql
+++ b/deployments/compose/kwil/init.sql
@@ -1,5 +1,3 @@
 -- These commands are run with psql inside the container after postgres starts.
 CREATE USER kwild WITH PASSWORD 'kwild' SUPERUSER REPLICATION;
 CREATE DATABASE kwild OWNER kwild;
-\c kwild
-CREATE PUBLICATION kwild_repl FOR ALL TABLES;
--- a/deployments/compose/kwil/single/docker-compose.yml
+++ b/deployments/compose/kwil/single/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       -c max_prepared_transactions=2
     volumes: 
       - pgkwil:/var/lib/postgresql/data
-      - ./init.sql:/docker-entrypoint-initdb.d/create_tables.sql
+      - ./init.sql:/docker-entrypoint-initdb.d/create_user.sql
     networks:
       kwilnet0:
         ipv4_address: 172.5.100.3
--- a/deployments/compose/kwil/single/init.sql
+++ b/deployments/compose/kwil/single/init.sql
@@ -1,5 +1,3 @@
 -- These commands are run with psql inside the container after postgres starts.
 CREATE USER kwild WITH PASSWORD 'kwild' SUPERUSER REPLICATION;
 CREATE DATABASE kwild OWNER kwild;
-\c kwild
-CREATE PUBLICATION kwild_repl FOR ALL TABLES;
--- a/deployments/compose/postgres/Dockerfile
+++ b/deployments/compose/postgres/Dockerfile
@@ -1,4 +1,9 @@
 FROM postgres:16.1
 
+# Might do these envs, but for now leave it to compose:
+# ENV POSTGRES_USER kwild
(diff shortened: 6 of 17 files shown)


jchappelow:
New node
Wednesday Oct 9, 2024 ⋅ 2pm – 3pm
Central Time - Chicago

Join with Google Meet
https://meet.google.com/tvs-uxnn-csp?hs=224


	
Join by phone
(US) +1 575-448-4077
PIN: 807785260

More phone numbers
https://tel.meet/tvs-uxnn-csp?pin=7176825850793&hs=0


Organizer
Jonathan Chappelow
***@***.***

Guests
Jonathan Chappelow - organizer
***@***.***
View all guest info  
https://calendar.google.com/calendar/event?action=VIEW&eid=Nm9wNjZkcGhjZ3NtYWI5ajZ0aTY2YjlrNjVpMzhiOW82Z3JqNGI5cDZjcjNhY3BtNjhzamlwMWk2YyByZXBseStiYnBuYWx0ajU0YnhzdmNpNml3dWJlb2R6MmducmV2Ym5oaGgzZXR5eG1AcmVwbHkuZ2l0aHViLmNvbQ&tok=MTIjam9uQGt3aWwuY29tOWU3YmYwNTI2YjdmMjM3YjRiMmUxMzFiYTBmM2U0YThiODgxMGYwYQ&ctz=America%2FChicago&hl=en&es=0

Reply for ***@***.***  
and view more details  
https://calendar.google.com/calendar/event?action=VIEW&eid=Nm9wNjZkcGhjZ3NtYWI5ajZ0aTY2YjlrNjVpMzhiOW82Z3JqNGI5cDZjcjNhY3BtNjhzamlwMWk2YyByZXBseStiYnBuYWx0ajU0YnhzdmNpNml3dWJlb2R6MmducmV2Ym5oaGgzZXR5eG1AcmVwbHkuZ2l0aHViLmNvbQ&tok=MTIjam9uQGt3aWwuY29tOWU3YmYwNTI2YjdmMjM3YjRiMmUxMzFiYTBmM2U0YThiODgxMGYwYQ&ctz=America%2FChicago&hl=en&es=0
Your attendance is optional.

~~//~~
Invitation from Google Calendar: https://calendar.google.com/calendar/

You are receiving this email because you are an attendee on the event. To  
stop receiving future updates for this event, decline this event.

Forwarding this invitation could allow any recipient to send a response to  
the organizer, be added to the guest list, invite others regardless of  
their own invitation status, or modify your RSVP.

Learn more https://support.google.com/calendar/answer/37135#forwarding

PR #550, opened 2024-02-26
kwild,utils/url: allow using default port or host

jchappelow:
Rebirth of https://github.com/kwilteam/kwil-db/pull/527

This allows the listen addresses for app gRPC and HTTP to be partially specified, with only one of port or host, in which case it will use the default parsed from the full default listen address.

A similar change is applied for the cometbft listen addresses, but also applying the default scheme since cometbft requires a fully qualified URL with scheme unlike our apps listeners.

### DIFF
--- a/cmd/kwild/config/config.go
+++ b/cmd/kwild/config/config.go
@@ -7,7 +7,9 @@ import (
 	"encoding"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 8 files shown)


jchappelow:
Just added a bunch of test and fixed an existing bug with our `ParseURL` not working the same with IP addresses as it does with host names like `localhost`

jchappelow:
Note to self: `github.com/kwilteam/kwil-db/internal/services/grpc_gateway.NewGateway` interprets and `httpAddress string` set like `:8080` to mean `0.0.0.0:8080` (sensible).  We relied on that.  The default address for the http gateway is 0.0.0.0 now.  I was previously only in quick-start mode, but it makes sense always and it matches the gateway's behavior.

Yaiba (approved):
LGTM.

PR #554, opened 2024-02-26
Use CometBFT's RFC to fetch ValidatorSet at a given height

charithabandi:
If you request Comet for Validator set by explicitly specifying the height, I don't see any asynchrony. Pushing it here to see if we still see any issue in the CI.

This can also remove the node restart issue and always get the Current validators from the Comet and thereby ensures that a given node is in fact a valid validator.

Updating the comet to use v0.38.5 to resolve the security advisories related to Vote extensions

Currently when resolution entry is created it uses the VoteExpiry specified in the consensus compared to the Resolution Expiry Height with which the resolution type is registered. Fixing that. If a VoteID is received before VoteBody, the resolution is initially created to use VoteExpiry from genesis and upon receiving VoteBody it gets updated to use the corresponding resolution types expiration height.


brennanjl (approved):
lgtm

jchappelow (approved):

PR #556, opened 2024-02-27
updated authenticator interface

brennanjl:
While writing docs, Luke noticed our authenticator interface was unnecessarily deconstructed. We would never have these separate.


Yaiba (approved):
LGTM

PR #557, opened 2024-02-27
deployments,config: further extend postgres Docker, default to 127.0.0.1

jchappelow:
This updates our extended postgres Dockerfile so that it inserts an init script to create the "kwild" superuser and a "kwild" database owned by that user.  We can do other init tasks in these files, and can also include arbitrary .sh scripts to customize further.

This also changes the default `app.pg_db_host` to `127.0.0.1` instead of `/var/run/postgres` so that it can work with the Docker images and services without having to specify the hosts.

With this, I have pushed to [`jchappelow/kwil-postgres:latest` on Docker Hub](https://hub.docker.com/r/jchappelow/kwil-postgres/tags), and we can do:

```sh
docker run -v /home/jon/kwil-pg-test:/var/lib/postgresql/data -p 5432:5432 \
    -e "POSTGRES_PASSWORD=kwild" jchappelow/kwil-postgres
```

Qs:

- is this a resolution for the quickstart?
- Can we push this Docker image to the kwil account instead
- For multi-node, this doesn't quite help.  Each node needs it's own postgres, so they'd have to run multiple pg containers with different exposed ports, and specify them for each node
- What's the point of the multi-node quick start? I don't quite see the purpose. To just toy with the kwild RPC on a fake chain, the single node does that.  It's not a realistic multi-node scenario, so it doesn't demonstrate a proper way to deploy multiple nodes, but the Tutorial does.
- If multi node quickstart is valuable, maybe we can use the docker compose definition in `deployments/compose/kwil`, which starts two containers, one for kwild and one for postgres, and they automatically connect to each other.  EDIT: would still be tricky to get the different nodes talking.  nvm that

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,8 +15,6 @@ jobs:
       postgres:
         image: jchappelow/kwil-postgres:latest
         env:
-          POSTGRES_USER: kwild
-          POSTGRES_PASSWORD: kwild
           POSTGRES_PORT: 5432
           POSTGRES_DB: kwil_test_db
           POSTGRES_HOST_AUTH_METHOD: trust
@@ -203,4 +201,4 @@ jobs:
 
       - name: Prune Docker
         if: ${{ always() }}
-        run: docker rm $(docker ps -a -q) -f && docker network prune -f || true
+        run: docker rm $(docker ps -a -q) -f ; docker network prune -f ; docker volume prune -f || true
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,6 @@ jobs:
         image: jchappelow/kwil-postgres:latest
         env:
           POSTGRES_PORT: 5432
-          POSTGRES_DB: kwil_test_db
           POSTGRES_HOST_AUTH_METHOD: trust
         options: >-
           --health-cmd pg_isready
--- a/cmd/kwild/config/config.go
+++ b/cmd/kwild/config/config.go
@@ -441,7 +441,7 @@ func DefaultConfig() *KwildConfig {
 			HTTPListenAddress:  "localhost:8080",
 			AdminListenAddress: "unix:///tmp/kwil_admin.sock",
 			// SqliteFilePath:     DefaultSQLitePath,
-			DBHost: "/var/run/postgresql",
+			DBHost: "127.0.0.1",
 			DBPort: "5432", // ignored with unix socket, but applies if IP used for DBHost
 			DBUser: "kwild",
 			DBName: "kwild",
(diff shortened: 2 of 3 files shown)


Yaiba:
> is this a resolution for the quickstart?

I think so

> Can we push this Docker image to the kwil account instead

Yeah we should do it, otherwise it feels not from official author. @KwilLuke 

> What's the point of the multi-node quick start?

I kind agree with you on this

Yaiba (approved):
lgtm

KwilLuke:
> is this a resolution for the quickstart?

This looks amazing, Do we need to do something to make it cross platform? I am getting this error when trying to start on mac:

```
docker run -v /home/jon/kwil-pg-test:/var/lib/postgresql/data -p 5432:5432 --name kwil \
    -e "POSTGRES_PASSWORD=kwild" jchappelow/kwil-postgres
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
docker: Error response from daemon: Mounts denied:
The path /home/jon/kwil-pg-test is not shared from the host and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.
See https://docs.docker.com/desktop/mac for more info.
ERRO[0000] error waiting for container:
```

> Can we push this Docker image to the kwil account instead

Yes, let's push to official docker. @brennanjl has the push access.

> multi-node

The benefit of the multinode setup is it allows people to see data sync between nodes quickly and easily, and get the full picture of what Kwil does. E.g., execute an action on one node and query the result from another node. It allows us to prove that Kwil does what we say we do (without someone having to run on multiple machines) and establish a common mental model with our reader.

Would it be possible to just have the docker command just map to a different port, then tell each kwild where to connect to? E.g.

### Step 1: Start node 1

`docker run -v /home/jon/kwil-pg-test:/var/lib/postgresql/data -p 5432:5432 --name kwil \
    -e "POSTGRES_PASSWORD=kwild" jchappelow/kwil-postgres`

`kwild --app.pg-db-port 5432`

### Step 2: Start node 2

`docker run -v /home/jon/kwil-pg-test:/var/lib/postgresql/data -p 5433:5432 --name kwil \
    -e "POSTGRES_PASSWORD=kwild" jchappelow/kwil-postgres`

`kwild --app.pg-db-port 5433`

Would that work?

jchappelow:
> Would it be possible to just have the docker command just map to a different port, then tell each kwild where to connect to? E.g.
> ...
> Would that work?

Yup, for sure, just more flags and all, but it would do the trick for multi-node quickstart.

I think I see your vision for multi-node quick start now BTW.

> Do we need to do something to make it cross platform? 

I found the `scripts/publish/dockerhub` script that @brennanjl mentioned in the meeting, and I'll make a similar script for postgres.  After I test pushing to my jchappelow docker hub account, I'll get these images on the kwildb account.

PR #560, opened 2024-02-27
PG Sanity checks at Genesis 

charithabandi:
Voters/Validators is one good indicator of whether we are using stale pg instance or a new instance as we always initialize voters in the genesis. So using that to do sanity checks at the genesis.


jchappelow (approved):
LGTM, thanks.

PR #561, opened 2024-02-28
kwil-admin,config: disable genesis-hash command

jchappelow:
We don't support genesis-hash computation with postgres yet. Disable the command.

Also fix the reset command so it does not try to delete the old sqlite path, which no longer exists.

Finally, clean-up old SQLite mentions and example configs.

### DIFF
--- a/cmd/kwild/config/reset.go
+++ b/cmd/kwild/config/reset.go
@@ -21,7 +21,7 @@ func ResetChainState(rootDir string) error {
 }
 
 // ResetAll removes all data.
-func ResetAll(rootDir, sqlitePath, snapshotDir string) error {
+func ResetAll(rootDir, snapshotDir string) error {
 	// Remove CometBFT's stuff first.
 	chainRoot := filepath.Join(rootDir, ABCIDirName)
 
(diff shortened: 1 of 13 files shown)


charithabandi (approved):
LGTM!

cmd/kwild/config/test_data/config.toml
```diff
@@ -51,8 +49,20 @@ without_gas_costs = true
 # Toggle to disable nonces for transactions and queries
 without_nonces = false
 
-# KWILD Sqlite database file path
```
charithabandi, line 54: remove `without_gas_costs` and `without_nonces` from config.toml as they are genesis params

cmd/kwild/config/reset.go
```diff
@@ -77,11 +77,12 @@ func ResetAll(rootDir, sqlitePath, snapshotDir string) error {
 
 	// The user-configurable paths
 
-	if err := os.RemoveAll(rootify(sqlitePath, rootDir)); err == nil {
```
charithabandi, line 80: ApplicationDir doesn't exist anymore too

```appDir := filepath.Join(rootDir, ApplicationDirName)
	if err := os.RemoveAll(appDir); err == nil {
		fmt.Println("Removed all application", "dir", appDir)
	} else {
		fmt.Println("Error removing all application", "dir", appDir, "err", err)
	}```

cmd/kwild/config/test_data/config.toml
jchappelow: I think I didn't remove this before merge.  We can sneak that into a docs PR.

PR #562, opened 2024-02-28
build,scripts: use non-experimental golang docker image

jchappelow:
`golang:alpine` docker image is "highly experimental, and not officially supported by the Go project" according to https://hub.docker.com/_/golang and https://github.com/golang/go/issues/19938
As a test, this switches to the official and versioned golang docker image.  This is a much larger docker image, but it is only used for building the kwild image.  The size of our published kwild image is unchanged.


jchappelow:
Well this seems just fine actually.  Not just a test anymore.  Initial download of the `golang:1.22` builder can take a little time (10-20 second), but it's cached after that.  Again, no affect on `kwild` image size.

Yaiba (approved):
lgtm;

PR #563, opened 2024-02-29
fix integraion test locally parallel issue

Yaiba:
This fix the issue running integration test locally in parallel mode:
- increase default testcontainer wait timeout to 20s
- use correct test flag `parallel` to control the maximum number of tests to run simultaneously
- no parallel number control on CI, since CI machine is more powerful, and we expect test to finish asap

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
 
     services:
       postgres:
-        image: jchappelow/kwil-postgres:latest
+        image: kwildb/postgres:latest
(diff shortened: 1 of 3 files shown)


Yaiba:
Add more changes due to we just opensource the repo:
- switch back to github public runner
- remove using secret token
- remove kgw test
- add back cache for go/linter action
- use kwildb/postgres image

jchappelow (approved):
Unfortunate, but I think this is how it has to be with the unsandboxed self-hosted runner design

PR #565, opened 2024-02-29
Adds Version table to all the internal datastores of kwild

charithabandi:
Added a single row version table ["version",  versionNumber] to all the existing internal datastores of kwild.


internal/events/sql.go
```diff
@@ -4,6 +4,17 @@ const (
...
 
+	eventStoreVersion = 1
+
+	// versionTable is the SQL table definition for the version table.
+	// The version table is used to track the version of the database schema.
+	// It has a single row with a single column, version, which is an integer.
+	versionTable = `CREATE TABLE IF NOT EXISTS ` + SchemaName + `.version (
+		name TEXT NOT NULL PRIMARY KEY,
+		version INT NOT NULL,
+		UNIQUE(name)
```
jchappelow, line 15: `PRIMARY KEY` creates a unique index, no explicit constraint needed.

jchappelow (changes requested):
We chatted on slack, but I think the actual checks should be included with this change:  (a) we want to make sure the table schemas are workable, and (b) if someone has an upgraded node and they accidentally run an old binary, it should complain about an unknown/future version.

Regarding the possibility of a top level "kwild version", that would not be mutually exclusive with these checks, but I think what's important it the versions of the individual stores rather than the application.  Only reason I can think of: if a new store is added, potentially all the other stores' versions would be unchanged but if the user then accidentally ran old software, the presence of a foreign store would not bother it, but a top level app version could detect this.  However, potentially the `kwil_internal` schema version could simply be bumped with broader changes such as these.

Yaiba:
I'm sorry I cancelled all CI on this pr, since it needs to update workflow file. It's merged to main.

charithabandi:
> We chatted on slack, but I think the actual checks should be included with this change: (a) we want to make sure the table schemas are workable, and (b) if someone has an upgraded node and they accidentally run an old binary, it should complain about an unknown/future version.
> 
> Regarding the possibility of a top level "kwild version", that would not be mutually exclusive with these checks, but I think what's important it the versions of the individual stores rather than the application. Only reason I can think of: if a new store is added, potentially all the other stores' versions would be unchanged but if the user then accidentally ran old software, the presence of a foreign store would not bother it, but a top level app version could detect this. However, potentially the `kwil_internal` schema version could simply be bumped with broader changes such as these.

Yeah, I agree, still thinking on top level version, but I have changes ready for datastore specific checks.

jchappelow:
Hmm, so here's something that would make a table containing the current height useful: you nuke postgresql, but you have blockchain data.  You start up and kwild is happy, at least until it tries to do something with missing data.  If it checked a height stamp in postgres then it wouldn't happen.

charithabandi:
Still debating weather having a sanity check based on height is useful to prevent scenarios during restart when blockchain and app are at different states. But rest of the changes are up for review

internal/accounts/sql.go
```diff
@@ -13,29 +13,143 @@ const (
...
+	getVersion = `SELECT version FROM ` + schemaName + `.version WHERE name = 'version';`
+
+	acctsExists = `SELECT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_schema = ` + schemaName + ` AND table_name = 'accounts');`
 )
 
-func initTables(ctx context.Context, tx sql.DB) error {
+func initOrUpgradeTables(ctx context.Context, db sql.DB) error {
+	// Get the current version of the account store.
+	curVersion, versionErr := getVersionNumber(ctx, db)
+	if versionErr != nil {
```
brennanjl, line 48: Based on the implementation of `getVersionNumber`, it seems like we would want to fail in case of an error here? It would suggest that there is something critically wrong

internal/accounts/sql.go
```diff
@@ -13,29 +13,143 @@ const (
...
+	);`
+
 	sqlCreateAccount = `INSERT INTO ` + schemaName + `.accounts (identifier, balance, nonce) VALUES ($1, $2, $3)`
 
 	sqlUpdateAccount = `UPDATE ` + schemaName + `.accounts SET balance = $1, nonce = $2
 		WHERE identifier = $3`
 
 	sqlGetAccount = `SELECT balance, nonce FROM ` + schemaName + `.accounts WHERE identifier = $1`
+
+	upsertVersion = `INSERT INTO ` + schemaName + `.version (name, version) VALUES ('version', $1) ON CONFLICT(name) DO UPDATE SET version = $1;`
```
brennanjl, line 38: It seems like these `upsertVersion` and `accountsVersionTable` type things are pretty redundant across all packages (pretty much the same thing implemented 4 times). Could we have a `version` or `versioning` module, that has a constructor that takes the schema name? This will consolidate a lot of logic.

internal/versioning/versioning.go
```diff
@@ -0,0 +1,107 @@
+// package versioning provides standard schema versioning for Kwil databases.
```
jchappelow, line 1: Would prefer this in `internal/sql/versioning` since it only pertains to SQL

internal/txapp/txapp.go
```diff
@@ -77,13 +77,16 @@ func (r *TxApp) GenesisInit(ctx context.Context, validators []*types.Validator,
...
+	fmt.Println("height: ", height, "error: ", err)
 	if err != nil {
-		return err
-	}
-	if len(vals) > 0 {
-		return errors.New("genesisInit: cannot initialize genesis state: database has data from previous reset instance")
+		// if the height is not found, we will assume that the database is empty
+		r.log.Info("GenesisInit: starting with empty database")
+	} else if initialHeight-1 != height {
+		return fmt.Errorf("genesisInit: expected database to be at height %d, got %d", initialHeight-1, height)
```
jchappelow, line 89: Correct to error if `!=`, but if `==`, then we'd want this to be a no-op and return right away since we've done the init already.

However, I realized we're doing InitChain wrongish: https://github.com/cometbft/cometbft/issues/203

internal/txapp/txapp.go
```diff
@@ -63,6 +63,7 @@ type TxApp struct {
 
 	// transaction that exists between Begin and Commit
 	currentTx        sql.OuterTx
+	genesisTx        sql.OuterTx
```
jchappelow, line 66: Perhaps some docs on the field describing intended use pattern (where set, where checked, etc.) with a link to https://github.com/cometbft/cometbft/issues/203 or the relevant docs that describe the InitChain exception.

internal/events/sql.go
```diff
@@ -52,3 +61,21 @@ const (
...
+
+func initTables(ctx context.Context, db sql.DB) error {
+	tx, err := db.BeginTx(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	defer tx.Rollback(ctx)
+
+	// Create the events and kv table if it does not exist.
+	if _, err := tx.Execute(ctx, eventsTable); err != nil {
```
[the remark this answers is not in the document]
charithabandi: Its created in the versioning.go.

internal/accounts/accounts.go
```diff
@@ -18,9 +19,13 @@ func InitializeAccountStore(ctx context.Context, db sql.DB) error {
 	}
 	defer sp.Rollback(ctx)
 
-	err = initTables(ctx, sp)
+	upgradeFns := map[int64]versioning.UpgradeFunc{
+		-1: initTables,
```
brennanjl, line 23: So this is a weird quirk (and is why I previously had the `-1` field be optional in the upgrader).

What we have defined here is our "first version" for our account store; essentially, our v0. No matter what changes we have in the future, this will always be our v0. Therefore, I think it just makes more sense for the original initialization to be a v0.

internal/accounts/sql.go
```diff
@@ -27,15 +29,18 @@ const (
 	sqlGetAccount = `SELECT balance, nonce FROM ` + schemaName + `.accounts WHERE identifier = $1`
 )
 
-func initTables(ctx context.Context, tx sql.DB) error {
-	if _, err := tx.Execute(ctx, sqlCreateSchema); err != nil {
-		return err
+func initTables(ctx context.Context, db sql.DB) error {
+	tx, err := db.BeginTx(ctx)
```
brennanjl, line 33: Tx seems unnecessary

internal/engine/execution/global.go
```diff
@@ -36,6 +38,19 @@ type GlobalContext struct {
 
 var ErrDatasetNotFound = fmt.Errorf("dataset not found")
 
+func InitializeEngine(ctx context.Context, tx sql.DB) error {
+	upgradeFns := map[int64]versioning.UpgradeFunc{
+		-1: initTables,
```
brennanjl, line 43: Similar to my comments on account store, this seems like a version 0 to me. -1 is meant as a fast-path for new nodes that have never initialized, and need to initialize

internal/engine/execution/queries.go
```diff
@@ -28,6 +31,20 @@ var (
 	sqlDeleteKwilSchema  = fmt.Sprintf(`DELETE FROM %s.kwil_schemas WHERE dbid = $1;`, pg.InternalSchemaName)
 )
 
+func initTables(ctx context.Context, db sql.DB) error {
+	tx, err := db.BeginTx(ctx)
```
brennanjl, line 35: ts seems unnecessary

internal/events/sql.go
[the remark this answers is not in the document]
brennanjl: Seems like we don't need any of the sqlCreateSchema statements. They are left unused in the packages

internal/sql/versioning/versioning.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return err
+	}
+
+	if current > targetVersion {
+		return fmt.Errorf(`%w: current version: %d, target version: %d`, ErrTargetVersionTooLow, current, targetVersion)
+	}
+
+	// Fresh DB, Raw initialization to the latest version
+	if current == targetVersion {
```
brennanjl, line 73: this seems not true? If you restart a node, this will run. It's not too big a deal since we always do `CREATE tbl IF NOT EXISTS`, but it would be nice to detect for this (potentially by detecting if the schema already exists?).

cmd/kwild/server/build.go
```diff
@@ -65,17 +65,17 @@ func buildServer(d *coreDependencies, closers *closeFuncs) *Server {
...
 
 	// engine
-	e := buildEngine(d, closers, db)
+	e := buildEngine(d, db)
 
 	// account store
 	initAccountRepository(d, db)
 
 	// validator updater and store
 	//vstore := buildValidatorManager(d, db) // TODO: get rid of dis
```
jchappelow, line 74: Well, let's get rid of it then :)

internal/sql/versioning/versioning.go
```diff
@@ -0,0 +1,107 @@
...
+		return err
+	}
+
+	// Create the version table if it does not exist
+	if _, err = tx.Execute(ctx, fmt.Sprintf(sqlVersionTable, schema)); err != nil {
+		return err
+	}
+
+	// Ensure that the version exists. If it does not, insert it with the target version.
+	if _, err = tx.Execute(ctx, fmt.Sprintf(sqlEnsureVersionExists, schema), targetVersion); err != nil {
```
brennanjl, line 35: this is almost certainly a bug. If have a new node + new DB and are at version 1, we will essentially be setting the "current version" to version 1. This will then inform subsequent code that we are already at the correct version, when in fact we are at no version

internal/accounts/sql.go
jchappelow, line 33: `InitializeAccountStore` also wraps with a tx, then `Upgrade` wraps with a tx, and a third here in `initTables`.  I can't tell why more than one transaction is needed on that path.

internal/voting/sql.go
```diff
@@ -51,6 +51,16 @@ const (
 		id BYTEA PRIMARY KEY
 	);`
 
+	tableHeight = `CREATE TABLE IF NOT EXISTS ` + votingSchemaName + `.height (
+		name TEXT PRIMARY KEY, -- name is 'height'
+		height INT NOT NULL
+	);`
+
+	getHeight = `SELECT height FROM ` + votingSchemaName + `.height WHERE name = 'height';`
```
brennanjl, line 59: It seems weird to put the height in the voting schema, right? It is never actually used in relation to the rest of the votes / resolutions.

internal/sql/versioning/versioning.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	if current > targetVersion {
+		return fmt.Errorf(`%w: current version: %d, target version: %d`, ErrTargetVersionTooLow, current, targetVersion)
+	}
+
+	// Fresh DB, Raw initialization to the latest version
+	if current == targetVersion {
+		fn, ok := versions[-1]
+		if ok {
```
jchappelow, line 75: Should this be `if current == -1` instead of `if current == targetVersion`?  If you're at the target version, you should do nothing, not init tables, right?

internal/voting/voting.go
```diff
@@ -552,3 +568,42 @@ func intDivUpFraction(val, numerator, divisor int64) int64 {
...
+	height, ok := sql.Int64(res.Rows[0][0])
+	if !ok {
+		return 0, fmt.Errorf("invalid type for height (%T)", res.Rows[0][0])
+	}
+
+	return height, nil
+}
+
+func SetHeight(ctx context.Context, db sql.DB, height int64) error {
+	tx, err := db.BeginTx(ctx)
```
brennanjl, line 598: tx seems unnecessary?

internal/voting/voting.go
```diff
@@ -552,3 +568,42 @@ func intDivUpFraction(val, numerator, divisor int64) int64 {
 	tempNumerator.Add(tempNumerator, new(big.Int).Sub(divBig, big.NewInt(1)))
 	return new(big.Int).Div(tempNumerator, divBig).Int64()
 }
+
+func GetHeight(ctx context.Context, db sql.DB) (int64, error) {
+	tx, err := db.BeginTx(ctx)
+	if err != nil {
```
brennanjl, line 574: tx seems potentially useless. Postgres uses ReadCommitted by default I believe.

internal/voting/sql.go
jchappelow: In the sense that it's not referenced by other tables, yes.  But it is convenient from the perspective of building the app and not having to create another store `kwild_internal` was really supposed to be for use internal to the `pg` package, although engine uses it too now for schemas.

The most "blockchainy" store seemed to be `voting`, so it's not terrible.  What's a better spot?  Not opposed to moving it, just lacking a better suggestion.

cmd/kwild/server/build.go
brennanjl: haha this is my bad. I left this in from a previous PR

internal/sql/versioning/versioning.go
jchappelow: Hmm, I think I now see why you did it this way, but I don't think it's ideal.  If `ensureVersionTableExists` ends up making the version table, we still need to do init tables.  So because you insert latest version into the fresh version table, you catch it this way.  Alternatively:
- create version table with -1/0 whatever indicates "fresh" and uninitialized
- OR have `ensureVersionTableExists` return a bool indicating if it created a new version table, and then decide to call init tables based on that.  Then simplify the logic down here.

internal/voting/sql.go
brennanjl: I don't have other suggestions other than creating another schema, however it does seem like we are redundantly storing this height.

In ABCI, we store both the block height and app hash in a local badger DB. ABCI uses this to know where the app left off on startup (called on (*AbciApp).Info()`. It seems like we are potentially duplicating this functionality in part here.

IMO, it seems like TxApp should separately store both the height and apphash, and return this to info. CometBFT will actually catch these discrepancies automatically. Tracking chain data for synchronicity purposes seems like it should be a duty of TxApp, and so I feel like chain metadata should be a different store?

brennanjl: Sort've related, but in the future it may be worthwhile to group all stores in a sub package of internal. Logically grouping the functions in their own packages is nice, but it does sort've bloat the top-level `internal`

internal/events/sql.go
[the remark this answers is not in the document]
charithabandi: removing them

internal/voting/sql.go
brennanjl: I just spent a little bit of time testing this out. For the sake of this release, I think leaving height in vote store is ideal.

jchappelow: The motivation for also putting height (or block hash ideally) in postgres is to ensure we can't accidentally start up with an old or fresh postgres database and new blockchain data (in `root-dir`) without realizing it until much later, like when a tx fails.  https://github.com/kwilteam/kwil-db/pull/565#issuecomment-1973459908  So the redundancy is for consistency between the two.

Both the badger DB and cometbft's blockchain data is in `root-dir`, and we're trying to tackle cases where your `root-dir` has data that is different from postgres.

brennanjl: Comet relies on badger db to ensure consistency though. But now that we are using postgres, this seems incorrect; we want comet to be checking against postgres. We could check both, but IMO it seems like the badger db is almost useless

jchappelow: Oh, yeah, I agree with that.

test/integration/helper.go
```diff
@@ -52,12 +52,12 @@ var (
 var logWaitStrategies = map[string]string{
 	ExtContainer:  "listening on",
 	Ext3Container: "listening on",
-	"node0":       "comet node is now started",
-	"node1":       "comet node is now started",
-	"node2":       "comet node is now started",
-	"node3":       "comet node is now started",
-	"node4":       "comet node is now started",
-	"node5":       "comet node is now started",
+	"node0":       "finalized block",
```
jchappelow, line 55: I think this might be the key with `r.genesisTx` and the `InitChain` change.

internal/sql/versioning/versioning.go
```diff
@@ -0,0 +1,95 @@
...
+		return err
+	}
+
+	// Create the version table if it does not exist
+	if _, err = tx.Execute(ctx, fmt.Sprintf(sqlVersionTable, schema)); err != nil {
+		return err
+	}
+
+	// Ensure that the version exists. If it does not, insert it with the target version.
+	if _, err = tx.Execute(ctx, fmt.Sprintf(sqlEnsureVersionExists, schema), -1); err != nil {
```
brennanjl, line 35: nice this is a good way to solve this

brennanjl (approved):
lgtm

jchappelow (approved):

PR #569, opened 2024-03-01
[release-v0.7] deps: remove replace directives and tests in non-workspace mode

jchappelow:
In preparation for a release version with importable packages and go install'able applications, this removes the `replace` directives from the main module's go.mod.

This also removes the replace directives from the `test` module so that it tests what is actually built, not just what is in the git workspace.

With the replaces removed from the go.mod files, development on `release-v0.6` branch may require a go.work with `replace` directives in it.  However, if a build fails or CI fails without any replaces in go.mod or a go.work, that indicates we probably have to tag either the `core` or `parse` submodules and then update the main module's go.mod `require` on it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -158,7 +158,7 @@ jobs:
           testGroupID=$(id -g)
           cp test/acceptance/docker-compose.override.yml.example test/acceptance/docker-compose.override.yml
           sed -i "s/\${UID}:\${GID}/${testUserID}:${testGroupID}/g" test/acceptance/docker-compose.override.yml
-          KACT_LOG_LEVEL=warn task test:act:nb -- -parallel-mode -parallel 2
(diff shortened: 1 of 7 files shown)


jchappelow:
Before this PR:

```
$  go install -v github.com/kwilteam/kwil-db/cmd/kwild@release-v0.7
go: github.com/kwilteam/kwil-db/cmd/kwild@release-v0.7 (in github.com/kwilteam/kwil-db@v0.7.0-beta.0.20240301174656-62e40b344efa):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

After this PR merges, it should be possible to do the above (or import from say `github.com/kwilteam/kwil-db/extensions` in the main module)

jchappelow:
Now doing CI business without workspace mode. After this merges, it's good to go.

Yaiba:
LGTM.  Sorry didn't get the time to review this pr.

PR #576, opened 2024-03-06
removed circular abci import

brennanjl:
As discussed in Slack, I have removed the circular imports between abci and cometbft. We now simply will:
a. not propose a block if the local node is not a validator in the application
b. reject blocks proposed by nodes that the application does not recognize as validators

### DIFF
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -61,6 +61,8 @@ func NewAbciApp(cfg *AbciConfig, kv KVStore, snapshotter SnapshotModule,
 		consensusParams: consensusParams,
 
 		log: log,
+
+		validatorAddressToPubKey: make(map[string][]byte),
 	}
 
 	return app
@@ -109,7 +111,9 @@ type AbciApp struct {
 	consensusParams *txapp.ConsensusParams
 
 	broadcastFn EventBroadcaster
-	validatorFn ValidatorGetter
+
+	// validatorAddressToPubKey is a map of validator addresses to their public keys
+	validatorAddressToPubKey map[string][]byte
 }
 
 func (a *AbciApp) ChainID() string {
@@ -217,18 +221,13 @@ func (a *AbciApp) CheckTx(ctx context.Context, incoming *abciTypes.RequestCheckT
 // FinalizeBlock is on the consensus connection
 func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinalizeBlock) (*abciTypes.ResponseFinalizeBlock, error) {
 	fmt.Printf("\n\n")
-	logger := a.log.With(zap.String("stage", "ABCI FinalizeBlock"), zap.Int("height", int(req.Height)))
+	logger := a.log.With(zap.String("stage", "ABCI FinalizeBlock"), zap.Int64("height", req.Height))
 
 	res := &abciTypes.ResponseFinalizeBlock{}
 
 	err := a.txApp.Begin(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("begin tx commit failed: %w", err)
-	}
-
-	valAddrMap, err := a.getValidatorPubKeyAddrMap(ctx, &req.Height)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get validator address map: %w", err)
 	}
 
 	initialValidators, err := a.txApp.GetValidators(ctx)
@@ -246,7 +245,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		logger.Info("punish validator", zap.String("addr", addr))
 
 		// This is why we need the addr=>pubkey map. Why, comet, why?
-		pubkey, ok := valAddrMap[addr]
+		pubkey, ok := a.validatorAddressToPubKey[addr]
 		if !ok {
 			return nil, fmt.Errorf("unknown validator address %v", addr)
 		}
@@ -259,7 +258,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 	}
 
 	addr := proposerAddrToString(req.ProposerAddress)
-	proposerPubKey, ok := valAddrMap[addr]
+	proposerPubKey, ok := a.validatorAddressToPubKey[addr]
 	if !ok {
 		return nil, fmt.Errorf("failed to find proposer pubkey corresponding to address %v", addr)
 	}
@@ -318,6 +317,16 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 
 	res.ValidatorUpdates = make([]abciTypes.ValidatorUpdate, len(valUpdates))
 	for i, up := range valUpdates {
+		addr, err := pubkeyToAddr(up.PubKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert pubkey to address: %w", err)
+		}
+		if up.Power == 0 {
+			delete(a.validatorAddressToPubKey, addr)
+		} else {
+			a.validatorAddressToPubKey[addr] = up.PubKey // there may be new validators we need to add
+		}
+
 		res.ValidatorUpdates[i] = abciTypes.Ed25519ValidatorUpdate(up.PubKey, up.Power)
 	}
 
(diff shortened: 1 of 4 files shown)


internal/abci/abci.go
```diff
@@ -945,7 +944,3 @@ func (a *AbciApp) SetEventBroadcaster(fn EventBroadcaster) {
 }
 
 type ValidatorGetter func(ctx context.Context, height *int64) ([]*types.Validator, error)
```
jchappelow, line 946: ValidatorGetter also seems to be a relic now.

internal/abci/abci.go
```diff
@@ -815,11 +849,12 @@ func (a *AbciApp) ProcessProposal(ctx context.Context, req *abciTypes.RequestPro
...
 
-	addr := proposerAddrToString(req.ProposerAddress)
-	proposerPubKey, err := a.getValidatorPubKeyByAddr(ctx, addr, &req.Height)
-	if err != nil {
-		a.log.Warn("received block proposal from unknown validator", zap.String("addr", addr))
-		return &abciTypes.ResponseProcessProposal{Status: abciTypes.ResponseProcessProposal_REJECT}, nil
+	proposerPubKey, ok := a.validatorAddressToPubKey[proposerAddrToString(req.ProposerAddress)]
+	if !ok {
+		// there is an edge case where cometbft will allow a node to PrepareProposal
+		// even if it is not a validator, if it was a validator in the most recent block.
```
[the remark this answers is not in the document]
jchappelow: Regardless, while this may be a workaround for not tracking the delayed updates to the canonical validator set, if cometbft calls `ProcessProposal` with a given `ProposerAddress`, that means they are a validator.  CometBFT processes all updates given to it and from that manages the validator set (and all the precommit voting involves its canonical validator set), which includes `ProposerAddress`.

All that to say that we should soften the error since with the current solution it does not indicate a malfunction of the application. It is expected and working as intended, even if it's a workaround.

internal/abci/abci.go
```diff
@@ -700,12 +730,16 @@ func (a *AbciApp) PrepareProposal(ctx context.Context, req *abciTypes.RequestPre
...
-	addr, err := a.getValidatorPubKeyByAddr(ctx, proposerAddrToString(req.ProposerAddress), &req.Height)
-	if err != nil {
-		return nil, fmt.Errorf("failed to find proposer pubkey corresponding to %s", proposerAddrToString(req.ProposerAddress))
+	pubKey, ok := a.validatorAddressToPubKey[proposerAddrToString(req.ProposerAddress)]
+	if !ok {
+		// there is an edge case where cometbft will allow a node to PrepareProposal
+		// even if it is not a validator, if it was a validator in the most recent block
+		// we check for this here and do not propose if the node is not a validator
+		// we will also need to check for this in ProcessProposal, in case of byzantine behavior
+		return nil, fmt.Errorf("unknown proposer: failed to find proposer pubkey corresponding to address %v", req.ProposerAddress)
```
jchappelow, line 739: Alright, I suppose, but since this is `PrepareProposal`, the `ProposerAddress` is *us*, the local validator, so we could surely retrieve the pubkey somehow.

brennanjl: Right. I'll change the error message, but you would agree this is the correct logic, correct?

internal/abci/abci.go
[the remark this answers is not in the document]
brennanjl: > Are these comments alluding to the h vs h+2 discrepancy, or some other edge where a validator updated to power 0 at height h can actually be asked to prepare a proposal for h+2 if it was (acceptably) the proposer at h+1? So this error reflects an unfixed big in some sense.

Yes

brennanjl:
After testing abci, it seems like an error returned will signal to the node a consensus failure. For `ProcessProposal`, we now return reject. I'm not sure what to do for `PrepareProposal`, since returning nil for the proposal causes a nil pointer dereference. For now, I just return an empty proposal, expecting it to be rejected.

jchappelow:
> After testing abci, it seems like an error returned will signal to the node a consensus failure.

For both prepare and process?

> For `ProcessProposal`, we now return reject. I'm not sure what to do for `PrepareProposal`, since returning nil for the proposal causes a nil pointer dereference. For now, I just return an empty proposal, expecting it to be rejected.

With prepare, we know our own validator pubkey, so we wouldn't ever have to return an error there on account of not recognizing the proposer address?  Are we trying to avoid abci being aware of the node key?

brennanjl:
> For both prepare and process?

Yes.

> With prepare, we know our own validator pubkey, so we wouldn't ever have to return an error there on account of not recognizing the proposer address? Are we trying to avoid abci being aware of the node key?

No, I'm more thinking that an honest node should not propose if it knows it is not a validator and will get rejected. It seems like we _have_ to propose, or else the local node will fail. I have instead made it propose an empty block if it expects to be rejected.

jchappelow:
> No, I'm more thinking that an honest node should not propose if it knows it is not a validator and will get rejected. It seems like we _have_ to propose, or else the local node will fail.

Yeah, I'm OK with these workarounds, I think, but this is exactly why I say that we *are* (still) a validator and that our application tracking is what is inaccurate.  CometBFT says we are, so we are.  I think we should consider a more proper fix at some point.

> I have instead made it propose an empty block if it expects to be rejected.

OK.  Empty block isn't wrong, so why bother with contents if it will be **reject**ed according to our current ProcessProposal code.  Makes sense.

Ultimately it seems like the right solution is to just not `delete` from the map, but move into another map that includes the height at which a key should be deleted.  And for startup to restore both maps accurately, this would seem to require a new table in voting store.

@charithabandi Do you have any perspective on this that might help clear up the "right" way to handle this?

brennanjl:
> Yeah, I'm OK with these workarounds, I think, but this is exactly why I say that we are (still) a validator and that our application tracking is what is inaccurate. CometBFT says we are, so we are. I think we should consider a more proper fix at some point.

I would still disagree with this. Our application has defined its own concept of a validator, how it has power, and some things it can do. IMO we are simply telling Comet how we want it to perform consensus.

jchappelow:
We can have our own concept of a validator for things like voting on events that are unknown to cometbft, but a different set of validators will be participating in Tendermint consensus to approve blocks and we can't change the rules for how that set changes.  The expectation is that the application should not maintaining a separate and different set of "validators", particularly if it's to be used to [reject block proposals](https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#processproposal):

> Moreover, application implementors SHOULD always set ResponseProcessProposal.status to ACCEPT, unless they really know what the potential liveness implications of returning REJECT are.

We can go through multiple rounds and possibly worse depending on the validator selection algorithm and number of validators.

Docs for `ProcessProposal` also say:

> `proposer_address` 	bytes 	[Address](https://docs.cometbft.com/v0.38/spec/core/data_structures#address) of the validator that created the proposal.

So we can say that they are not a validator, but cometbft is saying they are, and that we shouldn't reject the proposal if we disagree with that.

If we go this route, we'll need to test a lot of scenarios to be sure we don't halt the chain unintentionally.

internal/abci/abci.go
```diff
@@ -318,6 +317,16 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 
 	res.ValidatorUpdates = make([]abciTypes.ValidatorUpdate, len(valUpdates))
 	for i, up := range valUpdates {
+		addr, err := pubkeyToAddr(up.PubKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert pubkey to address: %w", err)
+		}
+		if up.Power == 0 {
+			delete(a.validatorAddressToPubKey, addr)
```
charithabandi, line 325: Okay we are deleting this from the map. So if a validator leaves and it becomes the proposer in the next block, we would just reject that proposed block.

internal/abci/abci.go
```diff
@@ -700,12 +730,16 @@ func (a *AbciApp) PrepareProposal(ctx context.Context, req *abciTypes.RequestPre
...
-	addr, err := a.getValidatorPubKeyByAddr(ctx, proposerAddrToString(req.ProposerAddress), &req.Height)
-	if err != nil {
-		return nil, fmt.Errorf("failed to find proposer pubkey corresponding to %s", proposerAddrToString(req.ProposerAddress))
+	pubKey, ok := a.validatorAddressToPubKey[proposerAddrToString(req.ProposerAddress)]
+	if !ok {
+		// there is an edge case where cometbft will allow a node to PrepareProposal
+		// even if it is not a validator, if it was a validator in the most recent block
+		// we check for this here and simply propose an empty block, since it will be rejected
+		a.log.Warn("local node was made proposer, but is no longer a validator")
+		return &abciTypes.ResponsePrepareProposal{}, nil
```
charithabandi, line 739: coool 👍🏻

charithabandi (approved):
LGTM. Sending REJECT if the proposer is not a validator is not something that would cause liveness issue, so should be okay, as eventually the validator set used by comet consensus would change.

jchappelow (approved):
Let's go ahead with this.

I will just do some tests to understand if the deterministic validator selection algorithm can get into a situation where each round picks the same removed validator forever or a long time.

PR #578, opened 2024-03-06
accounts: mock DB

jchappelow:
First of several updates so unit tests cover more code without the `pglive` tag.  This tag still exists, but for separate tests to ensure that the queries written are actually compatible and correct with postgres.


brennanjl (approved):

Yaiba:
Can you add `-s`  to `grep` in 'Print error log' step in CI.    Seems it errors out itself..

PR #579, opened 2024-03-06
Enfore MaxBlockSize limits

charithabandi:


internal/txapp/txapp.go
```diff
@@ -517,6 +537,13 @@ func (r *TxApp) ProposerTxs(ctx context.Context, txNonce uint64) ([]*transaction
 			continue
 		}
 
+		// MaxTxBytes restrictions per block are enforced here
+		// From observations, the rlp encoded events size is almost always smaller than the event
```
jchappelow, line 541: What is meant by "the rlp encoded events size is almost always smaller than the event"?  I get the RLP part, but what's size of just "the event" mean?

internal/abci/abci.go
```diff
@@ -717,6 +720,18 @@ func (a *AbciApp) PrepareProposal(ctx context.Context, req *abciTypes.RequestPre
...
 	}
 
+	var finalTxs [][]byte
+	maxTxSize := req.MaxTxBytes
+
+	// Txs order: MempoolProposerTxns, ProposerInjectedTxns, MempoolTxns
+	for _, tx := range propTxs {
+		txSize := int64(len(tx))
+		if maxTxSize < txSize {
+			return &abciTypes.ResponsePrepareProposal{Txs: finalTxs}, nil
```
jchappelow, line 730: Without the txns from `propTxs` that would fit?

internal/txapp/txapp.go
charithabandi: of "the event"

internal/txapp/txapp.go
```diff
@@ -34,10 +44,11 @@ func NewTxApp(db DB, engine common.Engine,
...
 		},
-		signer:           signer,
-		chainID:          chainID,
-		GasEnabled:       GasEnabled,
-		extensionConfigs: extensionConfigs,
+		signer:            signer,
+		chainID:           chainID,
+		GasEnabled:        GasEnabled,
+		extensionConfigs:  extensionConfigs,
+		emptyVoteBodyTxSz: voteBodyTxSz,
```
brennanjl, line 51: seems a bit unnecessary to do this here, since it is only used once. Maybe just calculate the emptyVoteBodyTxSize when you need it?

Also, a minor nit: I notice you do `sz` a lot for `size` (e.g. `emptyVoteBodyTxSz`, `eventSz`, etc.). I'd rather we just have the full word here, and have slightly longer variable names (especially if we are only saving two characters). Obviously there are times where abbreviation is good, but for something like this it seems overkill and could cause unnecessary confusion

brennanjl (commented):
I had done a partial review, but I will circle back since you said you have other changes

internal/txapp/txapp.go
jchappelow: Since voteBodyTxSz is constant for a given chain ID so it would be a shame to compute it repeatedly since it involves a bunch of rlp enoding.  It would be computed each call to `ProposerTxs` I think.

internal/txapp/txapp.go
jchappelow: > of "the event"

Sorry, I still don't understand what it's saying

internal/txapp/txapp.go
brennanjl: Ah that's a good point. Ok

internal/abci/abci.go
```diff
@@ -685,14 +681,71 @@ func (a *AbciApp) prepareMempoolTxns(txs [][]byte, maxBytes int, log *log.Logger
...
+			otherTxns[string(tx.Sender)] = append(otherTxns[string(tx.Sender)], tx)
+		}
 		nonces = append(nonces, tx.Body.Nonce)
 		i++
 	}
 
-	return finalTxns, proposerNonce
+	// truncate based on our max block size since we'll have to set
+	// ConsensusParams.Block.MaxBytes to -1 so that we get ALL transactions even
+	// if it goes beyond max_tx_bytes.  See:
```
jchappelow, line 700: We haven't yet set it to -1 AFAIK,

internal/abci/abci.go
```diff
@@ -685,14 +681,71 @@ func (a *AbciApp) prepareMempoolTxns(txs [][]byte, maxBytes int, log *log.Logger
...
+			propTxs = append(propTxs, tx)
+		} else {
+			// Append the transaction to the final list.
+			if _, ok := otherTxns[string(tx.Sender)]; !ok {
+				otherTxns[string(tx.Sender)] = make([]*indexedTxn, 0)
+				senders = append(senders, string(tx.Sender))
 
-		finalTxns = append(finalTxns, txs[tx.is])
+			}
+			otherTxns[string(tx.Sender)] = append(otherTxns[string(tx.Sender)], tx)
```
jchappelow, line 692: Every single `string(tx.Sender)` makes it copy the bytes.  We can do it once when we enter the `else`

internal/abci/abci.go
```diff
@@ -685,14 +681,71 @@ func (a *AbciApp) prepareMempoolTxns(txs [][]byte, maxBytes int, log *log.Logger
 
 		if bytes.Equal(tx.Sender, proposerAddr) {
 			proposerNonce = tx.Body.Nonce + 1
-		}
+			propTxs = append(propTxs, tx)
+		} else {
+			// Append the transaction to the final list.
+			if _, ok := otherTxns[string(tx.Sender)]; !ok {
+				otherTxns[string(tx.Sender)] = make([]*indexedTxn, 0)
```
jchappelow, line 688: This line is unneeded.

```
			sender := string(tx.Sender)
			senderTxns, ok := otherTxns[sender]
			if !ok {
				senders = append(senders, sender)
			}
			otherTxns[sender] = append(senderTxns, tx)
```

internal/abci/abci.go
```diff
@@ -685,14 +681,71 @@ func (a *AbciApp) prepareMempoolTxns(txs [][]byte, maxBytes int, log *log.Logger
...
+	for _, tx := range proposerTxs {
+		txSize := int64(len(tx))
+		if maxTxBytes < txSize {
+			break
+		}
+		maxTxBytes -= txSize
+		finalTxs = append(finalTxs, tx)
+	}
+
+	for _, sender := range senders {
```
jchappelow, line 733: A comment on the `senders` loop, and above when building the slice would be helpful.

jchappelow: It's not obvious why the mempool transactions should be grouped by sender, or if that's really desirable.

For instance, we could do without `senders` slices, and simplify the bottom mempool loop:

```
	for sender, senderTxs := range senders {
		for _, tx := range senderTxs {
```

That is still grouped, but random order of the groups.

internal/abci/abci.go
```diff
@@ -685,14 +681,71 @@ func (a *AbciApp) prepareMempoolTxns(txs [][]byte, maxBytes int, log *log.Logger
...
+		senderTxs, ok := otherTxns[sender]
+		if !ok {
+			continue
+		}
+
+		for _, tx := range senderTxs {
+			txSize := int64(len(txs[tx.is]))
+			if maxTxBytes < txSize {
+				break
```
jchappelow, line 742: `break senderLoop`, with:

```go
senderLoop:
    for _, sender := range senders {
```

internal/abci/abci.go
```diff
@@ -685,14 +681,71 @@ func (a *AbciApp) prepareMempoolTxns(txs [][]byte, maxBytes int, log *log.Logger
...
+			break
+		}
+		maxTxBytes -= txSize
+		finalTxs = append(finalTxs, tx)
+	}
+
+	for _, sender := range senders {
+		senderTxs, ok := otherTxns[sender]
+		if !ok {
+			continue
```
jchappelow, line 736: This is an impossible condition, right?

internal/abci/abci.go
```diff
@@ -623,7 +623,7 @@ type indexedTxn struct {
 //
 // NOTE: This is a plain function instead of an AbciApplication method so that
 // it may be directly tested.
-func (a *AbciApp) prepareMempoolTxns(txs [][]byte, maxBytes int, log *log.Logger, proposerAddr []byte) ([][]byte, uint64) {
+func (a *AbciApp) prepareMempoolTxns(txs [][]byte, log *log.Logger, maxTxBytes int64, proposerAddr []byte) [][]byte {
```
jchappelow, line 626: Should update docs for this method, and probably rename it from `prepareMempoolTxns` since it now injects the proposer transactions too, which used to be done in `PrepareProposal`

internal/abci/abci.go
charithabandi: Yeah, it was in TODO, I removed it by mistake.

internal/txapp/txapp.go
```diff
@@ -492,30 +492,44 @@ func (r *TxApp) AccountInfo(ctx context.Context, acctID []byte, getUncommitted b
...
+
 	readTx, err := r.Database.BeginReadTx(ctx)
 	if err != nil {
 		return nil, err
 	}
 	defer readTx.Rollback(ctx) // always rollback read tx
 
 	// Consider empty vote body tx and the rlp encoding overhead size(safety buffer)
-	maxTxSize -= r.emptyVoteBodyTxSz - rlpEncodingOverheadSize
+	maxTxSize -= r.emptyVoteBodyTxSize - rlpEncodingOverheadSize
```
jchappelow, line 524: Just to double check, do we want `maxTxSize -= (r.emptyVoteBodyTxSz - rlpEncodingOverheadSize)` (as you have it, equivalently) or `maxTxSize -= r.emptyVoteBodyTxSz + rlpEncodingOverheadSize` (subtract both from `maxTxSize`)?

internal/txapp/txapp.go
```diff
@@ -538,9 +552,8 @@ func (r *TxApp) ProposerTxs(ctx context.Context, txNonce uint64, maxTxSize int64
...
 
 		// MaxTxBytes restrictions per block are enforced here
-		// From observations, the rlp encoded events size is almost always smaller than the event
-		// so use eventsize for enfrocing maxTxSize and fill the rest of the blocksize with the mempool txns.
-		evtSz := eventSize(event)
+		// As the rlp encoded size is almost always smaller, enforcing it on the unencoded data for ease of use
+		evtSz := int64(len(event.Type)) + int64(len(event.Body))
 		if evtSz > maxTxSize {
 			break
 		}
```
jchappelow, line 559: Do we want to `maxTxSize -= evtSz` whenever we `finalEvents = append(finalEvents, event)`?

internal/txapp/txapp.go
charithabandi: Oops, I wanted the second one.

internal/abci/abci.go
charithabandi: I think I added it, so that its testable. But probably we don't have to group them by sender

internal/txapp/txapp.go
```diff
@@ -667,3 +712,18 @@ func logErr(l log.Logger, err error) {
...
+	}, chainID, 0)
+	if err != nil {
+		return 0, err
+	}
+	sz, err := tx.MarshalBinary()
+	if err != nil {
+		return 0, err
+	}
+	return int64(len(sz)), nil
+}
```
jchappelow, line 729: Let's be more conservative/realistic with a non-zero nonce (zero is encoded shorter by RLP), and a non-zero fee (larger big.Int encoding):

```go
func computeEmptyVoteBodyTxSize(chainID string) (int64, error) {
	// Create a transaction with an empty payload to measure the fixed size without the payload.
	tx, err := transactions.CreateTransaction(&transactions.ValidatorVoteBodies{
		Events: []*types.VotableEvent{},
	}, chainID, 1<<63)
	if err != nil {
		return 0, err
	}
	tx.Body.Fee, _ = big.NewInt(0).SetString("987654000000000000000000000000000", 10)
	sz, err := tx.MarshalBinary()
	if err != nil {
		return 0, err
	}
	return int64(len(sz)), nil
}
```

```go
func Test_computeEmptyVoteBodyTxSize(t *testing.T) {
	tests := []struct {
		name    string
		chainID string
		want    int64
		wantErr bool
	}{
		{"asdf", "asdf", 72, false},
		{"abcdefghijklmnop", "abcdefghijklmnop", 84, false},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got, err := computeEmptyVoteBodyTxSize(tt.chainID)
			if (err != nil) != tt.wantErr {
				t.Errorf("computeEmptyVoteBodyTxSize() error = %v, wantErr %v", err, tt.wantErr)
				return
			}
			if got != tt.want {
				t.Errorf("computeEmptyVoteBodyTxSize() = %v, want %v", got, tt.want)
			}
		})
	}
}
```

charithabandi: Makes sense.

internal/txapp/txapp.go
charithabandi: Yeah, that's probably because there is only 1 transaction that's introduced by the proposer. If we have more than one, we could account for these overhead multiple times.

internal/abci/abci.go
charithabandi: Updated, so that the order is still fifo, except for proposer txns.

internal/abci/abci.go
```diff
@@ -685,14 +680,73 @@ func (a *AbciApp) prepareMempoolTxns(txs [][]byte, maxBytes int, log *log.Logger
...
+		maxTxBytes -= txSize
+		finalTxs = append(finalTxs, tx)
+	}
+
+	// senders map is used to keep track of the senders that
+	// trasactions should be ignored from due to maxTxBytes limits
+	senders := make(map[string]bool)
+	for _, tx := range otherTxns {
+		sender := string(tx.Sender)
+		// If we have already added a transaction from this sender, skip it.
```
jchappelow, line 734: To make sure I understand the intent, the comment means "If a previous transaction from this senders would have pushed over the bytes limit for the block, we skip all subsequent transactions from the sender because nonces need to be sequential. Keep checking transactions for other senders that may be smaller and fit in the remaining space."

charithabandi: Yes, correct.

jchappelow (approved):
Looks fine but have not tested.  @brennanjl the `spam` oracle extension sounds kinda useful.  Can you paste it somewhere?

brennanjl:
> Looks fine but have not tested. @brennanjl the spam oracle extension sounds kinda useful. Can you paste it somewhere?

I simply deleted it but can re-implement it in about 2 minutes.

jchappelow:
Oh, hah, no worries.  It's just a loop writing junk to the event store I imagine.

brennanjl:
```go
package spam

import (
	"context"
	"fmt"

	"github.com/kwilteam/kwil-db/common"
	"github.com/kwilteam/kwil-db/extensions/oracles"
)

func init() {
	err := oracles.RegisterOracle("spam", func(ctx context.Context, service *common.Service, eventstore oracles.EventStore) {
		count := 0
		for {
			select {
			case <-ctx.Done():
				return
			default:
				err := eventstore.Broadcast(ctx, "spam", []byte("spam"+fmt.Sprint(count)))
				if err != nil {
					panic(err)
				}

				count++

				if count == 10000 {
					return
				}
			}
		}
	})
	if err != nil {
		panic(err)
	}
}
```

brennanjl:
You may also need to register a "spam" resolution

brennanjl:
```go
	err = resolutions.RegisterResolution("spam", resolutions.ResolutionConfig{
		ResolveFunc: func(ctx context.Context, app *common.App, resolution *resolutions.Resolution) error {
			app.Service.Logger.Info("spammed resolution")
			return nil
		},
	})
```

brennanjl (approved):

PR #580, opened 2024-03-07
CI: add print error step

Yaiba:
Since we switch back to github's public runner, we no longer have access to the logs for failed test case.
This pr tries to print error if a step fails, by just `grep` the log files


Yaiba:
hmm, I should only grep log file

PR #581, opened 2024-03-07
Spam events

brennanjl:
This PR fixes a few issues with oracles:
- If an oracle voted on a non-existent resolution type, it would go through consensus before failing. The node will now immediately detect if the resolution type is unknown.
- During broadcast of vote IDs, there were a few major inefficiencies that the node made that required more calls to postgres than necessary. These have no been solved. They can be further optimized, but the bulk of it is fixed for now

This should go in after https://github.com/kwilteam/kwil-db/pull/579


extensions/register.go
```diff
@@ -4,4 +4,7 @@ package extensions
 // this directory has 0 dependencies, so it can import anything
 // it is imported by cmd/kwild/main.go, so any other files in this directory will be compiled
 
-import _ "github.com/kwilteam/kwil-db/extensions/oracles/eth_deposits"
+import (
+	_ "github.com/kwilteam/kwil-db/extensions/oracles/eth_deposits"
+	_ "github.com/kwilteam/kwil-db/extensions/oracles/spam"
```
jchappelow, line 9: Did you need to `git add` a new `spam` package for testing?

brennanjl: oops. I actually meant to remove the spam package. I think we need a more formal way of testing spam, but for now I was just testing manually with a basic for loop

brennanjl: While you were gone, we discussed a testing framework for compiled extensions. This would apply to oracles as well. I don't really want to implement this before the release

internal/voting/sql.go
```diff
@@ -105,6 +105,12 @@ const (
 
 	// containsBody checks if a resolution has a body
 	containsBody = `SELECT body is not null FROM ` + votingSchemaName + `.resolutions WHERE id = $1;`
+	// containsBodyArray checks if a set of resolutions have bodies.
+	// it will return as many rows as the input array has elements, with a boolean for each.
+	containsBodyArray = `WITH input AS (SELECT unnest($1::BYTEA[]) AS id)
```
jchappelow, line 110: This is good for the contract of `ResolutionsContainsBody`, but you're ultimately after `doesNotHaveBody` in `ProposerTxs`, so a left join (and the subquery with unnest) looks to be too much for the job (we want a set of ids that have body, a subset of the inputs with no real need for the false entries since that is implied by their absence).

`SELECT id FROM kwild_voting.resolutions WHERE id =ANY($1)`

and `ResolutionsContainsBody` would just be doing `idWithBody := row[0].([]byte)` in the rows loop, and the app would do the matching to create the `[]bool`.

What you have works, but I wanted to point out the simpler SQL approach in case it wasn't apparent.  It's more testable and less error prone, less at the mercy of quirks and guarantees of postgres in handling of nulls, defaults, and ordering with array-to-rows etc.

internal/voting/sql.go
```diff
@@ -125,6 +131,13 @@ const (
 	// alreadyProcessed checks if a resolution has already been processed
 	alreadyProcessed = `SELECT id FROM ` + votingSchemaName + `.processed WHERE id = $1;`
 
+	// manyProcessed checks if many resolutions have already been processed
+	// it will return a boolean for each resolution in the input array
+	manyProcessed = `WITH input AS (SELECT unnest($1::BYTEA[]) AS id)
```
jchappelow, line 136: Same basic comment as the other query (can return a subset of processed ids rather than a bool set of same length and order), but nbd.

internal/voting/vote_test.go
```diff
@@ -1,9 +1,10 @@
-//go:build pglive
+// go:build pglive
```
jchappelow, line 1: Can revert.  BTW, you and @charithabandi do this frequently, and I'm not clear on why.  You can set VSCode test tags if that's the reason.

```json
    "go.testTags": "ext_test,pglive",
    "go.buildTags": "ext_test,pglive",
```

internal/events/sql.go
```diff
@@ -30,6 +30,7 @@ const (
 	insertEventIdempotent  = `INSERT INTO ` + SchemaName + `.events (id, data, event_type) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING;`
 	deleteEvent            = `DELETE FROM ` + SchemaName + `.events WHERE id = $1;`
 	getEvents              = `SELECT data, event_type FROM ` + SchemaName + `.events;`
+	getEventsLimit         = `SELECT data, event_type FROM ` + SchemaName + `.events LIMIT $1;`
```
charithabandi, line 33: Is this used somewhere? probably include the #events in the GetEvents(ctx, db, #events)

internal/events/events.go
```diff
@@ -108,6 +114,8 @@ func (e *EventStore) Store(ctx context.Context, data []byte, eventType string) e
 }
 
 // GetEvents gets all events in the event store.
+// It can be given a maximum number of events to retrieve.
+// If 0 is given, it will retrieve all events.
```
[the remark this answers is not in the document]
brennanjl: Ah sorry I added and removed that

internal/voting/sql.go
brennanjl: Ah that's a good point. I think we maybe just leave this for now, simply because the first thing I am going to do after this release is address the unnecessarily inefficient access patterns around events and voting. Is that ok with you?

internal/voting/vote_test.go
brennanjl: Woah mind blown

internal/events/sql.go
brennanjl: Ah I made something for it but got rid of it

brennanjl:
> Ci tests are failing and there are floods of these errors. which I don't understand. This should generally occur if multiple nodes issue a voteBodyTransaction on same event. But we prevent that in ProposerTxs.

Hmm I'm guessing that nodes aren't properly noting what has been given a body. Let me check

internal/voting/sql.go
brennanjl: Never mind, I think failing CI is actually related to the ordering of the booleans so I am going to address it now

internal/txapp/txapp.go
```diff
@@ -560,34 +560,41 @@ func (r *TxApp) ProposerTxs(ctx context.Context, txNonce uint64, maxTxsSize int6
...
 
-		// MaxTxBytes restrictions per block are enforced here
-		// As the rlp encoded size is almost always smaller, enforcing it on the unencoded data for ease of use
 		evtSz := int64(len(event.Type)) + int64(len(event.Body))
 		if evtSz > maxTxsSize {
 			break
 		}
-		finalEvents = append(finalEvents, event)
 		maxTxsSize -= evtSz
+		maxTxsSize -= rlpEncodingOverheadSize
```
charithabandi, line 596: If we are doing this for every event, we should probably reduce the OverheadSize to something proportional to size. Because right now its defined to be quite high upto 1000.

internal/voting/vote_test.go
charithabandi: This is still not updated. remove the extra space?

internal/txapp/txapp.go
brennanjl: Hmm I see. I think we should do it per event, since it will be constant. Let me try smthn

internal/voting/vote_test.go
brennanjl: Lol I fixed and unfixed sorry

brennanjl:
I screwed up the commit but it will get squashed anyways. It should now more accurately determine the additional size of including rlp (we might want to add a safety buffer as well)

charithabandi (approved):
Lgtm!

PR #583, opened 2024-03-08
config,nodecfg: write out config file in autogen mode

jchappelow:
When doing `--autogen`, write out the `config.toml` from the generated configuration.

Also, in autogen mode, change the default `HTTPListenAddr` to all interfaces rather than localhost.  I've kept the default in  other cases at `localhost:8080`.  This adds a little more logic.

Testing in `deployments/compose/kwil`:

```
kwild-single           | Root directory "/app/.kwild"
kwild-single           | Private key path: /app/.kwild/private_key
kwild-single           | Generated new private key, path: /app/.kwild/private_key
kwild-single           | Generated new genesis file: /app/.kwild/genesis.json
kwild-single           | Writing config file to /app/.kwild/config.toml
```

image

### DIFF
--- a/cmd/kwild/config/default_config.toml
+++ b/cmd/kwild/config/default_config.toml
@@ -34,19 +34,13 @@
 grpc_listen_addr = "localhost:50051"
 
 # TCP address for the KWILD App's HTTP server to listen on
-http_listen_addr = "localhost:8080"
+http_listen_addr = "0.0.0.0:8080"
 
 # UNIX socket for KWILD Admin server to listen on
 admin_unix_socket = "/tmp/kwil_admin.sock"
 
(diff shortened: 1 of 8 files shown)


cmd/kwild/config/default_config.toml
```diff
@@ -34,20 +34,14 @@
 grpc_listen_addr = "localhost:50051"
 
 # TCP address for the KWILD App's HTTP server to listen on
-http_listen_addr = "localhost:8080"
+http_listen_addr = "0.0.0.0:8080"
```
jchappelow, line 37: Oh, this file isn't actually "used" anywhere, but the default from `Default()` config is still `localhost:8080`.  I should change this line back.

brennanjl (approved):

charithabandi (approved):

PR #584, opened 2024-03-08
chore: compose with build

Yaiba:
This allow `docker compose` to build kwild in one go


jchappelow (approved):
Cool, works as expected with `--build` (or without)

jchappelow:
Will need https://github.com/kwilteam/kwil-db/pull/585 for CI to pass reliably though

Issue #577, opened 2024-03-06
impossible error in (*OracleMgr).Start, no errors from oracles.OracleFunc


jchappelow:
This error cannot be hit:

https://github.com/kwilteam/kwil-db/blob/c6a411c63e89927b137f3b2e20242bab04156db5/internal/oracles/mgr.go#L103-L105

Looking at the registered `OracleFunc`s, I also see no error return, but then there are a lot of unhandled errors and questionable conditions in the eth deposit oracle: https://github.com/kwilteam/kwil-db/blob/c6a411c63e89927b137f3b2e20242bab04156db5/extensions/oracles/eth_deposits/deposits.go#L39-L47

That runs as a goroutine, so that kinda makes sense, but you might not notice that it didn't start, or started with the incorrect configuration.

brennanjl:
Oh that error is a remnant of when the oracle function did return an error. It can simply be removed

brennanjl:
regarding the eth deposit oracle, I think that was just not updated properly after changing the oracle function signature. Will handle now

jchappelow:
For instance, how is appropriate to handle the many possible errors in https://github.com/kwilteam/kwil-db/blob/c6a411c63e89927b137f3b2e20242bab04156db5/extensions/oracles/eth_deposits/deposits.go#L39-L47?
Can return them or can panic etc.

charithabandi:
https://github.com/kwilteam/kwil-db/compare/oracles-start?expand=1 How about something like this?

brennanjl:
They should be logged and then just return.

I didn't really want it to return an error because there is no way for us to handle the error. By not accepting a returned error, I think it is clear to the implementer that they should be very diligent in handling errors. This is particularly important for an oracle, where transient errors are the most common.

brennanjl:
Resolved in: https://github.com/kwilteam/kwil-db/pull/586

brennanjl:
I sort've assumed that Kwild would start without the oracle running, yeah. I guess I can see why this may not be ideal. Do you think we should add the error?

jchappelow:
The way I see it is that if the config.toml defines and oracle and it's config, but it fails to start, the operator would want to know, because if it didn't start, then they are running in an unintended state.

brennanjl:
Ok that makes sense. So I guess we allow them to return errors, but make it clear that an error returned will shut down the node?

jchappelow:
[...]

PR #586, opened 2024-03-08
fix: unused err check and deposit bug

brennanjl:
Fixes 
https://github.com/kwilteam/kwil-db/issues/577

### DIFF
--- a/extensions/oracles/eth_deposits/deposits.go
+++ b/extensions/oracles/eth_deposits/deposits.go
@@ -34,17 +34,16 @@ func init() {
 // "Credit(address,uint256)" and create a deposit event in Kwil when it sees a matching event. It uses the
 // "credit_account" resolution, defined in extensions/resolutions/credit/credit.go, to create the deposit event.
 // It will search for a local extension configuration named "eth_deposit".
-func Start(ctx context.Context, service *common.Service, eventstore oracles.EventStore) {
+func Start(ctx context.Context, service *common.Service, eventstore oracles.EventStore) error {
 	config := &EthDepositConfig{}
 	oracleConfig, ok := service.ExtensionConfigs[OracleName]
 	if !ok {
 		service.Logger.Info("no eth_deposit configuration found, eth_deposit oracle will not start")
-		return
+		return nil // no configuration, so we don't start the oracle
 	}
 	err := config.setConfig(oracleConfig)
 	if err != nil {
-		service.Logger.Error("failed to set eth_deposit configuration", "error", err)
-		return
+		return fmt.Errorf("failed to set eth_deposit configuration: %w", err)
 	}
 
 	// we need to catch up with the ethereum chain.
@@ -53,8 +52,7 @@ func Start(ctx context.Context, service *common.Service, eventstore oracles.Even
 	// whichever is greater
 	lastHeight, err := getLastStoredHeight(ctx, eventstore)
 	if err != nil {
-		service.Logger.Error("failed to get last stored height", "error", err)
-		return
+		return fmt.Errorf("failed to get last stored height: %w", err)
 	}
 
 	if config.StartingHeight > lastHeight {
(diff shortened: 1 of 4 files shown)


brennanjl:
I am going to make the changes mentioned here https://github.com/kwilteam/kwil-db/issues/577, and will then move it out of draft

extensions/oracles/eth_deposits/deposits.go
```diff
@@ -128,8 +122,11 @@ func Start(ctx context.Context, service *common.Service, eventstore oracles.Even
 		return nil
 	})
 	if outerErr != nil {
-		service.Logger.Error("failed to listen to blocks", "error", err)
+		return fmt.Errorf("ethereum client error: %w", outerErr)
 	}
+
+	<-ctx.Done() // wait for the context to be cancelled
```
charithabandi, line 128: Wait, why would we need this here? Doesn't it mess with the start of other oracles ?

brennanjl: No, we run them all in separate goroutines. I guess there isn't really a need for it to wait there though, since the above function would already return in case of context cancellation

charithabandi (approved):
lgtm

jchappelow:
Two seconds.  looking

internal/oracles/mgr.go
```diff
@@ -50,18 +50,25 @@ func NewOracleMgr(config map[string]map[string]string, eventStore *events.EventS
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-errChan:
+			return err
+		default:
 			// still in the catch up mode, do nothing
 			if omgr.cometNode.IsCatchup() {
```
jchappelow, line 73: Not new, but I have recently noticed intense cpu use on startup until sync is achieved, which can be forever until a peer is established.  This loop has no delays, it just calls IsCatchup as fast as possible.

brennanjl: Oops, this is supposed to have a 1s delay in the default case

brennanjl: It does have a 1s delay, at the end of the default case

jchappelow: Yeah, that is too late though because it does `continue` in this `IsCatchup` branch.  I would suggest something along the lines of:


```go
	for {
		select {
		case <-ctx.Done():
			return nil
		case err := <-errChan:
			return err
		case <-time.After(time.Second):
		}
		
		// still in the catch up mode, do nothing
		if omgr.cometNode.IsCatchup() {
			continue
		}
...
```

brennanjl: Just fixed

internal/oracles/mgr.go
```diff
@@ -91,30 +98,32 @@ func (omgr *OracleMgr) Start() {
...
-					}, &scopedKVEventStore{
-						ev: omgr.eventStore,
-						// we add a space to prevent collisions in the KV
-						// oracle names cannot have spaces
-						KV: omgr.eventStore.KV([]byte(name + " ")),
-					})
-					if err != nil {
-						omgr.logger.Error("failed to start oracle", zap.String("name", name), zap.Error(err))
-					}
+					go func(ctx context.Context, start oracles.OracleFunc, name string) {
```
jchappelow, line 101: Possibly a mistake, this `ctx` parameter is not used in the closure, which captures the callers `ctx2` (which is fine).  Probably remove this input.

internal/oracles/mgr.go
```diff
@@ -50,18 +50,25 @@ func NewOracleMgr(config map[string]map[string]string, eventStore *events.EventS
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-errChan:
+			return err
+		case <-time.After(1 * time.Second):
 			// still in the catch up mode, do nothing
 			if omgr.cometNode.IsCatchup() {
 				continue
```
jchappelow, line 74: To confirm, the `break` in the error condition after `GetValidators` is actually intended to break the `switch` , no the `for` loop, and is thus semantically equivalent to `continue`?

brennanjl: I think it is intended to break out of the whole thing.

jchappelow (approved):

PR #588, opened 2024-03-09
Add negative path testcases for validator transactions

charithabandi:
Updated integration tests to test out the below negative path scenarios for validators update transactions

- [x]  Existing validator can't issue a Join request
- [x]  Non validator can't demote its status (Leave request)
- [x]  Joiner can't self approve its join request
- [x]  Only validators can approve the join requests
- [x]  remove requests are issued only on existing validators


brennanjl (approved):
lgtm

PR #589, opened 2024-03-11
parse: fix action recog func call as sql stmt

Yaiba:
This pr fix bug where action parser will parser function calls starts with SQL keywords as SQL statement.

Need to merge kwilteam/action-grammar-go#7 first, then update go mod.

This will requires kuneiform updates as well


brennanjl:
Other is merged, this looks good too.

Yaiba:
After this is merged, we'll need to bump `parse` version up, then update kuneiform, then update kwil-db.

Not sure how do we release on `core` or `parse` package, do we do same release branch like main module? @jchappelow

jchappelow:
> After this is merged, we'll need to bump `parse` version up(patch), then update kuneiform, then update kwil-db.
> 
> Not sure how do we release on `core` or `parse` package, do we do same release branch like main module? @jchappelow

The module branches can be wherever appropriate.   Need to look at the git history for the diffs in these modules.  It's best to keep the needed changes ported to release branch anyway.

brennanjl (approved):

PR #594, opened 2024-03-11
mods: update requires for tagged kuneiform, core, and parse

jchappelow:
This should be the last thing for `v0.7.0` final.

- main module: update kuneiform require to `v0.6.1`
- main module: update to `core/v0.1.1` and `parse/v0.1.2` (remember, no replace statements even though same repo)
- test module: same updates as above, plus the require for the `kwil-db` main module.  Note that I'm doing this in the same commit because the `test` go.mod manually hoists the `core` and `parse` require versions too, and there are no actual code changes in the main module here.


Yaiba (approved):
LGTM

PR #598, opened 2024-03-11
[release-v0.7] pg: reset seq num in replication after commit msg

jchappelow:
https://github.com/kwilteam/kwil-db/pull/597 but for release
Last thing, I think, before tag.

Untracked transaction are intended to have a sequence number of -1, indicating there was no sequence update, which is used by transactions from a pg.DB instance to correlate WAL data updates with a certain transaction.

This commit fixes the replication monitor to work correctly with concurrent untracked transactions by:
- reset the seq var to -1 in the captureRepl loop after each commit
- when a commit is reached with seq still -1 (not updated by an untracked transaction), do NOT send the commit ID on the channel

This also beefs up the logging and error handling in unexpected cases, which corresponds to improper use of replMon methods or concurrent (ab)use of the sentry table to break correct transaction sequencing.

### DIFF
--- a/internal/sql/pg/repl.go
+++ b/internal/sql/pg/repl.go
@@ -51,6 +51,13 @@ func startRepl(ctx context.Context, conn *pgconn.PgConn, publicationName, slotNa
 	// Launch the receiver goroutine, which will send commit digests and an
 	// error on return.
 	done := make(chan error, 1)
+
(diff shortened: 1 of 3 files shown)


charithabandi (approved):

brennanjl (approved):

jchappelow:
Dumb low-level unit test fixed.

PR #599, opened 2024-03-11
spammer listener extension to push the eventstore and pg performance limitation

charithabandi:
Probably not to be pushed in the release till we figure out a better way to test custom extensions. this needs a ext-test build tag and run the tests with -ext flag.

### DIFF
--- a/extensions/others.go
+++ b/extensions/others.go
@@ -1,0 +1,8 @@
+//go:build ext_test
+
+package extensions
+
+import (
+	_ "github.com/kwilteam/kwil-db/extensions/listeners/spammer"
+	_ "github.com/kwilteam/kwil-db/extensions/resolutions/spam"
+)
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -266,3 +266,18 @@ tasks:
     dir: test # different module
     cmds:
       - go test -count=1 -timeout 0 ./integration -v {{.CLI_ARGS}}
+
+  test:it:ext:
+    desc: Run integration tests ('ext' and 'short' mode)
+    deps:
+      - task: build:cli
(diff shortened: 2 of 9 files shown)


jchappelow:
I just noticed that the `-s` in the `grep` of the "Show error log" CI step does not actually make the exit code zero.  I think we need to have it be `grep ... || true`.

extensions/others.go
```diff
@@ -0,0 +1,8 @@
+//go:build ext_test
+
+package extensions
+
+import (
+	_ "github.com/kwilteam/kwil-db/extensions/listeners/spammer"
+	_ "github.com/kwilteam/kwil-db/extensions/resolutions/spam"
```
jchappelow, line 7: This spam test has obvious value as we've identified critical bugs and performance issues, but it hurts a little to have these so readily enabled in the product code of our main module.

How crazy would it be to instead have a "spam node" build defined in the `test` module, in the vein of https://github.com/kwilteam/kwil-db/pull/570. Maybe my worries about this are somewhat illogical?  If others think it's fine to bake these spam extensions into kwild main module code behind a build tag, I guess I'm alright with it too.  Any opinions @Yaiba @brennanjl ?

Yaiba: Seems both are fine, I kind prefer it to be in `test` module if it's not too much work. It seems also more reusable for testing?

charithabandi: Yup, I like it in `test` module too. Will update it

jchappelow: Doing here: https://github.com/jchappelow/kwil-db/tree/spamd

Making a different docker image so it can't get accidentally built as `kwild:latest`

PR #600, opened 2024-03-12
VoteStore refactor

charithabandi:
Refactoring votestore:
Trying to optimize and reduce the db accesses to vote store as much as possible. The changes would touch the processed table as less as possible (because we only have non-processed events in the event store - plz let me know if there any scenarios when this assumption would be false)

Vote store is updated to work with the below principles:
1. Only non-processed events are added to the EventStore
2. Proposer creates resolutions through VoteBodies (and only includes events to which resolutions doesn't already exist)
3. Other nodes will be able to vote only for the existing resolutions. They include the vote in the VoteID Transaction only if they have observed the event corresponding to the resolution. Approval of non-existing resolutions will result in a failure.
4. events gets deleted from the event store when the votes issued by the nodes are processed or when a resolution is approved

I will further look into the access patterns and figure out if we can use in-memory caches for any of these tables, to further reduce these costs, probably in a separate PR.

### DIFF
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/kwilteam/kwil-db/common/sql"
 	"github.com/kwilteam/kwil-db/core/log"
 	"github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/core/types/transactions"
@@ -216,7 +217,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 
 	res := &abciTypes.ResponseFinalizeBlock{}
 
-	err := a.txApp.Begin(ctx)
+	db, err := a.txApp.Begin(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("begin tx commit failed: %w", err)
 	}
@@ -295,7 +296,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 
 	// Broadcast any events that have not been broadcasted yet
 	if a.broadcastFn != nil && len(proposerPubKey) > 0 {
-		err := a.broadcastFn(ctx, proposerPubKey)
+		err := a.broadcastFn(ctx, proposerPubKey, db)
 		if err != nil {
 			return nil, fmt.Errorf("failed to broadcast events: %w", err)
 		}
@@ -881,7 +882,7 @@ func (a *AbciApp) Query(ctx context.Context, req *abciTypes.RequestQuery) (*abci
 	return &abciTypes.ResponseQuery{}, nil
 }
 
-type EventBroadcaster func(ctx context.Context, proposer []byte) error
+type EventBroadcaster func(ctx context.Context, proposer []byte, db sql.DB) error
 
 func (a *AbciApp) SetEventBroadcaster(fn EventBroadcaster) {
 	a.broadcastFn = fn
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -10,7 +10,6 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/kwilteam/kwil-db/common/sql"
 	"github.com/kwilteam/kwil-db/core/log"
 	"github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/core/types/transactions"
@@ -217,7 +216,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 
 	res := &abciTypes.ResponseFinalizeBlock{}
 
-	db, err := a.txApp.Begin(ctx)
+	err := a.txApp.Begin(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("begin tx commit failed: %w", err)
 	}
@@ -296,7 +295,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 
 	// Broadcast any events that have not been broadcasted yet
 	if a.broadcastFn != nil && len(proposerPubKey) > 0 {
-		err := a.broadcastFn(ctx, proposerPubKey, db)
+		err := a.broadcastFn(ctx, proposerPubKey)
 		if err != nil {
 			return nil, fmt.Errorf("failed to broadcast events: %w", err)
 		}
@@ -882,7 +881,7 @@ func (a *AbciApp) Query(ctx context.Context, req *abciTypes.RequestQuery) (*abci
 	return &abciTypes.ResponseQuery{}, nil
 }
 
-type EventBroadcaster func(ctx context.Context, proposer []byte, db sql.DB) error
+type EventBroadcaster func(ctx context.Context, proposer []byte) error
 
 func (a *AbciApp) SetEventBroadcaster(fn EventBroadcaster) {
 	a.broadcastFn = fn
--- a/internal/abci/interfaces.go
+++ b/internal/abci/interfaces.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"math/big"
 
-	"github.com/kwilteam/kwil-db/common/sql"
 	"github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/core/types/transactions"
 
@@ -51,8 +50,8 @@ type TxApp interface {
 	ChainInfo(ctx context.Context) (int64, []byte, error)
 	ApplyMempool(ctx context.Context, tx *transactions.Transaction) error
 	// Begin signals that a new block has begun.
-	Begin(ctx context.Context) (sql.OuterTx, error)
-	Finalize(ctx context.Context, blockHeight int64) (apphash []byte, validatorUpgrades []*types.Validator, err error)
+	Begin(ctx context.Context) error
+	Finalize(ctx context.Context, blockHeight int64) (appHash []byte, validatorUpgrades []*types.Validator, err error)
 	Commit(ctx context.Context) error
 	Execute(ctx txapp.TxContext, tx *transactions.Transaction) *txapp.TxResponse
 	ProposerTxs(ctx context.Context, txNonce uint64, maxTxSize int64, proposerAddr []byte) ([][]byte, error)
--- a/internal/abci/interfaces.go
+++ b/internal/abci/interfaces.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"math/big"
 
+	"github.com/kwilteam/kwil-db/common/sql"
 	"github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/core/types/transactions"
 
@@ -50,8 +51,8 @@ type TxApp interface {
 	ChainInfo(ctx context.Context) (int64, []byte, error)
 	ApplyMempool(ctx context.Context, tx *transactions.Transaction) error
 	// Begin signals that a new block has begun.
-	Begin(ctx context.Context) error
-	Finalize(ctx context.Context, blockHeight int64) (appHash []byte, validatorUpgrades []*types.Validator, err error)
+	Begin(ctx context.Context) (sql.OuterTx, error)
+	Finalize(ctx context.Context, blockHeight int64) (apphash []byte, validatorUpgrades []*types.Validator, err error)
 	Commit(ctx context.Context) error
 	Execute(ctx txapp.TxContext, tx *transactions.Transaction) *txapp.TxResponse
 	ProposerTxs(ctx context.Context, txNonce uint64, maxTxSize int64, proposerAddr []byte) ([][]byte, error)
--- a/internal/events/broadcast/broadcast.go
+++ b/internal/events/broadcast/broadcast.go
@@ -39,9 +39,9 @@ var (
 // EventStore allows the EventBroadcaster to read events
 // from the event store.
 type EventStore interface {
-	// GetUnreceivedEvents gets events that this node has not yet broadcasted.
-	// Events are only marked as "broadcasted" when they have been included in a block.
-	FilterObservedEvents(ctx context.Context, ids []types.UUID) ([]types.UUID, error)
+	// GetObservedEvents filters out the events observed by the validator
+	// that are not previously broadcasted.
+	GetObservedEvents(ctx context.Context, ids []types.UUID) ([]types.UUID, error)
 
 	// MarkBroadcasted marks list of events as broadcasted.
 	MarkBroadcasted(ctx context.Context, ids []types.UUID) error
@@ -126,7 +126,7 @@ func (e *EventBroadcaster) RunBroadcast(ctx context.Context, Proposer []byte, db
 	}
 
 	// Vote only if the note observed the event corresponding to the resolution.
-	ids, err := e.store.FilterObservedEvents(ctx, outstandingResolutionIDs)
+	ids, err := e.store.GetObservedEvents(ctx, outstandingResolutionIDs)
 	if err != nil {
 		return err
 	}
--- a/internal/events/broadcast/broadcast.go
+++ b/internal/events/broadcast/broadcast.go
@@ -22,9 +22,11 @@ import (
 	"math/big"
 
 	cmtCoreTypes "github.com/cometbft/cometbft/rpc/core/types"
+	sql "github.com/kwilteam/kwil-db/common/sql"
 	"github.com/kwilteam/kwil-db/core/crypto/auth"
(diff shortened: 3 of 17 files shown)


charithabandi:
With the updated architecture, the number of roundtrips are reduced drastically and the heavy bottleneck table access like "processed" are reduced too.

- ProposerTxs:  reduced from 3 to 1 
- VoteBody execution: for each event in the voteBody, roundtrips to db reduced from 9 to 3
- VoteID execution: for each eventID, roundtrips to db reduced from 7 to 2

Additionally accesses to the validators/voters table are currently expensive. Below are the stats for the access to voters table Per block:

- GetValidators: called at least 9 times in a block assuming block interval is 6 secs
- GetValidatorPower: called with every validator specific transaction
- TotalPower: called at least #resolutionTypes + 2
Thus at the very least 15 roundtrips to the database can be reduced by maintaining validators/voters info locally.

internal/voting/voting.go
```diff
@@ -23,9 +23,10 @@ const (
 func InitializeVoteStore(ctx context.Context, db sql.DB) error {
 	upgradeFns := map[int64]versioning.UpgradeFunc{
 		0: initTables,
+		1: upgradeV0ToV1,
 	}
 
-	err := versioning.Upgrade(ctx, db, votingSchemaName, upgradeFns, voteStoreVersion)
+	err := versioning.Upgrade(ctx, db, VotingSchemaName, upgradeFns, voteStoreVersion)
```
jchappelow, line 29: Think we need to bump `voteStoreVersion = 1` (well, 2 after rebase) to actually make it run the upgrades.  Probably that says that const is pointless when we can just do `len(upgradeFns)` but that's how we work for now.

jchappelow:
> Vote store is updated to work with the below principles:

How did it work before?  Any trade-offs?

>     1. Only non-processed events are added to the EventStore

Before, we added both processed and non-processed events to the event store?  Wasn't that to prevent repeated approvals for the same event over time?

>     2. Proposer creates resolutions through VoteBodies (and only includes events to which resolutions doesn't already exist)
> 
>     3. Other nodes will be able to vote only for the existing resolutions. They include the vote in the VoteID Transaction only if they have observed the event corresponding to the resolution. Approval of non-existing resolutions will result in a failure.

Before was it that approval could be given before the resolution was recorded?
What's expected when multiple validators observe and publish resolutions for the same event in the same block?  All handled OK, same as before or new handling?  Do the duplicate resolution bodies count as approvals?

>     4. events gets deleted from the event store when the votes issued by the nodes are processed or when a resolution is approved

Same Q as for 1.

charithabandi:
> > Vote store is updated to work with the below principles:
> 
> How did it work before? Any trade-offs?
> 
> > ```
> > 1. Only non-processed events are added to the EventStore
> > ```
> 
> Before, we added both processed and non-processed events to the event store? Wasn't that to prevent repeated approvals for the same event over time?
> 
> > ```
> > 2. Proposer creates resolutions through VoteBodies (and only includes events to which resolutions doesn't already exist)
> > 
> > 3. Other nodes will be able to vote only for the existing resolutions. They include the vote in the VoteID Transaction only if they have observed the event corresponding to the resolution. Approval of non-existing resolutions will result in a failure.
> > ```
> 
> Before was it that approval could be given before the resolution was recorded? What's expected when multiple validators observe and publish resolutions for the same event in the same block? All handled OK, same as before or new handling? Do the duplicate resolution bodies count as approvals?
> 
> > ```
> > 4. events gets deleted from the event store when the votes issued by the nodes are processed or when a resolution is approved
> > ```
> 
> Same Q as for 1.

Even before we were just adding non-processed events in the event store, but vote store methods still kept checking whether an event is processed multiple times in the flow. which is not necessary. 

previously we could register a resolution either with VoteBody or VoteID, now only VoteBody will create a resolution and then all other nodes will look at the created resolutions and vote if the event is observed. Whereas previously, validators look at the events received and vote for all of them, which would also create resolutions if they don't exist. Even if multiple voteIDs were processed in the same block, they would create a single resolution with unique UUID, the only difference btw the voteID vs voteBody created resolutions is the proposerInfo, which exists only with the voteBody. 

and there can only be a single VoteBody transaction in a block by the proposer, so there will never be duplicate resolutions.

internal/events/broadcast/broadcast.go
```diff
@@ -39,7 +41,7 @@ var (
 type EventStore interface {
 	// GetUnreceivedEvents gets events that this node has not yet broadcasted.
 	// Events are only marked as "broadcasted" when they have been included in a block.
-	GetUnreceivedEvents(ctx context.Context) ([]*types.VotableEvent, error)
+	FilterObservedEvents(ctx context.Context, ids []types.UUID) ([]types.UUID, error)
```
jchappelow, line 44: Docs need an update here.

internal/events/events.go
```diff
@@ -152,6 +153,13 @@ func DeleteEvent(ctx context.Context, db sql.Executor, id types.UUID) error {
...
+// DeleteEvents deletes a list of events from the event store.
+// It is idempotent. If the event does not exist, it will not return an error.
+func DeleteEvents(ctx context.Context, db sql.DB, ids ...types.UUID) error {
+	_, err := db.Execute(ctx, deleteEvents, types.UUIDArray(ids))
+	return err
+}
+
 // GetUnreceivedEvents retrieves events that are neither received by the network nor previously broadcasted.
 // An event is considered "received" only after its inclusion in a block.
 // The function excludes events that have been broadcasted but are still pending in the mempool, awaiting block inclusion.
```
jchappelow, line 165: I think we'd remove `GetUnreceivedEvents` now.  But I'm still digesting that the new approach would pull the event IDs that need broadcast from voting store now.

internal/events/sql.go
```diff
@@ -27,11 +28,24 @@ const (
...
+	LEFT JOIN ` + voting.VotingSchemaName + `.resolutions AS r ON e.id = r.id
+	WHERE r.id IS NULL;`
+
+	getUnbroadcastedEvents = `SELECT data, event_type FROM ` + schemaName + `.events WHERE NOT broadcasted;`
+
+	// FilterObservedEvents returns the list of events that are observed by the node and not yet broadcasted.
+	filterObservedEvents = `SELECT unnested.id
+	FROM unnest($1::BYTEA[]) AS unnested(id)
+	LEFT JOIN ` + schemaName + `.events AS res ON unnested.id = res.id
+	WHERE res.id IS NOT NULL AND res.broadcasted = FALSE;`
```
jchappelow, line 48: I think we should seriously consider performing the set difference in the application.  I don't see a better way to actually get the unobserved events from an input array with just one SQL query, but it's clearly a much simpler query to just get a list of the event ids that are observed and in the application do the filter to be left with the ones that aren't in the table.

internal/events/sql.go
```diff
@@ -59,6 +73,9 @@ const (
 		DELETE FROM ` + kvTableName + `
 		WHERE key = $1;
 	`
+
+	// V0 to V1 migration
+	dropReceivedColumn = `ALTER TABLE ` + schemaName + `.events DROP COLUMN received;`
```
jchappelow, line 78: `received` is still in the `CREATE TABLE` statement.  Wait, does our store initialization on a fresh machine run through all the upgrades rather than just starting at the current version?

charithabandi: This is what confusing to me in your PR,. We actually go from version 0 to current. I had it earlier to just start at latest, but I had a strong pushback to go from 0 to latest. So we shouldn't update the tables like u did in yours

jchappelow: This is unrelated to the ability to support existing schema that just lack the version table, which we can live without.

internal/abci/interfaces.go
```diff
@@ -50,8 +51,8 @@ type TxApp interface {
 	ChainInfo(ctx context.Context) (int64, []byte, error)
 	ApplyMempool(ctx context.Context, tx *transactions.Transaction) error
 	// Begin signals that a new block has begun.
-	Begin(ctx context.Context) error
-	Finalize(ctx context.Context, blockHeight int64) (appHash []byte, validatorUpgrades []*types.Validator, err error)
+	Begin(ctx context.Context) (sql.OuterTx, error)
```
jchappelow, line 54: *If* we want to pass the transaction up to abci and on to the broadcaster, we'd give it the minimum required interface.  `sql.OuterTx` gives too much power.

`voting.OutstandingResolutions` only needs a `sql.Executor`.  The `EventBroadcaster` says it only needs a `sql.DB`, which is probably good for more generality (allow Execute or making a transaction/savepoint), so I think we should go with `sql.DB` for this `Begin` interface.

Still thinking through how the evt store's broadcast fn now needs voting store.  Separation of concerns between the two stores is getting hard to track.

jchappelow (commented):
I think I might need to see a drawing or something how how vote+event store work and where their concerns meet. :laughing: I follow the code, but not all the intent.

internal/events/sql.go
jchappelow: Oh, I guess we have to decide about if a new node should run through all the upgrades in empty tables.

But at far as redefining v1 upgrade function, that's ok.  It has only been a couple days since it was defined, and I've not deployed it.  If we had CD or nightly builds then that would be a no no, but it's fine for now.

jchappelow: On second thought, about the v1 redefine, a consideration is if we'd like the meta data update on the release branch but not this vote store refactor.  Just for instance.  I don't know the answer to that

internal/abci/abci.go
```diff
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/kwilteam/kwil-db/common/sql"
```
brennanjl, line 13: Previously, abci wasn't really made aware to the database, and txapp contained all of the logic and transaction management. Breaking that here feels pretty weird IMO. Not sure if I have a concrete suggestion on how to fix it.

internal/events/sql.go
```diff
@@ -27,11 +28,22 @@ const (
...
+	getNewEvents = `SELECT e.data, e.event_type
+	FROM ` + schemaName + `.events AS e
+	LEFT JOIN ` + voting.VotingSchemaName + `.resolutions AS r ON e.id = r.id
+	WHERE r.id IS NULL;`
+
+	// FilterObservedEvents returns the list of events that are observed by the node and not yet broadcasted.
+	filterObservedEvents = `SELECT unnested.id
+	FROM unnest($1::BYTEA[]) AS unnested(id)
+	LEFT JOIN ` + schemaName + `.events AS res ON unnested.id = res.id
+	WHERE res.id IS NOT NULL AND res.broadcasted = FALSE;`
```
brennanjl, line 46: The `WHERE res.id IS NOT NULL` clause is useless. `res.id` is the primary key, it cannot be null

internal/txapp/txapp.go
```diff
@@ -262,25 +298,25 @@ func (r *TxApp) Execute(ctx TxContext, tx *transactions.Transaction) *TxResponse
 }
 
 // Begin signals that a new block has begun.
-// It contains information on any validators whose power should be updated.
-func (r *TxApp) Begin(ctx context.Context) error {
+// It contains information on any validators whos power should be updated.
```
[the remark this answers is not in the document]
charithabandi: Hahah, it wasn't intentional. merge conflict issue 😂

internal/events/sql.go
brennanjl: I would maybe disagree, since the event store could grow fairly large. But why don't you just do:

```sql
SELECT id FROM schema.events WHERE id =ANY($1) AND res.broadcasted = FALSE;
```

internal/abci/interfaces.go
charithabandi: Event store was depending already on VoteStore to ensure that it doesn't add processed events into the Eventstore. I took it one more step ahead with 2 other APIs `GetObservedEvents` and `GetEvents`

internal/events/sql.go
brennanjl: Also, you want to use an inner join here right? Otherwise you are going to be broadcasting vote ids for events you have never observed

internal/events/broadcast/broadcast.go
```diff
@@ -117,18 +120,19 @@ func (e *EventBroadcaster) RunBroadcast(ctx context.Context, Proposer []byte) er
...
-	events, err := e.store.GetUnreceivedEvents(ctx)
+	outstandingResolutionIDs, err := voting.OutstandingResolutions(ctx, db)
 	if err != nil {
 		return err
 	}
 
-	if len(events) == 0 {
-		return nil
+	// Vote only if the note observed the event corresponding to the resolution.
+	ids, err := e.store.GetObservedEvents(ctx, outstandingResolutionIDs)
```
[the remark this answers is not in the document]
charithabandi: Yeah, here we are querying the store if it has observed events corresponding to any of the existing resolutions in the vote store, so yeah if no resolution, no need to broadcast votes for any.

brennanjl:
It seems like a lot of the complexity between the vote store and event store could be solved by joining them? This is obviously not something we current support cross data store, but essentially:

```sql
SELECT v.id
FROM kwild_voting.resolutions AS v
INNER JOIN kwild_events.events AS e ON v.id=e.id
WHERE e.broadcasted = FALSE;
```
I think this is also quite a bit clearer wrt intent of what we are trying to accomplish.

internal/txapp/txapp.go
```diff
@@ -262,25 +298,25 @@ func (r *TxApp) Execute(ctx TxContext, tx *transactions.Transaction) *TxResponse
...
+func (r *TxApp) Begin(ctx context.Context) (sql.OuterTx, error) {
```
brennanjl, line 302: Returning a tx from a method and expecting that the caller does NOT commit it seems extremely weird.

brennanjl: If we have to return a tx to abci, you should return the interface sql.DB instead. This makes it clear that the caller does not have to manage the lifecycle. I still don't think we should return it though.

internal/events/sql.go
jchappelow: This is my confusion - man the input `$1` contain IDs that are NOT not the table?

jchappelow: But to clarify my comment "consider performing the set difference in the application" I don't mean grab all the events back to the app.  Still filter by id (with `=ANY` if it really should NOT be an outer join), but if it really is possible to have events _not in the table_ and we want those included (as outer join suggest) you can deal with that in the app.

charithabandi:
> It seems like a lot of the complexity between the vote store and event store could be solved by joining them? This is obviously not something we current support cross data store, but essentially:
> 
> ```sql
> SELECT v.id
> FROM kwild_voting.resolutions AS v
> INNER JOIN kwild_events.events AS e ON v.id=e.id
> WHERE e.broadcasted = FALSE;
> ```
> 
> I think this is also quite a bit clearer wrt intent of what we are trying to accomplish.

Yeah, I just realized that, updated it. This also removes the need for the need for db in the broadcaster \m/

internal/voting/sql.go
```diff
@@ -4,192 +4,198 @@ package voting
...
 )
+
+// upgrades V0 -> V1
+const (
+	dropHeightTable = `DROP TABLE IF EXISTS ` + VotingSchemaName + `.height`
+)
+
+// upgrades V1 -> V2
+const (
+	dropExtraVoteID = `ALTER TABLE ` + VotingSchemaName + `.resolutions DROP COLUMN extra_vote_id;`
```
jchappelow, line 200: We could combine these, but I only see downsides, no upsides.  Fine like this IMO

charithabandi: yeah, lets keep them this way.

internal/voting/sql.go
```diff
@@ -4,192 +4,198 @@ package voting
...
+		- name: bytea
+		- power: int8
+
+	votes:
+		- resolution_id: uuid
+		- voter_id: uuid
+
+	processed:
+		- id: uuid
+*/
```
jchappelow, line 30: Thanks for laying this out as requested.  @brennanjl This was because I noted that the `CREATE TABLE` statement creates a potentially obsolete table definition, and all the queries written for the application at a newer table structure will be confusing or hard to write without an *effective* schema description.

I had mistakenly assumed there was a short circuit in `versioning.Upgrade` where it would break out of the version loop after applying the `initTables` step and set version to latest without applying all the upgrades to empty tables.  This is somewhat strange to do, but it should be alright if properly documented like this.

If that behavior isn't intended or desired, we can change it.

jchappelow (approved):
LGTM, but untested.
Will run an upgrade on existing node, but seems fine to merge.

internal/voting/sql.go
brennanjl: Yeah, this is why I assumed that a -1 version would be the fastpath with the whole `CREATE TABLE` for the current schema.

brennanjl (approved):

PR #601, opened 2024-03-13
sql,pg,engine: rework sql interfaces

jchappelow:
Reduce sql interfaces, and clean up pg methods.

This change does:

* rework all of the interfaces in `common/sql/sql.go` for clarity and versatility.  See updated type docs there.  Types and interfaces are generally defined from lowest to highest level of composition for easy of understanding.

* With a `*pg.DB`, it was previously required to call `Precommit` before `Commit`, but made the type less able to effectively satisfy the simpler sql interfaces, and it was technically of no need if a commit ID (or 2pc) was not required.  Now it is optional, simplifying it's use in contexts where these features are not relevant (only commit or rollback needed).

* Remove `AccessMode` from the `DB` (and thus `Tx`) interfaces, making it an optional type assertion required by the engine for mutable actions.  This wasn't totally necessary to change, but it was an awkward method for all other systems and their tests.

* Remove many obsolete types and methods from `internal/sql/pg/conn.go`, mostly from when we had implicit sessions rather than transaction-scoped Executors.

* The `OuterTxMaker` interface no longer collides with `TxMaker`, allowing `*pg.DB` to satisfy both.

* Fix `(*Pool).BeginReadTx` beginning a RW txn instead of a RO txn.

* Globally narrow the interfaces on function input arguments to what is actually required.  This improves semantics and testability.

Many other less notable changes...

Some significant ugliness in the event store test setup (events_tests.go) is eliminated.


brennanjl (approved):

brennanjl:
lgtm, will let you merge since the first line says DO NOT MERGE

PR #605, opened 2024-03-14
docker: no wal sender timeout in postgres image

jchappelow:
This removes the wal sender timeout, which is not helpful for kwild's application:

- https://bugzilla.redhat.com/show_bug.cgi?id=1342255
- https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html
- https://www.postgresql.org/docs/current/runtime-config-replication.html

### DIFF
--- a/build/package/docker/postgres.dockerfile
+++ b/build/package/docker/postgres.dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:16.1
+FROM postgres:16.2
 
 # Inject the init script that makes the kwild superuser and a kwild database
 # owned by that kwild user, as well as a kwil_test_db database for tests.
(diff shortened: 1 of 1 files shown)


jchappelow:
I haven't tested this nearly enough, just a few times, but @KwilLuke also hit the error on restore from a closed laptop lid, so to address that edge as well as get on PostgreSQL 16.2 (we were using 16.1), I think we can go ahead and update the `postgres` docker image.  When merged I'll build and push to our Hub account.

brennanjl:
Should we merge this?

charithabandi (approved):

charithabandi:
LGTM! Let's merge this.

jchappelow:
Cool, will push to docker hub shortly.

jchappelow:
`chain_info` errors in CI are so annoying.  However, for some reason I have not observed them causing the intended retry in this experiment PR: https://github.com/kwilteam/kwil-db/pull/641  Either that PR is super lucky or I'm not looking in the right place for the logs...

jchappelow:
> `chain_info` errors in CI are so annoying. However, for some reason I have not observed them causing the intended retry in this experiment PR: #641 Either that PR is super lucky or I'm not looking in the right place for the logs...

Ohhhh yeah, the integration tests are not run with `-v` so stdout is hidden except in failure.  I think that PR is actually working...

PR #609, opened 2024-03-18
engine: create a max procedure call stack depth

jchappelow:
This resolves a bug with the handling of deeply nested action calls. Ultimately the fix may be a gas cost for each action call, possibly exponentially increasing with depth, but I think a hard limit is still a good idea when user-defined recursion is a possibility.  See code comments for more.

I don't think we want to scheme authors to write recursive algorithms, but if we change our mind I think we can life this restriction.

### DIFF
--- a/extensions/precompiles/actions.go
+++ b/extensions/precompiles/actions.go
@@ -62,6 +62,10 @@ type ProcedureContext struct {
 	Procedure string
 	// Result is the result of the most recent SQL query.
 	Result *sql.ResultSet
+
+	// StackDepth tracks the current depth of the procedure call stack. It is
+	// incremented each time a procedure calls another procedure.
+	StackDepth int
 }
 
 // SetValue sets a value in the scope.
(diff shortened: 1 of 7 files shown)


extensions/precompiles/actions.go
```diff
@@ -96,15 +100,16 @@ func (p *ProcedureContext) Values() map[string]any {
...
-		values:    make(map[string]any),
-		DBID:      p.DBID,
-		Procedure: p.Procedure,
+		Ctx:        p.Ctx,
+		Signer:     p.Signer,
+		Caller:     p.Caller,
+		values:     make(map[string]any),
+		DBID:       p.DBID,
+		Procedure:  p.Procedure,
+		StackDepth: p.StackDepth,
```
jchappelow, line 112: Initially I had this with a `+1` from the parent procedure's context, but `(*baseDataset).Call` also does `NewScope` so it can get incremented twice unintentionally.

internal/engine/execution/dataset.go
```diff
@@ -23,9 +23,11 @@ type baseDataset struct {
 	global *GlobalContext
 }
 
+var _ precompiles.Instance = (*baseDataset)(nil)
+
 // Call calls a procedure from the dataset.
 // If the procedure is not public, it will return an error.
-// It implements the Namespace interface.
+// It satisfies precompiles.Instance.
```
jchappelow, line 30: I think these interfaces have been renamed a few times.  This seems to be the one it should mention.

internal/engine/execution/procedure.go
```diff
@@ -19,10 +19,28 @@ import (
...
+//
+// In addition to exorbitant memory required to support a call stack 1 million
+// deep (>1GB), the execution of that many calls can take seconds, even if they
+// do nothing else.
+//
+// Progressive gas metering may be used in the future to limit resources used by
+// abusive recursive calls, but a hard upper limit will likely be necessary
+// unless the price of an action call is extremely expensive or rises
+// exponentially at each level of the call stack.
+const MaxStackDepth = 1000
```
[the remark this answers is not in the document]
brennanjl: I think 1000 is certainly more than enough

brennanjl (approved):

PR #611, opened 2024-03-20
server,pg: fix shutdown sequence

jchappelow:
This resolves the shutdown triggered appHash mismatch I have been observing.  It also resolves some shutdown sequence issues that led to the issue.

@charithabandi's debugging lead to identifying these issues and their fixes. Thanks so much!

In short, we were pulling the plug on parts of the DB subsystems before killing cometbft's node.  Also, we actually didn't have a way to stop the node's replay process before, but now we are using `NewNodeWithContext`.  Also, the `pg` package now does not shutdown the replication monitor goroutines on global context cancellation, only the connection establishment functions.  It now correctly stops the goroutines where it was intended in the `Close` method.

### DIFF
--- a/cmd/kwild/root/root.go
+++ b/cmd/kwild/root/root.go
@@ -2,6 +2,7 @@ package root
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	_ "net/http/pprof"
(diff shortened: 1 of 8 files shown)


charithabandi (approved):
lgtm

brennanjl (approved):

PR #613, opened 2024-03-22
[v0.7.1] back port fixes

jchappelow:
NOTE: Do not merge yet, testing and running CI.  **Also, must merge via rebase, not squash.**

This backports a handful of fixes since v0.7.0.   This is not a direct merge from `main` since there are other breaking changes on `main` that are not included. i.e. these are cherry-picked.

- Support the `cometbft` tool's `insepect` and `rollback` commands by writing effective config to `abci/config`. https://github.com/kwilteam/kwil-db/pull/607
- Fix action recursion issue. https://github.com/kwilteam/kwil-db/pull/609
- Migrate from badger-based chain meta store, fixing possible height disagreement between pg and badger on bad crash timing. https://github.com/kwilteam/kwil-db/commit/2cba60f9c082f78828ec865ae500c1c5e63c87fc
- Fix shutdown sequence bugs that could lead to writing incorrect appHash when shutting down DBs before cometbft node during replay.  https://github.com/kwilteam/kwil-db/pull/611
- sql interface updates.  No notable functional changes, just keeping interfaces and method sets in sync, and avoid conflicts while including in above commits.  https://github.com/kwilteam/kwil-db/commit/6493dc54b26e68d748cd025814748ec47d107f64


charithabandi (approved):

internal/voting/sql.go
```diff
@@ -51,15 +51,7 @@ const (
 		id BYTEA PRIMARY KEY
 	);`
 
-	tableHeight = `CREATE TABLE IF NOT EXISTS ` + votingSchemaName + `.height (
```
charithabandi, line 54: We should not delete the table here?

jchappelow: Ah, yes.  Readded so it actually gets deleted on init. :)

Testing it out again.

PR #614, opened 2024-03-23
Merge events and voting packages

charithabandi:
As events and voting packages are tightly coupled, this PR merges both the packages. 
Still keeping the `kwild_events` and `kwild_voting` schemas separate. `kwild_events` is outside the consensus connection as its accessed asynchronously by the listeners to store externally observed event data. Whereas `kwild_voting` is within consensus and every updates to it has to be deterministic through transactions. 

Other terminology: 
- Events observed by the oracle listeners are added to the event store. and they are referred as eventIDs
- Block proposer takes these events and creates a resolution(proposal) in the vote store for the network to vote on. Once a resolution gets threshold votes, these are applied to the network.
- resolutions are identified within vote store as resolutionID which is essentially same as the eventID.


internal/voting/events.go
```diff
@@ -13,7 +13,86 @@ import (
...
+		received BOOLEAN NOT NULL DEFAULT FALSE, -- received is set to true if the network has received the vote for this event
+		broadcasted BOOLEAN NOT NULL DEFAULT FALSE -- broadcasted is set to true if the event has been broadcasted by the validator. It may or may not have been received by the network
+	);`
+	dropEventsTable = `DROP TABLE IF EXISTS ` + schemaName + `.events;`
+
+	insertEventIdempotent = `INSERT INTO ` + schemaName + `.events (id, data, event_type) VALUES ($1, $2, $3) ON CONFLICT DO NOTHING;`
+	deleteEvent           = `DELETE FROM ` + schemaName + `.events WHERE id = $1;`
+
+	deleteEvents = `DELETE FROM ` + schemaName + `.events WHERE id =ANY($1);`
+	getEvents    = `SELECT data, event_type FROM ` + schemaName + `.events;`
```
jchappelow, line 53: `getEvents` is also unused now.

internal/voting/events.go
```diff
@@ -13,7 +13,86 @@ import (
...
+	getNewEvents = `SELECT e.data, e.event_type
+	FROM ` + schemaName + `.events AS e
+	LEFT JOIN ` + votingSchemaName + `.resolutions AS r ON e.id = r.id
+	WHERE r.id IS NULL;`
+
+	// eventsToBroadcast returns the list of the resolutionIDs observed by the validator that are not previously broadcasted
+	eventsToBroadcast = `SELECT e.id
+	FROM ` + votingSchemaName + `.resolutions AS r
+	INNER JOIN ` + schemaName + `.events AS e ON r.id = e.id
+	WHERE  e.broadcasted = FALSE;`
```
jchappelow, line 65: `e.broadcasted` is a `BOOLEAN` so it's more pg idiomatic to say `WHERE NOT e.broadcasted` or `WHERE e.broadcasted`

internal/voting/events_test.go
```diff
@@ -188,7 +179,7 @@ func Test_EventStore(t *testing.T) {
 			require.NoError(t, err)
 			defer cleanup()
 
-			err = voting.InitializeVoteStore(ctx, db)
+			err = InitializeVoteStore(ctx, db)
 			require.NoError(t, err)
 
 			e, err := NewEventStore(ctx, db) // needs BeginReadTx
```
jchappelow, line 185: One of the main things I though this work could help improve was the burden of having the two constructors, and knowing that one needs to be initialized before the other.

Now that there are queries that JOIN between the schemas, it no longer makes sense to be have separate "init" or constructors.  It was already a problem before because event store methods required voting store to be initialized since it used query functions from `voting`, and now we're solidifying that dependency.

I see that it's possible to test voting schema/store function independently in `vote_test.go`, but does it make sense for both the events test and the actual application to ensure voting store is ready via the event store constructor?  That's the main thing I figured we'd gain from this.

internal/voting/events.go
```diff
@@ -13,7 +13,86 @@ import (
...
+
+// Life cycle of an event:
+// 1. Events received from an external source is stored in the event store.
+// 2. Block proposer will create resolutions for the events observed.
+// Resolution is a proposal based on an event, to which the network will vote on
+// to get processed and applied.
+// 3. Other voters vote on the existing resolutions if they witnessed the event.
+// 4. Once the resolution meets required threshold votes, the resolution gets
+// approved and processed.
+// 5. Events are deleted from the event store (either in step 2, 3 or 4)
```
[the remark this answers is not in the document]
charithabandi: No, if u are other voter, it gets deleted when your vote is processed.  If your vote has not yet processsed, but the resolution is approved, u delete it when its processed.

jchappelow (commented):
I anticipated a single constructor making it less error prone and more clear, but it actually went the other way. I see why and am trying a change presently.  If it works I'll suggest it shortly.  Sorry for the delay.

internal/txapp/txapp.go
```diff
@@ -671,6 +666,8 @@ func (r *TxApp) AccountInfo(ctx context.Context, acctID []byte, getUncommitted b
 // largest nonce of the transactions included in the block that are authored by the proposer.
 // This transaction only includes voteBodies for events whose bodies have not been received by the network.
 // Therefore, there won't be more than 1 VoteBody transaction per event.
+//
+// @charitha: does this method need to return [][]byte or just []byte (one tx)?
```
charithabandi, line 670: RIght now we only have 1 proposerTx, but might expand in future if we need to add any other kind of proposerTxs, so designed it to accommodate that.

internal/voting/voting.go
```diff
@@ -69,37 +72,40 @@ func dropExtraVoteIDColumn(ctx context.Context, db sql.DB) error {
...
 	if err != nil {
 		return err
 	}
 	defer tx.Rollback(ctx)
 
-	// Expectation is that the resolution is already created when the voteBody is submitted. and nodes wont submit the voteIDs for events which doesn't have resolutions.
-	userId := types.NewUUIDV5(from)
+	// Expectation is that the resolution is already created when the voteBody
+	// is submitted. Nodes won't submit the voteIDs for events which don't have
+	// resolutions. @charitha: is this still accurate?
```
charithabandi, line 90: Correct, Votes can only be made for existing resolutions. If resolution doesn't exist, this will fail.

jchappelow:
Repro for the issue I was working around:

```
task build:docker
cd test/integration
go test -v -count 1 -timeout 0 -run TestKwildEthDepositOracleValidatorUpdates
```

Very often when node5 rejoins, they fail to broadcast their vote in the event despite logs showing that it was observed and processed when the listener starts up with the validator status change.  I have a case ready to see now.

jchappelow (approved):
Looks like we got all the issues sorted!

PR #619, opened 2024-03-28
Mark rebroadcast only on the voteID Tx issuer

charithabandi:
If there is a ValidatorVoteID transaction, mark the voteIDs for rebroadcasting only on the issuer's node. Currently all the nodes would mark the ids as rebroadcast.

### DIFF
--- a/internal/txapp/mempool.go
+++ b/internal/txapp/mempool.go
@@ -1,6 +1,7 @@
 package txapp
 
 import (
+	"bytes"
 	"context"
 	"encoding/hex"
 	"errors"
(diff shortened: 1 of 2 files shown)


jchappelow:
Thanks, looks as we discussed.

This is a real security concern, but only in the presence of a malicious validator since before writing the needs-rebroadcast flag to DB we were rejecting `PayloadTypeValidatorVoteIDs` if the sender was not a current validator.  Could consider backport, but does not seem critical.

jchappelow (approved):
LGTM. `nodeAddr` is potentially misleading since the identity returned (and compared to tx.Sender) is the ed25519 key rather than the cometbft node ID/address, but this is clear enough.

PR #623, opened 2024-04-01
core: gateway response change

Yaiba:
This update gateway client to be compatible with new kgw repsonse structure, which is compatible with swagger error response

This requires both kwil-cli and kgw update

This should be merged once after kwilteam/kgw#29 is merged

PR #624, opened 2024-04-01
[v0.7.2] core: fix parse kgw error, so it output detailed message

Yaiba:
The issuse: when kwil-cli request a provider, error detail is not avaliable for kgw response.

Since both kwild and kgw requests are handled by same http client, and for now they have different error response(#623 try to fix this)
This pr change the `parseErrorResponse` function, to support kgw error response.

The output(mocked):

before

```
$ .build/kwil-cli database call --action owner_only --name testdb --kwil-provider http://localhost:8090 --authenticate
error calling action: call action: 400 Bad Request
err code = 0, msg =

$ .build/kwil-cli database call --action owner_only --name testdb --kwil-provider http://localhost:8080
error calling action: call action: 400 Bad Request
err code = 3, msg = mock call bad request error
```

after 
```
$ .build/kwil-cli database call --action owner_only --name testdb --kwil-provider http://localhost:8090 --authenticate
error calling action: call action: 400 Bad Request
JUST SOME CALL ERROR !!!

$ .build/kwil-cli database call --action owner_only --name testdb --kwil-provider http://localhost:8080
error calling action: call action: 400 Bad Request
err code = 3, msg = mock call bad request error
```
8090 is kgw, 8080 is kwil-db

### DIFF
--- a/core/rpc/client/user/http/error.go
+++ b/core/rpc/client/user/http/error.go
@@ -90,6 +90,18 @@ func parseBroadcastError(respTxt []byte) (bool, error) {
 			return false, err
 		}
 	}
+
+	// for kgw error response
+	if protoStatus.GetCode() == 0 { // if it's grpc status, it should have a code, code 0 is not error code
+		// NOTE: this could be removed once #623 is merged
+		// try to parse kgw error first
+		var kgwErr gatewayErrResponse
+		err := json.Unmarshal(respTxt, &kgwErr)
+		if err == nil {
+			return true, errors.Join(errors.New(kgwErr.Error))
+		}
+	}
+
 	stat := grpcStatus.FromProto(&protoStatus)
 	code, message := stat.Code(), stat.Message()
 	rpcErr := &client.RPCError{
--- a/core/rpc/client/user/http/error.go
+++ b/core/rpc/client/user/http/error.go
@@ -17,12 +17,27 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
+type gatewayErrResponse struct {
+	Error string `json:"error"`
+}
+
 func parseErrorResponse(respTxt []byte) error {
 	// NOTE: here directly use status.Status from googleapis/rpc/status
 	var res status.Status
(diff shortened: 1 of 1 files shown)


core/rpc/client/user/http/error.go
```diff
@@ -25,6 +29,17 @@ func parseErrorResponse(respTxt []byte) error {
 		return err
 	}
 
+	// for kgw error response
+	if res.GetCode() == 0 { // if it's grpc status, it should have a code, code 0 is not error code
```
Yaiba, line 33: It seems I also need to change `parseBroadcastError` function

Yaiba:
for broadcast error:(mocked output)
```
$ .build/kwil-cli database deploy -p test/acceptance/test-data/test_db.kf --kwil-provider http://127.0.0.1:8090
failed to deploy database: JUST SOME BEEEEEEEE error

.build/kwil-cli database deploy -p test/acceptance/test-data/test_db.kf --kwil-provider http://127.0.0.1:8090
failed to deploy database: err code = 12, msg = method Broadcast not implemented
```

jchappelow (approved):
OK, makes sense for release 0.7 branch.  This code would become dead with:
- https://github.com/kwilteam/kgw/pull/29
- https://github.com/kwilteam/kwil-db/pull/623

jchappelow:
Although I'm not testing with kgw, I looked again at `github.com/grpc-ecosystem/grpc-gateway/v2@v2.18.0/runtime.DefaultHTTPErrorHandler`, and it is indeed going to:
- select a non-zero grpc `codes.Code` for a non-nil `error`
- AND set the `http.Response.Status` to >=300, which is what the swagger *client* keys on when deciding to return a `httpTx.GenericSwaggerError` from any of the `TxServiceApi` methods.

This all looks right.  Please merge when you're comfortable @Yaiba

Yaiba:
> Member

cool, thanks for the double check

Yaiba:
Should we bump 'core' version? 
And once our main(v0.8) release, does 'core' need bump to v0.2? since it's KGW API breaking change

PR #628, opened 2024-04-02
Enable state sync and snapshot support

charithabandi:
This PR enables the support to create and manage logical Postgres snapshots and accelerate the catch-up process of new node using statesync.

## Snapshot store
Snapshot store in the snapshots module is responsible for creating and maintaining snapshots. Snapshots are the logical sql dump of the Postgres database and is stored on the device in chunks of max size 16MB (imposed by the cometBft statesync). Snapshot directory has to follow this strict directory pattern `snapshot-dir/block-<number>/format-0/chunks/`. 
Snapshot header stored under `snapshot-dir/block-<number>/format-0/header.json` would represent the snapshot metadata such as the snapshot height, hash, number of chunks, each chunk hash etc.
Snapshot store also ensures that the snapshots are taken at correct heights and maintains only max-snapshots on disk at any point in time. 

### Snapshot creation:
- Snapshot transaction with Serializable isolation level is created at the end of the block commit
-  Snapshot ID is created to track the DB state at the end of the block
- `pg_dump` utility is used for creating snapshots at that specific snapshot ID. 
-  To make these dumps deterministic, this process goes through extra sanitization to order all the rows in the tables, to remove SET/SELECT statements which are not required, and white spaces are removed to reduce the space. 
- Sanitized snapshots are further compressed to minimize the space usage
- Compressed snapshots are divided into chunks of max size 16MB.

## StateSync
Having the snapshots enables us to bootstrap the new node's catchup process rapidly using the CometBFT's statesync feature. 
Requirements:
Trusted Snapshot providers:  Network should have a bunch of full nodes that can provide and validate block headers and snapshots.

New node that is joining, requests the networks for snapshots. All the peers respond to this with the metadata corresponding to the  list of snapshots that they currently have. Once the new node validate the snapshot with the trusted snapshot provider's snapshot at the same height, the new node requests for the snapshot chunks. 

Statesync module is responsible for validating the snapshots, keeping track of the chunks received, decompressing and restoring the DB state. Once the DB is restored and the app hash is validated, if there are more blocks to sync, Cometbft would switch to the block sync mode. This new node caught up using the statesync would have truncated block history with starting height at the snapshot height.

## Docker builds
Kwild Docker images are now shipped with the  the` postgresql-client` tools to run `pg-dump` and `psql`

Spammer tests are run with the snapshots enabled for performance tests 

## New additions

- Trust options for Statesync are automated to enhance the user experience.

- Kwil-admin snapshot create command is added to  create a db snapshot for genesis state during network migrations. This command also spews out the genesis file with validators and genesis hash defined based on the previous network. Other params can be updated manually by the network operators. 

- Network migrations are enforced through `snapshot_file` in `config.toml` and  `app_hash` in the `genesis.json`. If the app hash is present in the genesis file, it implies that the node is supposed to start with an initial state and the initial state can be loaded using the snapshot file. Also the hash of the uncompressed logical dump should match that of the genesis hash ensuring that the node is starting with the correct state. Also reinforces that all nodes start with same state. However, the db is restored from the snapshot only if the node is starting fresh from genesis height. DB restoration is skipped if the node is not at genesis height or if statesync is enabled or if genesis app hash is not specified. 

- Removed dependencies between statesync and snapshots packages.


brennanjl:
I think it'd probably be useful to include your comment sections of `Snapshot Store` and `StateSync` as a Github Wiki. They're very helpful

charithabandi:
> I think it'd probably be useful to include your comment sections of `Snapshot Store` and `StateSync` as a Github Wiki. They're very helpful

Yup, gonna add docs and wiki for both of these.

charithabandi:
Might be a good thing to have a kwil-cli support to create snapshots on demand. (instead of waiting on the recurring-height)

brennanjl:
> Might be a good thing to have a kwil-cli support to create snapshots on demand. (instead of waiting on the recurring-height)

I think this would probably be `kwil-admin` rather than `kwil-cli`, but yeah totally agree

cmd/kwild/server/build.go
```diff
@@ -441,31 +452,52 @@ func initAccountRepository(d *coreDependencies, tx sql.Tx) {
...
-	// 	cfg.SnapshotConfig.SnapshotDir,
-	// 	cfg.SnapshotConfig.RecurringHeight,
-	// 	cfg.SnapshotConfig.MaxSnapshots,
-	// 	snapshots.WithLogger(*d.log.Named("snapshotStore")),
-	// )
+	dbCfg := &snapshots.DBConfig{
+		DBUser: cfg.DBUser,
+		DBPass: cfg.DBPass,
+		DBHost: cfg.DBHost,
+		DBPort: cfg.DBPort,
```
brennanjl, line 465: You need to also include db name here. I noticed you have it hardcoded in the pgDumpCmd to `kwild`, which is our default, however it is configurable.

internal/abci/abci.go
```diff
@@ -428,6 +411,11 @@ func (a *AbciApp) Info(ctx context.Context, _ *abciTypes.RequestInfo) (*abciType
 		a.validatorAddressToPubKey[addr] = val.PubKey
 	}
 
+	err = a.txApp.ReloadDB(ctx)
```
brennanjl, line 414: This certainly seems like the wrong place to call this. `Info` is called every time `/abci_info` is called. Not only do we not want to do a full DB reload here, but there is also a race condition with the engine's in-memory databases

internal/txapp/txapp.go
```diff
@@ -205,6 +208,33 @@ func (r *TxApp) ChainInfo(ctx context.Context) (int64, []byte, error) {
 	return height, appHash, nil
 }
 
+// ReloadDB reloads the database state into the engine.
+func (r *TxApp) ReloadDB(ctx context.Context) error {
+	tx, err := r.Database.BeginReadTx(ctx)
```
brennanjl, line 213: Having this here also seems potentially wrong. I am assuming the point of ReloadDB is to reload the in-memory info after syncing data, sync the info is added when the engine constructor is called. Is there a simpler way to manage this?

brennanjl:
This is more of a general question, but what was the reason for having `snapshots` and `statesync` be separate packages? It seems like they are extremely related, and almost everything exported from `snapshots` is only exported so that statesync can access it.

cmd/kwild/server/build.go
```diff
@@ -441,31 +452,52 @@ func initAccountRepository(d *coreDependencies, tx sql.Tx) {
...
+
+	snapshotCfg := &snapshots.SnapshotConfig{
+		SnapshotDir:     cfg.Snapshots.SnapshotDir,
+		RecurringHeight: cfg.Snapshots.RecurringHeight,
+		MaxSnapshots:    int(cfg.Snapshots.MaxSnapshots),
+	}
+
+	snapshotter := snapshots.NewSnapshotter(dbCfg, cfg.Snapshots.SnapshotDir, *d.log.Named("snapshotter"))
+
+	ss, err := snapshots.NewSnapshotStore(snapshotCfg, snapshotter, *d.log.Named("snapshotStore"))
```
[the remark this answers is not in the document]
charithabandi: I did that for ease of testing. It enabled me to  mock-out the snapshotter while testing the store.  But that can be done with a single constructor too.

internal/snapshots/snapshotter.go
```diff
@@ -0,0 +1,444 @@
...
+		"--snapshot", snapshotID,
+		// Connection options
+		"-U", s.dbConfig.DBUser,
+		"-h", s.dbConfig.DBHost,
+		"-p", s.dbConfig.DBPort,
+	)
+
+	s.log.Debug("Executing pg_dump", zap.String("cmd", pgDumpCmd.String()))
+
+	output, err := pgDumpCmd.CombinedOutput()
```
brennanjl, line 142: I'm a bit concerned about running out of memory here. We are essentially dumping the entire DB to memory, right?

charithabandi:
> This is more of a general question, but what was the reason for having `snapshots` and `statesync` be separate packages? It seems like they are extremely related, and almost everything exported from `snapshots` is only exported so that statesync can access it.

This is something I went back and forth on. I did refactor at the end separating these two, to maintain separation of concerns. Snapshots are responsible for snapshot creation and maintenance and distribution of them. Whereas Statesync is just one feature which uses snapshots and this feature is only used by the new node that want to accelerate the catch up process. There can be many other features later that can make use of snapshot store later. So I wanted to keep snapshots out of the statesync

internal/snapshots/snapshotter.go
```diff
@@ -0,0 +1,444 @@
...
+	}
+
+	s.log.Info("Sanitized dump file", zap.Uint64("height", height), zap.String("dumpFile hash", fmt.Sprintf("%x", hash)))
+
+	return hash, nil
+}
+
+// CompressDump is the STAGE3 of the snapshot creation process
+// This method compresses the sanitized dump file using gzip compression
+// Should we do inline compression? or using exec.Command?
```
brennanjl, line 301: Inline seems right here. I see no reason to use exec, it just introduces more room for issues

internal/statesync/codes.go
```diff
@@ -0,0 +1,52 @@
...
+	ErrInvalidSnapshot           = fmt.Errorf("invalid snapshot")
+
+	ErrAbortSnapshotChunk  = fmt.Errorf("abort snapshot chunk")
+	ErrRetrySnapshotChunk  = fmt.Errorf("retry snapshot chunk") // retries without refetching the chunk
+	ErrRetrySnapshot       = fmt.Errorf("retry snapshot")
+	ErrRejectSnapshotChunk = fmt.Errorf("reject snapshot chunk")
+)
+
+func ToABCIOfferSnapshotResponse(err error) abciTypes.ResponseOfferSnapshot_Result {
+	switch err {
```
brennanjl, line 25: Function is unused

internal/statesync/codes.go
```diff
@@ -0,0 +1,52 @@
...
+
+func ToABCIApplySnapshotChunkResponse(err error) abciTypes.ResponseApplySnapshotChunk_Result {
+	switch err {
+	case nil:
+		return abciTypes.ResponseApplySnapshotChunk_ACCEPT
+	case ErrAbortSnapshotChunk:
+		return abciTypes.ResponseApplySnapshotChunk_ABORT
+	case ErrRetrySnapshotChunk:
+		return abciTypes.ResponseApplySnapshotChunk_RETRY
+	case ErrRejectSnapshotChunk:
```
brennanjl, line 47: Should probably use `errors.Is` here?

internal/engine/execution/global.go
```diff
@@ -156,6 +156,27 @@ func NewGlobalContext(ctx context.Context, db sql.Executor, extensionInitializer
 	return g, nil
 }
 
+// Reload is used to reload the global context based on the current state of the database.
+// It is used after state sync to ensure that the global context is up to date.
+func (g *GlobalContext) Reload(ctx context.Context, db sql.Executor) error {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	schemas, err := getSchemas(ctx, db)
```
brennanjl, line 165: Maybe rename this to `ReloadCache`, since it is more descriptive of what is occurring?

Also, remembering the bug Paulo ran into where a failed drop was still in memory: since we are full table scanning and reloading our cache, should we delete the cache in the first place?

internal/txapp/txapp.go
brennanjl, line 213: Super minor, but I feel like `ReloadCache` or `ReloadMemory` is more descriptive. When I first saw `ReloadDB`, I was very confused what that meant, and the implications it could have on data. `ReloadCache` is much more innocuous (which this is).

internal/txapp/txapp.go
```diff
@@ -205,6 +208,33 @@ func (r *TxApp) ChainInfo(ctx context.Context) (int64, []byte, error) {
...
+
+	// Check if we are switching from StateSync?
+	// If so, we need to reload the engine.
+	height, app_hash, err := getChainState(ctx, tx)
+	if err != nil {
+		return err
+	}
+
+	if r.height == 0 && height > r.height && r.appHash == nil {
+		err = r.Engine.Reload(ctx, tx)
```
brennanjl, line 227: This is a bit confusing to me. A method called `ReloadDB` that reloads our in-memory catalog that is actually dependent on the current state of the chain. I'm not totally sure what the point of this logic is, but given how not often this method would be called, why wouldn't we just call `r.Engine.Reload` every time?

This might lead to us table scanning the database catalog unnecessarily on startup, but I feel like that is not a very big deal, and makes this function better do what it says it does (which makes it less bug prone)

If this logic is really necessary, we should get a better function name.

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+var (
+	createLongExplain = `Creates a snapshot of the database. The command is used during network migration to get the state of the KwilDB.
+		It interacts directly with the Postgres server without intervening the kwild node.`
+)
+
+func createCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create <db-name> <db-user> <db-host> <db-port> <snapshot-dir>",
```
brennanjl, line 24: Could the db config be optional flags instead?

Defaults would likely be:
name: postgres
user: postgres
host: localhost
port: 5432

Also, you forgot password (which the default would be "password")

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -0,0 +1,145 @@
...
+		Short: "Creates a snapshot of the database.",
+		Long:  createLongExplain,
+		Args:  cobra.ExactArgs(5),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
```
brennanjl, line 29: We should use `cmd.Context()`. If the user cancels the command, we want the context to cancel too.

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -0,0 +1,145 @@
...
+		return fmt.Errorf("failed to close gzip writer: %w", err)
+	}
+
+	if err := sedCmd.Wait(); err != nil {
+		return fmt.Errorf("failed to wait for sed command: %w", err)
+	}
+
+	// Generate snapshot hash and use it as the genesis hash.
+	hash := hasher.Sum(nil)
+	hashStr := hex.EncodeToString(hash)
```
brennanjl, line 129: Isn't all of this logic already defined in `internal/snapshots`?

It seems like it would be preferable to rely on that, especially because this seems like something that could change version to version

cmd/kwil-admin/cmds/snapshot/restore.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+var (
+	restoreLongExplain = `Creates a snapshot of the database. The command is used during network migration to get the state of the KwilDB.
+		It interacts directly with the Postgres server without intervening the kwild node.`
+)
+
+func restoreCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "restore <db-user> <db-host> <db-port> <snapshot-file>",
```
brennanjl, line 22: You are missing both password and db name. I also think these should be flags.

cmd/kwil-admin/cmds/snapshot/restore.go
```diff
@@ -0,0 +1,88 @@
...
+		Short: "Restore the db state from the snapshot file.",
+		Long:  restoreLongExplain,
+		Args:  cobra.ExactArgs(4),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := context.Background()
```
brennanjl, line 27: Should be `cmd.Context()`

cmd/kwil-admin/cmds/snapshot/restore.go
```diff
@@ -0,0 +1,88 @@
...
+		"--port", dbPort,
+		"--dbname", "postgres",
+	)
+	psqlCmd.Stderr = os.Stderr
+
+	stdinPipe, err := psqlCmd.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("failed to get stdin pipe: %w", err)
+	}
+	defer stdinPipe.Close()
```
brennanjl, line 56: This seems to be mostly already defined in `internal/statesync`

internal/abci/abci.go
```diff
@@ -496,16 +479,30 @@ func (a *AbciApp) InitChain(ctx context.Context, req *abciTypes.RequestInitChain
 
 // ApplySnapshotChunk is on the state sync connection
 func (a *AbciApp) ApplySnapshotChunk(ctx context.Context, req *abciTypes.RequestApplySnapshotChunk) (*abciTypes.ResponseApplySnapshotChunk, error) {
-	refetchChunks, status, err := a.bootstrapper.ApplySnapshotChunk(req.Chunk, req.Index)
+	if a.statesyncer == nil {
+		return nil, errors.New("bootstrapper not set")
```
brennanjl, line 483: Since I suspect that this will be a common error users hit, could we be more descriptive about what is happening here? Something along the lines of `snapshots are not configured on your local node...` (ideally a more specific diagnosis than that if that is not what is wrong here).

internal/abci/abci.go
```diff
@@ -496,16 +479,30 @@ func (a *AbciApp) InitChain(ctx context.Context, req *abciTypes.RequestInitChain
...
+	}
+
+	dbRestored, refetch, err := a.statesyncer.ApplySnapshotChunk(ctx, req.Chunk, req.Index)
 	if err != nil {
-		return &abciTypes.ResponseApplySnapshotChunk{Result: abciStatus(status), RefetchChunks: refetchChunks}, nil
+		var refetchChunks []uint32
```
brennanjl, line 488: could you add a comment here explaining what is going on, specifically regarding refetch?

internal/abci/abci.go
```diff
@@ -880,6 +898,34 @@ func (a *AbciApp) ProcessProposal(ctx context.Context, req *abciTypes.RequestPro
...
+				snapshot = s
+				break
+			}
+		}
+
+		if !exists {
+			return &abciTypes.ResponseQuery{}, nil
+		}
+
+		bts, err := json.Marshal(snapshot)
```
brennanjl, line 923: I don't believe this is an issue, but just want to flag: json.Marshal is non-deterministic. I'm not sure if that is an issue here

cmd/kwil-admin/cmds/snapshot/create.go
jchappelow: I think password and port need to be optional. If we're connecting with unix sockets, port has to be empty, and if and "trust" auth in pg_hba.conf, password also should be empty.

Needing to have empty strings for some if these, together with sensible defaults that will work with many installs including our docker configs, it suggests flags would be preferable to positionals.

Is this why you were asking about URI instead? Yeah, maybe. I like to be able to just specify e.g. host, and go with the defaults for the other pieces though, and I can't do that with a URI.

internal/txapp/txapp.go
charithabandi: Ah yeah, I had this logic earlier to determine if Info is getting called from Statesync or normal block sync init. Now that we moved it out of Info() we don't these checks, I will remove them

cmd/kwil-admin/cmds/snapshot/create.go
charithabandi: Not actually, the compression part is same, but the pg_dump args and the sanitization process is a bit different.

cmd/kwil-admin/cmds/snapshot/restore.go
```diff
@@ -0,0 +1,88 @@
...
+		reader = gzipReader
+	} else {
+		reader = snapfs
+	}
+
+	// Copy data to stdin of the psql command
+	if _, err := io.Copy(stdinPipe, reader); err != nil {
+		return fmt.Errorf("failed to copy snapshot file to psql: %w", err)
+	}
+	stdinPipe.Close() // Ensure closure of the pipe
```
jchappelow, line 80: meaning you have to close the pipe or psqlCmd.Wait will hang?  Or just need to close before returing?

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -149,6 +149,22 @@ hostname = "{{ .AppCfg.Hostname }}"
...
+[app.snapshots]
+
+# Enables snapshots
+enabled = {{.AppCfg.Snapshots.Enabled}}
+
+# Path to the snapshots directory
+snapshot_dir = "{{.AppCfg.Snapshots.SnapshotDir}}"
+
+# Heights at which the database is snapshotted.
+snapshot_heights = {{.AppCfg.Snapshots.RecurringHeight}}
```
jchappelow, line 164: Same comment about more clear doc.  The template var *Recurring* provides the clearest clue that this is a modulo.

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -257,4 +273,34 @@ max_tx_bytes = {{ .ChainCfg.Mempool.MaxTxBytes }}
 
 # Size of the cache (used to filter transactions we saw earlier) in transactions
 cache_size = {{ .ChainCfg.Mempool.CacheSize }}
+
+#######################################################
+###         State Sync Configuration Options        ###
+#######################################################
+[chain.statesync]
```
jchappelow, line 280: Let's put a newline between the section heading and anything after, for readability.

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -257,4 +273,34 @@ max_tx_bytes = {{ .ChainCfg.Mempool.MaxTxBytes }}
...
+# Snapshots are accepted for statesync only after verifying it with these trusted snapshot providers.
+# These are also used for light client verification of the synced state machine and
+# retrieval of state data for node bootstrapping.
+# Light client verification needs a trusted height and block hash at the configured block height obtained from a
+# trusted source, and a period during which validators can be trusted.
+# Atleast 2 trusted rpc servers are required for enabling state sync.
+rpc_servers = "{{ .ChainCfg.StateSync.RPCServers }}"
+trust_height = {{ .ChainCfg.StateSync.TrustHeight }}
+trust_hash = "{{ .ChainCfg.StateSync.TrustHash }}"
+trust_period = "{{ .ChainCfg.StateSync.TrustPeriod }}"
```
jchappelow, line 298: These trust settings are missing from cmd/kwild/config/default_config.toml and kwil-db/cmd/kwild/config/test_data/config.toml

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -257,4 +273,34 @@ max_tx_bytes = {{ .ChainCfg.Mempool.MaxTxBytes }}
...
+
+# Time to spend discovering snapshots before initiating a restore.
+discovery_time = "{{ .ChainCfg.StateSync.DiscoveryTime }}"
```
jchappelow, line 301: What are the considerations in setting `discovery_time`.  Looks like the default is 15 seconds, but that seems very short if your node has to use PEX for any amount of time to reach nodes that are serving the snap.

What happens if it's longer?  Does that make it annoying if there aren't any snap-serving nodes?  You have the default of `Enable: false` (sensibly), so it seems a longer discover time would be fine if they explicitly turn it on.

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -257,4 +273,34 @@ max_tx_bytes = {{ .ChainCfg.Mempool.MaxTxBytes }}
...
+
+# The timeout duration before re-requesting a chunk, possibly from a different
+# peer (default: 1 minute).
```
jchappelow, line 304: When does the re-request occur?  1 minute after what previous event?

cmd/kwild/config/flags.go
```diff
@@ -34,16 +34,14 @@ func AddConfigFlags(flagSet *pflag.FlagSet, cfg *KwildConfig) {
...
-	// TODO: Snapshots are not supported yet
-	// // Snapshot Config flags
-	// flagSet.BoolVar(&cfg.AppCfg.SnapshotConfig.Enabled, "app.snapshots.enabled", cfg.AppCfg.SnapshotConfig.Enabled, "Enable snapshots")
-	// flagSet.Uint64Var(&cfg.AppCfg.SnapshotConfig.RecurringHeight, "app.snapshots.recurring-height", cfg.AppCfg.SnapshotConfig.RecurringHeight, "Recurring snapshot height")
-	// flagSet.Uint64Var(&cfg.AppCfg.SnapshotConfig.MaxSnapshots, "app.snapshots.max-snapshots", cfg.AppCfg.SnapshotConfig.MaxSnapshots, "Maximum snapshots")
-	// flagSet.StringVar(&cfg.AppCfg.SnapshotConfig.SnapshotDir, "app.snapshots.snapshot-dir", cfg.AppCfg.SnapshotConfig.SnapshotDir, "Snapshot directory path")
+	// Snapshot Config flags
+	flagSet.BoolVar(&cfg.AppCfg.Snapshots.Enabled, "app.snapshots.enabled", cfg.AppCfg.Snapshots.Enabled, "Enable snapshots")
+	flagSet.Uint64Var(&cfg.AppCfg.Snapshots.RecurringHeight, "app.snapshots.recurring-height", cfg.AppCfg.Snapshots.RecurringHeight, "Recurring snapshot height")
+	flagSet.Uint64Var(&cfg.AppCfg.Snapshots.MaxSnapshots, "app.snapshots.max-snapshots", cfg.AppCfg.Snapshots.MaxSnapshots, "Maximum snapshots")
```
jchappelow, line 40: "Maximum snapshots"... to create?  (rotating?)

cmd/kwild/config/flags.go
```diff
@@ -34,16 +34,14 @@ func AddConfigFlags(flagSet *pflag.FlagSet, cfg *KwildConfig) {
...
 
-	// TODO: Snapshots are not supported yet
-	// // Snapshot Config flags
-	// flagSet.BoolVar(&cfg.AppCfg.SnapshotConfig.Enabled, "app.snapshots.enabled", cfg.AppCfg.SnapshotConfig.Enabled, "Enable snapshots")
-	// flagSet.Uint64Var(&cfg.AppCfg.SnapshotConfig.RecurringHeight, "app.snapshots.recurring-height", cfg.AppCfg.SnapshotConfig.RecurringHeight, "Recurring snapshot height")
-	// flagSet.Uint64Var(&cfg.AppCfg.SnapshotConfig.MaxSnapshots, "app.snapshots.max-snapshots", cfg.AppCfg.SnapshotConfig.MaxSnapshots, "Maximum snapshots")
-	// flagSet.StringVar(&cfg.AppCfg.SnapshotConfig.SnapshotDir, "app.snapshots.snapshot-dir", cfg.AppCfg.SnapshotConfig.SnapshotDir, "Snapshot directory path")
+	// Snapshot Config flags
+	flagSet.BoolVar(&cfg.AppCfg.Snapshots.Enabled, "app.snapshots.enabled", cfg.AppCfg.Snapshots.Enabled, "Enable snapshots")
+	flagSet.Uint64Var(&cfg.AppCfg.Snapshots.RecurringHeight, "app.snapshots.recurring-height", cfg.AppCfg.Snapshots.RecurringHeight, "Recurring snapshot height")
```
jchappelow, line 39: `recurring-height` makes sense, but it's `snapshot_heights` for the toml

internal/txapp/txapp.go
```diff
@@ -648,6 +678,47 @@ func (r *TxApp) Commit(ctx context.Context) error {
 
 	r.announceValidators()
 
+	// Take a snapshot of the database if node is not in the catchup mode and snapshots are enabled
+	if r.replayStatusFn != nil && !r.replayStatusFn() &&
+		r.snapshotter != nil && r.snapshotter.IsSnapshotDue(uint64(r.height)) {
```
jchappelow, line 683: Try to order the conditions to exploit short-circuit behavior.  In particular, try to have `r.replayStatusFn()` behind other less expensive checks, especially if the other checks are likely to preclude the need to even check replay status almost all of the time.

jchappelow (changes requested):
Not done, but here's some comment before it gets too late.

cmd/kwil-admin/cmds/setup/testnet.go
```diff
@@ -92,5 +96,8 @@ func testnetCmd() *cobra.Command {
 	cmd.Flags().BoolVar(&withoutNonces, "without-nonces", false, "disable account nonces")
 	cmd.Flags().BoolVar(&withGas, "gas", false, "enable gas")
 	cmd.Flags().Var(&allocs, "alloc", "account=amount pairs of genesis account allocations")
+	cmd.Flags().BoolVar(&snapshotsEnabled, "snapshots.enable", false, "enables db snapshots")
+	cmd.Flags().Uint64Var(&maxSnapshots, "snapshots.max-snapshots", 3, "Maximum number of snapshots to store in the device")
+	cmd.Flags().Uint64Var(&snapshotHeights, "snapshots.heights", 10000, "Heights to snapshot the db")
```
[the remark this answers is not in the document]
jchappelow: Eh, I guess if it's desirable for the flags to match kwild's flags...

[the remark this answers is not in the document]
charithabandi: it will be multiples of these heights. like if the configured height is 10000, we would snapshot at 10k, 20k, 30k etc. I choose this approach so that all the nodes can snapshot deterministically at specific heights. Not sure if there is another good way of specifying this.

cmd/kwil-admin/nodecfg/toml.go
jchappelow: Actually, how does one set these?  I kinda though everything from `rpc_servers` was trusted.

charithabandi: https://docs.cometbft.com/v0.34/core/light-client#where-to-obtain-trusted-height--hash

jchappelow: I see this in cometbft:

```go
// TrustOptions are the trust parameters needed when a new light client
// connects to the network or when an existing light client that has been
// offline for longer than the trusting period connects to the network.
//
// The expectation is the user will get this information from a trusted source
// like a validator, a friend, or a secure website. A more user friendly
// solution with trust tradeoffs is that we establish an https based protocol
// with a default end point that populates this information. Also an on-chain
// registry of roots-of-trust (e.g. on the Cosmos Hub) seems likely in the
// future.
type TrustOptions struct {
	// tp: trusting period.
	//
	// Should be significantly less than the unbonding period (e.g. unbonding
	// period = 3 weeks, trusting period = 2 weeks).
	//
	// More specifically, trusting period + time needed to check headers + time
	// needed to report and punish misbehavior should be less than the unbonding
	// period.
	Period time.Duration

	// Header's Height and Hash must both be provided to force the trusting of a
	// particular header.
	Height int64
	Hash   []byte
}
```

The part "The expectation is the user will get this information from a trusted source like a validator, a friend, or a secure website. A more user friendly  solution with trust tradeoffs is that we establish an https based protocol with a default end point that populates this information" what what I thought rpc_servers did.

cmd/kwild/config/flags.go
charithabandi: Yeah, we purge old ones and update them with the latest.

cmd/kwild/config/flags.go
charithabandi: AH will update that

cmd/kwil-admin/cmds/setup/testnet.go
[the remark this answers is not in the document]
jchappelow: Yeah, that's totally logical.  The doc here is not clear though.

cmd/kwil-admin/nodecfg/toml.go
charithabandi: Screenshot 2024-04-26 at 4 40 08 PM

Umm, but cometbft rejects statesync if these configs are not provided, what the new node does is, it uses these trust options and queries the rpc servers and verifies these options before starting statesync.

jchappelow: I noticed that too. But in practice, how is a kwild operator going to set these?  IT sounds to me that they'd literally be using the same RPC provider to get these `trust_` settings as the host(s) then use for the `rpc_servers` setting.

charithabandi: yup.

cmd/kwild/config/flags.go
jchappelow: Sorry I ask a lot of rhetorical questions when commending on docs or help strings.  I'm pretty sure I see the answer, but docs could be tweaked to spell it out to the naive reader.

cmd/kwil-admin/nodecfg/toml.go
charithabandi: https://medium.com/tendermint/everything-you-need-to-know-about-the-tendermint-light-client-f80d03856f98

Essentially, they do this sequential verification mentioned in this doc, starting from this trust height.

But yeah, they would query the known rpc endpoints, get a block header and populate this information.

cmd/kwild/config/flags.go
charithabandi: Hahah, That's okay, Thanks for pointing these out, I will improve in my documentation.

cmd/kwil-admin/cmds/snapshot/restore.go
charithabandi: This just implies that the input is received completely. if we dont close it the psqlCmd will hang waiting for the input.

cmd/kwil-admin/cmds/snapshot/create.go
brennanjl: Yeah, this is sort've why I was asking about URI. IMO it feels more flexible as a user, but that's just because it is what I have used before.

cmd/kwil-admin/cmds/snapshot/create.go
brennanjl: I'm actually more confused by this then. After looking deeper, it seems like this does not perform the same cleaning and determinism logic that the internal engine does. Is this because it is not making it a network snapshot?

I feel like it might still be important to have determinism here, even if it is not strictly required. I would imagine users creating a new network would likely want to compare their starting snapshots that they are using for genesis.

Obviously getting a bit speculative, but just found it surprising that this was not the same process

cmd/kwil-admin/cmds/snapshot/create.go
jchappelow: Fair enough.   I find URI overly restrictive and requires you to assemble it (knowing all the pieces) if you just want to set user for example.  I see where you're coming from though.  I personally think we should not wire up URI.  Also just about all the `pg_` commands take the pieces with flags, as do most examples for the commands.

internal/abci/abci.go
charithabandi: Not a problem. We just need the snapshot metadata to compare, no need it to be deterministic

cmd/kwil-admin/cmds/snapshot/create.go
charithabandi: The determinism is needed in Statesync because, the snapshot chunks are coming from multiple nodes and we want literally each byte to be same, so that we can coalesce  chunks from different senders and validate them based off the trusted snapshot providers metadata. 

Whereas in Network migrations, there would be one snapshot file, distributed to other nodes along with the genesis file and the genesis hash. So, it doesn't need that sorting ordering that we do in the statesync snapshots. But we sanitize it to remove the commands that are not required in the sql dump.

internal/snapshots/snapshotter.go
```diff
@@ -0,0 +1,444 @@
...
+		return nil, fmt.Errorf("failed to scan the dump file: %w", err)
+	}
+
+	// remove the dump file
+	err = os.Remove(dumpFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to remove dump file: %w", err)
+	}
+
+	hash, err := utils.HashFile(sanitizedDumpFile)
```
jchappelow, line 289: The file isn't closed or guaranteed to be fsynced yet.   You can manually do `outputFile.Sync()` (or manually `Close`) here first.  Otherwise you could first `outputFile.Seek(0, io.SeekStart)` and then use `outputFile` as an `io.Writer` with a hasher.

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -0,0 +1,253 @@
...
+			}
+
+			return pgDump(cmd.Context(), dbName, dbUser, dbPass, dbHost, dbPort, snapshotDir)
+		},
+	}
+
+	cmd.Flags().StringVar(&snapshotDir, "snapdir", "kwild-snaps", "Directory to store the snapshot and hash files")
+	cmd.Flags().StringVar(&dbName, "dbname", "kwild", "Name of the database to snapshot")
+	cmd.Flags().StringVar(&dbUser, "user", "postgres", "User with administrative privileges on the database")
+	cmd.Flags().StringVar(&dbPass, "pass", "", "Password for the database user")
```
brennanjl, line 69: I believe Postgres's default password is password, right?

Also, could you make the full flag `--password`?

internal/statesync/streamer.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (s *Streamer) Close() error {
+	if s.currentChunk != nil {
+		s.currentChunk.Close()
+	}
+
+	return nil
+}
+
+func (s *Streamer) Read(p []byte) (n int, err error) {
```
brennanjl, line 63: Some better documentation on the exported structs and methods here could be helpful.

jchappelow (commented):
A few quick comments on the latest commit, but I'm testing and reading docs still.

internal/statesync/statesync.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// Check if the chunk has already been applied
+	applied, ok := ss.chunks[index]
+	if ok && applied {
+		return false, false, nil
+	}
+
+	// Check if the chunk index is valid
+	if index >= ss.snapshot.ChunkCount {
+		ss.log.Error("Invalid chunk index", zap.Uint32("index", index), zap.Uint32("chunkCount", ss.snapshot.ChunkCount))
```
brennanjl, line 124: can you use hyphenated log info? `chunk-count` instead of `chunkCount`. Just to keep it consistent.

internal/statesync/store.go
```diff
@@ -0,0 +1,300 @@
...
+		header, err := loadSnapshot(headerFile)
+		if err != nil {
+			s.log.Debug("Invalid snapshot header file, ignoring the snapshot", log.String("height", height), log.String("Error", err.Error()))
+		}
+
+		// Ensure that the chunk files exist
+		for i := uint32(0); i < header.ChunkCount; i++ {
+			chunkFile := snapshotChunkFile(s.cfg.SnapshotDir, heightInt, DefaultSnapshotFormat, i)
+			if _, err := os.Open(chunkFile); err != nil { // chunk file doesn't exist
+				s.log.Debug("Invalid snapshot chunk file, ignoring the snapshot", log.String("chunkFile", chunkFile))
```
brennanjl, line 258: `chunk-file`

brennanjl:
Just ran statesync for the first time, and ran into something unexpected: I specified `--app.snapshots.snapshot-dir=./snaps`, which makes me think that it should make the directory in `working-dir/snaps`. Instead, it made it in my kwild root dir.

I think it should probably go where the user specifies, since I am not sure if it is super valuable that the user can configure where it goes only inside of `.kwild`

brennanjl:
I got a weird error. When trying to connect to a node that has snapshots, I run:
```
kwild --root-dir ~/.kwild2 --chain.statesync.enable=true --chain.statesync.snapshot-dir=snaps --chain.statesync.rpc-servers=http://localhost:26657,http://localhost:26657
```

And I get:
```
{"level":"error","ts":1715107586.393905,"logger":"kwild.cometbft","caller":"service/service.go:170","msg":"Not stopping State service -- has not been started yet","module":"consensus","impl":"ConsensusState","goversion":"go1.21.0","mod_version":"(devel)","revision":"7d1b891f77f82183f2040f3adcefb3e8ef8db87c","build_time":1714667038,"stacktrace":"github.com/cometbft/cometbft/libs/service.(*BaseService).Stop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/libs/service/service.go:170\ngithub.com/cometbft/cometbft/consensus.(*Reactor).OnStop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/consensus/reactor.go:97\ngithub.com/cometbft/cometbft/libs/service.(*BaseService).Stop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/libs/service/service.go:181\ngithub.com/cometbft/cometbft/p2p.(*Switch).OnStop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/p2p/switch.go:259\ngithub.com/cometbft/cometbft/libs/service.(*BaseService).Stop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/libs/service/service.go:181\ngithub.com/cometbft/cometbft/node.(*Node).OnStop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/node/node.go:590\ngithub.com/cometbft/cometbft/libs/service.(*BaseService).Stop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/libs/service/service.go:181\ngithub.com/kwilteam/kwil-db/internal/abci/cometbft.(*CometBftNode).Stop\n\t/Users/brennanlamey/kwil-db/internal/abci/cometbft/node.go:216\ngithub.com/kwilteam/kwil-db/cmd/kwild/server.(*Server).Start.func7\n\t/Users/brennanlamey/kwil-db/cmd/kwild/server/server.go:203\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/Users/brennanlamey/go/pkg/mod/golang.org/x/sync@v0.6.0/errgroup/errgroup.go:78"}
{"level":"error","ts":1715107586.393999,"logger":"kwild.cometbft","caller":"consensus/reactor.go:98","msg":"Error stopping consensus state","module":"consensus","err":"not started","goversion":"go1.21.0","mod_version":"(devel)","revision":"7d1b891f77f82183f2040f3adcefb3e8ef8db87c","build_time":1714667038,"stacktrace":"github.com/cometbft/cometbft/consensus.(*Reactor).OnStop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/consensus/reactor.go:98\ngithub.com/cometbft/cometbft/libs/service.(*BaseService).Stop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/libs/service/service.go:181\ngithub.com/cometbft/cometbft/p2p.(*Switch).OnStop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/p2p/switch.go:259\ngithub.com/cometbft/cometbft/libs/service.(*BaseService).Stop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/libs/service/service.go:181\ngithub.com/cometbft/cometbft/node.(*Node).OnStop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/node/node.go:590\ngithub.com/cometbft/cometbft/libs/service.(*BaseService).Stop\n\t/Users/brennanlamey/go/pkg/mod/github.com/cometbft/cometbft@v0.38.6/libs/service/service.go:181\ngithub.com/kwilteam/kwil-db/internal/abci/cometbft.(*CometBftNode).Stop\n\t/Users/brennanlamey/kwil-db/internal/abci/cometbft/node.go:216\ngithub.com/kwilteam/kwil-db/cmd/kwild/server.(*Server).Start.func7\n\t/Users/brennanlamey/kwil-db/cmd/kwild/server/server.go:203\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/Users/brennanlamey/go/pkg/mod/golang.org/x/sync@v0.6.0/errgroup/errgroup.go:78"}
```

Could this be due to connecting to the same snapshot rpc?

brennanjl:
> I got a weird error. When trying to connect to a node that has snapshots

Regardless of the reason, I think we need a better error message here.

brennanjl:
Ah, it appears changing my target rpc servers from `http://localhost:26657` to `http://0.0.0.0:26657` works. It also works fine if I specify the same snapshot provider twice.

@charithabandi, could you make it so that if only one snapshot provider is provided, it duplicates it so that cometbft can work?

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -139,10 +141,7 @@ func pgDump(ctx context.Context, dbName, dbUser, dbPass, dbHost, dbPort, snapsho
 		"--port", dbPort,
 		"--no-password",
 	)
-
-	if dbPass != "" {
-		pgDumpCmd.Env = append(os.Environ(), "PGPASSWORD="+dbPass)
-	}
+	pgDumpCmd.Env = append(os.Environ(), "PGPASSWORD="+dbPass)
```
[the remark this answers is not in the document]
charithabandi: I think with -W there is a concern of passwords being displayed in the logs afair.

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -66,9 +68,9 @@ func createCmd() *cobra.Command {
 	cmd.Flags().StringVar(&snapshotDir, "snapdir", "kwild-snaps", "Directory to store the snapshot and hash files")
 	cmd.Flags().StringVar(&dbName, "dbname", "kwild", "Name of the database to snapshot")
 	cmd.Flags().StringVar(&dbUser, "user", "postgres", "User with administrative privileges on the database")
-	cmd.Flags().StringVar(&dbPass, "pass", "", "Password for the database user")
+	cmd.Flags().StringVar(&dbPass, "password", "password", "Password for the database user")
```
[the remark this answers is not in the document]
charithabandi: Yeah, that's why I had no defaults set. I am gonna keep it that way to support trust auth and unix socket cases.

cmd/kwild/config/config.go
```diff
@@ -167,11 +170,28 @@ type ConsensusConfig struct {
...
+	// Time to spend discovering snapshots before initiating starting
+	// the db restoration using snapshot.
+	DiscoveryTime Duration `mapstructure:"discovery_time"`
+
+	// The timeout duration before re-requesting a chunk, possibly from a different
+	// peer (default: 1 minute).
 	ChunkRequestTimeout Duration `mapstructure:"chunk_request_timeout"`
+
+	// Light client verification options, Automatically fetched from the RPC Servers
+	// during the node initialization.
```
jchappelow, line 191: "Automatically fetched" means these shouldn't be in config.toml, right?

cmd/kwild/server/build.go
```diff
@@ -450,31 +540,92 @@ func initAccountRepository(d *coreDependencies, tx sql.Tx) {
...
+
+	configDone := false
+	for _, p := range providers {
+		clt, err := statesync.ChainRPCClient(p)
+		if err != nil {
+			continue
+		}
+
+		// Try to fetch the status of the remote server.
+		res, err := clt.Commit(d.ctx, nil)
```
jchappelow, line 605: `clt.Header` would be sufficient.  We don't need the commit evidence, only the header (`Hash()` method give the hash)

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// Logger
+	log log.Logger
+}
+
+// NewStateSyncer will initialize the state syncer that enables the node to
+// receive and validate snapshots from the network and initialize the database state.
+// It takes the database configuration, snapshot directory, and the trusted snapshot providers.
+// Trusted snapshot providers are special nodes in the network trusted by the nodes and
+// have snashopt creation enabled. These nodes are responsible for creating and validating snapshots.
```
jchappelow, line 57: snashopt => snapshot

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+		clt, err := ChainRPCClient(s)
+		if err != nil {
+			logger.Error("Failed to create rpc client", log.String("server", s), log.Error(err))
+			return nil
+		}
+		ss.snapshotProviders = append(ss.snapshotProviders, clt)
+	}
+
+	// Ensure that the snasphot directory exists and is empty
+	if err := os.RemoveAll(snapshotDir); err != nil {
```
jchappelow, line 76: I didn't pick up on this before.  Why are we deleting any existing snapshots on start?  Doesn't that hurt the overall snapshot availability for the network as a whole?

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+		logger.Error("Failed to create snapshot directory", log.String("dir", snapshotDir), log.Error(err))
+		return nil
+	}
+
+	return ss
+}
+
+// OfferSnapshot checks if the snapshot is valid and kicks off the state sync process
+// accepted snapshot is stored on disk for later processing
+func (ss *StateSyncer) OfferSnapshot(snapshot *Snapshot) error {
```
jchappelow, line 90: Let's pass the context from abci into this method and then into `validateSnapshot`

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+	err = os.WriteFile(chunkFileName, chunk, 0755)
+	if err != nil {
+		os.Remove(chunkFileName)
+		return false, false, ErrRetrySnapshotChunk
+	}
+
+	ss.log.Info("Applied snapshot chunk", log.Uint("height", ss.snapshot.Height), log.Uint("index", index))
+
+	ss.chunks[index] = true
+	ss.rcvdChunks += 1
```
jchappelow, line 153: `ss.rcvdChunks++`

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	// Validate the snapshot
+	err := ss.validateSnapshot(*snapshot)
+	if err != nil {
+		return err
+	}
+
+	ss.snapshot = snapshot
+	ss.inProgress = true
```
jchappelow, line 105: The possible concurrency of calls to `OfferSnapshot` and `ApplySnapshotChunk`, or multiple concurrent calls to either one is unclear to me.

Do we know that one call to `OfferSnapshot` must return before it can be called again?

The concerns are with a number of fields including `inProgress` and `chunks`.

Maybe it's not a concern at all.  Depends on how these methods are used by cometbft.

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+// ApplySnapshotChunk accepts a chunk and stores it on disk for later processing if valid
+// If all chunks are received, it starts the process of restoring the database
+func (ss *StateSyncer) ApplySnapshotChunk(ctx context.Context, chunk []byte, index uint32) (bool, bool, error) {
+	if !ss.inProgress {
+		return false, false, ErrStateSyncNotInProgress
+	}
+
+	// Check if the chunk has already been applied
+	applied, ok := ss.chunks[index]
+	if ok && applied {
```
jchappelow, line 120: Just `if ss.chunks[index] {` is preferable.  Non-existent entries return the zero value, which is `false`

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// Check if the chunk index is valid
+	if index >= ss.snapshot.ChunkCount {
+		ss.log.Error("Invalid chunk index", log.Uint("index", index), log.Uint("chunk-count", ss.snapshot.ChunkCount))
+		return false, false, ErrRejectSnapshotChunk
+	}
+
+	// Validate the chunk hash
+	chunkHash := sha256.Sum256(chunk)
+	if !bytes.Equal(chunkHash[:], ss.snapshot.ChunkHashes[index][:]) {
```
jchappelow, line 132: Arrays are `comparable`.  This was one of the benefits of converting the  hashes to arrays in the last iteration.  Just `if chunkHash != ss.snapshot.ChunkHashes[index] {` works.

jchappelow (changes requested):
Code I missed last time

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+		return false, true, ErrRejectSnapshotChunk
+	}
+
+	// store the chunk on disk
+	chunkFileName := filepath.Join(ss.snapshotsDir, fmt.Sprintf("chunk-%d.sql.gz", index))
+	chunkFile, err := os.Create(chunkFileName)
```
jchappelow, line 138: I think this isn't used?

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+	height := fmt.Sprintf("%d", snapshot.Height)
+	verified := false
+	for _, clt := range ss.snapshotProviders {
+		res, err := clt.ABCIQuery(context.Background(), ABCISnapshotQueryPath, []byte(height))
+		if err != nil {
+			ss.log.Info("Failed to query snapshot", log.Error(err)) // failover to next provider
+			continue
+		}
+
+		if res.Response.Value != nil {
```
jchappelow, line 270: I think it would be safer with `len(res.Response.Value) > 0`.  An empty non-nil slice might get past `json.Unmarshal` and declare an invalid snapshot instead of continuing.

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+			err = json.Unmarshal(res.Response.Value, &snap)
+			if err != nil {
+				ss.log.Error("Failed to unmarshal snapshot", log.Error(err))
+				continue
+			}
+
+			if snap.Height != snapshot.Height || snap.SnapshotSize != snapshot.SnapshotSize ||
+				snap.ChunkCount != snapshot.ChunkCount || !bytes.Equal(snap.SnapshotHash, snapshot.SnapshotHash) {
+				ss.log.Error("Invalid snapshot", log.Uint("height", snapshot.Height), log.Any("Expected ", snap), log.Any("Actual", snapshot))
+				break
```
jchappelow, line 281: Also, should an invalid snapshot from one provider cause us to break and not try the next provider?

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+		return ErrInvalidSnapshot
+	}
+
+	return nil
+}
+
+func (ss *StateSyncer) resetStateSync() {
+	ss.inProgress = false
+	ss.snapshot = nil
+	ss.chunks = nil
```
jchappelow, line 299: Is `inProgress` redundant with `snapshots` and/or `chunks` fields?  i.e. does `ss.snapshot != nil` imply in-progress?

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+	// Check if the chunk index is valid
+	if index >= ss.snapshot.ChunkCount {
+		ss.log.Error("Invalid chunk index", log.Uint("index", index), log.Uint("chunk-count", ss.snapshot.ChunkCount))
+		return false, false, ErrRejectSnapshotChunk
+	}
+
+	// Validate the chunk hash
+	chunkHash := sha256.Sum256(chunk)
+	if !bytes.Equal(chunkHash[:], ss.snapshot.ChunkHashes[index][:]) {
+		return false, true, ErrRejectSnapshotChunk
```
jchappelow, line 133: If you want to retry, shouldn't this error be `ErrRetrySnapshotChunk`? 
Other thing is that it's an anti-pattern to interpret any other returns when the error is non-nil.  You seem to have set things up quite well with `ToABCIApplySnapshotChunkResponse` to solve this with just the error.  You should be able to get rid of the `refetch` return.

internal/statesync/statesync.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	// store the chunk on disk
+	chunkFileName := filepath.Join(ss.snapshotsDir, fmt.Sprintf("chunk-%d.sql.gz", index))
+	chunkFile, err := os.Create(chunkFileName)
+	if err != nil {
+		return false, false, ErrRetrySnapshotChunk
```
jchappelow, line 140: I'm confused by this return too.  retry is `false`, but the error is a retry error.
If possible, let's try to eliminate the second bool return and communicate the response with just the error type.

internal/txapp/txapp.go
```diff
@@ -205,6 +206,29 @@ func (r *TxApp) ChainInfo(ctx context.Context) (int64, []byte, error) {
 	return height, appHash, nil
 }
 
+// ReloadCache reloads the database state into the engine.
+func (r *TxApp) ReloadCache(ctx context.Context) error {
```
jchappelow, line 210: I'm personally confused by the name `ReloadCache`.  The fields of TxApp and Engine that are now incorrect and need updating are a lot more than "cache" IMO.  They're just the working state of the application.  I guess let this one alone, but `Reload` seemed fine to me.  "Cache" refers to some low-latency storage of data for performance optimization, where entries can expire or be evicted, and where cache misses result in more costly retrieval from storage etc.  That's not analogous to what we're doing here.  We are refreshing indispensable application state.  e.g. We could achieve a similar effect by creating a new *instance* of Engine via it's constructor where it loads datasets.  Or a more broad reinstantiation of everything, but that's just harder to engineer.

internal/statesync/statesync.go
charithabandi: Yeah right, we trust all the providers equally, and our snapshots are deterministic, So we expect all providers to return same info.

jchappelow: Meaning, we should have this `break` if one of them returns incorrect data, or it should be a `continue`?

internal/statesync/statesync.go
charithabandi: Yeah, I decided not use these snapshots used for statesync. Especially because this node might not have snapshots enabled and it makes management of snapshot store easy. also this directory is different from the snapshot direction u use if snapshot creation is enabled. This directory just stores chunks received during statesync and deletes them after restore. If incase snapshot creation is enabled, it would make its own chunks at later block heights.

cmd/kwild/server/build.go
```diff
@@ -450,31 +540,92 @@ func initAccountRepository(d *coreDependencies, tx sql.Tx) {
...
+	providers := strings.Split(d.cfg.ChainCfg.StateSync.RPCServers, ",")
+
+	if len(providers) == 0 {
+		failBuild(nil, "failed to configure state syncer, no remote servers provided.")
+	}
+
+	if len(providers) == 1 {
+		d.log.Warn("multiple state sync providers are not supported yet, using the first one")
+		// Duplicating the same provider to satisfy cometbft statesync requirement of having at least 2 providers.
+		d.cfg.ChainCfg.StateSync.RPCServers += "," + providers[0]
```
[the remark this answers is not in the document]
charithabandi: Yeah, this is just to satisfy the cometbft statesync requirements. Whereas Statesyncer's snapshot providers can do with just one provider, we use these to validate snapshots, it wouldn't make sense to store this server multiple times in the statesync module for us.

internal/statesync/statesync.go
jchappelow: > this directory is different from the snapshot direction u use if snapshot creation is enabled. This directory just stores chunks received during statesync and deletes them after restore. If incase snapshot creation is enabled, it would make its own chunks at later block heights.

ahhhhhhhh

internal/statesync/statesync.go
charithabandi: These are not concurrently called at all.

internal/statesync/statesync.go
charithabandi: Yeah, true. They both are kinda same, i will remove in-progress

jchappelow:
Any edits to the following proposed commit message @charithabandi ?

```
statesync,multi: enable snapshots and state sync, plus genesis data

This enables the support to create and manage logical Postgres snapshots
and accelerate the catch-up process of new node using statesync.

## Snapshot store

Snapshot store in the snapshots module is responsible for creating and
maintaining snapshots. Snapshots are the logical sql dump of the
Postgres database and is stored on the device in chunks of max size 16MB
(imposed by the cometBft statesync). Snapshot directory has to follow
this strict directory pattern
`snapshot-dir/block-<number>/format-0/chunks/`. Snapshot header stored
under `snapshot-dir/block-<number>/format-0/header.json` would represent
the snapshot metadata such as the snapshot height, hash, number of
chunks, each chunk hash etc. Snapshot store also ensures that the
snapshots are taken at correct heights and maintains only max-snapshots
on disk at any point in time.

### Snapshot creation

- Snapshot transaction with Serializable isolation level is created at
  the end of the block commit
- Snapshot ID is created to track the DB state at the end of the block
- `pg_dump` utility is used for creating snapshots at that specific
  snapshot ID.
- To make these dumps deterministic, this process goes through extra
   sanitization to order all the rows in the tables, to remove
   SET/SELECT statements which are not required, and white spaces are
   removed to reduce the space.
- Sanitized snapshots are further compressed to minimize the space usage
- Compressed snapshots are divided into chunks of max size 16MB.

## StateSync

Having the snapshots enables us to bootstrap the new node's catchup
process rapidly using the CometBFT's statesync feature. Requirements:
Trusted Snapshot providers:  Network should have a bunch of full nodes
that can provide and validate block headers and snapshots.

New node that is joining, requests the networks for snapshots. All the
peers respond to this with the metadata corresponding to the  list of
snapshots that they currently have. Once the new node validate the
snapshot with the trusted snapshot provider's snapshot at the same
height, the new node requests for the snapshot chunks.

Statesync module is responsible for validating the snapshots, keeping
track of the chunks received, decompressing and restoring the DB state.
Once the DB is restored and the app hash is validated, if there are more
blocks to sync, Cometbft would switch to the block sync mode. This new
node caught up using the statesync would have truncated block history
with starting height at the snapshot height.

## Docker builds

Kwild Docker images are now shipped with the  the` postgresql-client`
tools to run `pg-dump` and `psql`

Spammer tests are run with the snapshots enabled for performance tests

## New additions

- Trust options for Statesync are automated to enhance the user
  experience.

- Kwil-admin snapshot create command is added to  create a db snapshot
  for genesis state during network migrations. This command also spews
  out the genesis file with validators and genesis hash defined based on
  the previous network. Other params can be updated manually by the
  network operators.

- Network migrations are enforced through `snapshot_file` in
  `config.toml` and  `app_hash` in the `genesis.json`. If the app hash
  is present in the genesis file, it implies that the node is supposed
  to start with an initial state and the initial state can be loaded
  using the snapshot file. Also the hash of the uncompressed logical
  dump should match that of the genesis hash ensuring that the node is
  starting with the correct state. Also reinforces that all nodes start
  with same state. However, the db is restored from the snapshot only if
  the node is starting fresh from genesis height. DB restoration is
  skipped if the node is not at genesis height or if statesync is
  enabled or if genesis app hash is not specified.

- Removed dependencies between statesync and snapshots packages.
```

jchappelow:
I've just taken that from PR desc.  Probably several things out of date.  Might be missing some of the kwil-admin changes.

charithabandi:
> I've just taken that from PR desc. Probably several things out of date. Might be missing some of the kwil-admin changes.

Updated the commit message to reflect latest changes

jchappelow (approved):

PR #630, opened 2024-04-03
test: ganache => hardhat

jchappelow:
This replaces [Ganache](https://github.com/trufflesuite/ganache) with [Hardhat](https://github.com/NomicFoundation/hardhat).

Ganache project is now dead, and the advice is to migrate to hardhat: https://consensys.io/blog/consensys-announces-the-sunset-of-truffle-and-ganache-and-new-hardhat

Several things were done to accomplish this:
- create a hardhat docker image like the one we were using for ganache: [our `kwildb/hardhat` Dockerfile repo](https://github.com/kwilteam/hardhat) and [pushed to Hub](https://hub.docker.com/r/kwildb/hardhat/tags)
- <s>`eth_maxPriorityFeePerGas` is not supported by Hardhat (it's a go-ethereum invention, not in the RPC spec): https://github.com/ethereum/go-ethereum/issues/23479.  We must set the dynamic transaction fee caps, otherwise the go-ethereum methods try to use `eth_maxPriorityFeePerGas` to get the "tip" cap with this non-existent RPC</s>
- <s>contract deploy fails if we don't give it a gas limit set because it tries fee estimation for the deployment transaction and errors with "eth_estimateGas: Contract creation without any data provided".  I don't know the cause of this bug, but we can work around it.</s>
- The compose health check waits for the log "Started HTTP and WebSocket JSON-RPC server" in the hardhat docker container.

<s>NOTE: back to draft because there are too many hoops to jump through until hardhat is updated for it's `eth_call` handling. See https://github.com/kwilteam/kwil-db/pull/630#issuecomment-2037353403</s>
The v2.22.5 release fixed the issues: https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.22.5


jchappelow:
Changing to draft since this one is proving painful to fully resolve until there is resolution to https://github.com/NomicFoundation/edr/issues/360, such as https://github.com/NomicFoundation/hardhat/pull/4718.  In short, hardhat does not recognize `input` as the field for parameters in `eth_call` and some other RPCs although that is to-spec, which is understandable since `data` was the field used broadly and in go-ethereum prior to 1.13.

brennanjl (approved):
Approving but I think @Yaiba should take a look since he knows more on this

jchappelow:
We should make this change before we make the release branch.  Our hardhat dockerfile is here: https://github.com/kwilteam/hardhat

test/integration/docker-compose.yml.template
```diff
@@ -454,19 +454,12 @@ services:
...
+    image: kwildb/hardhat:latest
     ports:
       - "8545"
     networks:
       - {{ .Network }}
-    command:
-      - --wallet.hdPath
-      - m/44'/60'/0'
-      - --wallet.mnemonic
-      - test test test test test test test test test test test junk
```
Yaiba, line 467: interesting the tests pass without setting the mnemonic, as we have _deployer_ that need funds, it's configured to use this wallet in `KIT_CREATOR_PK`

jchappelow: Yeah, this `test test test test test test test test test test test junk` wallet is the default.  https://github.com/kwilteam/hardhat/blob/main/hardhat.config.js
We only tweak the derivation path a bit. https://hardhat.org/hardhat-network/docs/reference#accounts

Yaiba: oh, where do we change the hdPath?

jchappelow: In that linked hard hat config in our own docker image: https://github.com/kwilteam/hardhat/blob/main/hardhat.config.js
We can also give a cli flag to a config file, but we control the docker image so we can set the defaults right.

Yaiba: ah sorry, I just totally ignored that config

PR #634, opened 2024-04-04
kwil-admin: fix minor typo

KwilLuke:
fix minor typo for `validators list-validators` command

### DIFF
--- a/cmd/kwil-admin/cmds/validators/list.go
+++ b/cmd/kwil-admin/cmds/validators/list.go
@@ -16,7 +16,7 @@ var (
 	listLong = `List the current validator set of the network.`
 
 	listExample = `# List the current validator set of the network
-kwil-admin validators list-validators`
+kwil-admin validators list`
 )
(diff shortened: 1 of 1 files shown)


cmd/kwil-admin/cmds/validators/list.go
```diff
@@ -16,7 +16,7 @@ var (
 	listLong = `List the current validator set of the network.`
 
 	listExample = `# List the current validator set of the network
-kwild validators list`
+kwil-admin validators list-validators`
```
jchappelow, line 19: Ah, `kwild` here is totally a typo, but we have since aliased `list-validators` to `list` since it's already a subcommand of `validators` command.  Can you verify?

KwilLuke: Ah right, it looks like both commands work. Will update to use the alias.

jchappelow (approved):

PR #631, opened 2024-04-03
CI/fix release & trigger workflow

Yaiba:
This pr fix couple CI issue:
- version parsing erro in release.yaml
- try checkout to pr branch using sha instead of branch, bcz it's not available in kwilteam repo if pr comes from forked repo

I used https://github.com/nektos/act to test locally, which I really recommend, it could save a lot of time

PR #626, opened 2024-04-02
[WIP] CI: try kgw test on main branch update

Yaiba:
As there are:
- permission issue running kgw test in PRs from forked branch
- privacy/security issue since kgw is private repo

Running kgw test on main branch seems a way to do it, as by the time it's merged to main, it's reviewed. And also (seems) no permission issue.

The `main` workflow tries to load the kwild image from cache built from `ci` workflow, then build kgw image and run only the kgw integration tests.

I'm doing experimenting rn. Not ready for review yet.

PR #640, opened 2024-04-05
[v0.7.2] ci: test kgw in release workflow

Yaiba:
This backports KGW test in release workflow(#626 and #631), so v0.7 release will trigger KGW test.

From https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release, it will checkout to `Tag ref of release refs/tags/<tag_name>`, so we need this workflow on release-v.07

### DIFF
--- a/.github/workflows/kgw-test-reuse.yaml
+++ b/.github/workflows/kgw-test-reuse.yaml
@@ -1,0 +1,179 @@
+name: kgw-test-reuse
+
+on:
+  workflow_call:
+    inputs:
+      # for logging purpose
+      kdb-repo:
+        type: string
+      # for workflow_run trigger, we need to check out to the sha, bcz most of
+      # the pr come from forked repo, and we can't access the branch, but we can
+      # access the sha(i.e, pull/xxx/head)
+      kdb-sha:
+        type: string
+      kdb-branch:
+        required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/kgw-test-reuse.yaml
```diff
@@ -0,0 +1,179 @@
...
+      kdb-repo:
+        type: string
+      # for workflow_run trigger, we need to check out to the sha, bcz most of
+      # the pr come from forked repo, and we can't access the branch, but we can
+      # access the sha(i.e, pull/xxx/head)
+      kdb-sha:
+        type: string
+      kdb-branch:
+        required: true
+        type: string
```
jchappelow, line 16: Is this current with main's?  `required` is true for branch one but can't it work with sha?  Otherwise PR looks good to me.

jchappelow (approved):

.github/workflows/kgw-test-reuse.yaml
jchappelow: Oh, it is current.  I'm thinking of manual kgw test....

Yaiba:
Ok, it works https://github.com/kwilteam/kwil-db/actions/runs/8572976999

PR #641, opened 2024-04-05
[EXPERIMENT] retry in WrapClient if error is connect or net failure rather than app error

jchappelow:
We have frequent errors from `chain_info` when doing `core/client.WrapClient` to establish the initial connection.  This is a test to see if we can catch some of these transient errors that probably come from trying to connect too soon.

DO NOT MERGE

### DIFF
--- a/cmd/kwil-cli/cmds/common/roundtripper.go
+++ b/cmd/kwil-cli/cmds/common/roundtripper.go
@@ -41,7 +41,7 @@ func DialClient(ctx context.Context, cmd *cobra.Command, flags uint8, fn RoundTr
 
 	needPrivateKey := flags&WithoutPrivateKey == 0
 
-	clientConfig := clientType.Options{}
+	clientConfig := clientType.Options{} // note for debugging: Logger: log.NewStdOut(log.DebugLevel)
 	if conf.PrivateKey != nil {
 		clientConfig.Signer = &auth.EthPersonalSigner{Key: *conf.PrivateKey}
 		if needPrivateKey { // only check chain ID if signing something
--- a/core/client/client.go
+++ b/core/client/client.go
@@ -109,6 +109,7 @@ CHAIN_INFO:
 			case <-ctx.Done():
 				return nil, ctx.Err()
 			}
+			c.logger.Info("retrying chain_info")
 			goto CHAIN_INFO
 		}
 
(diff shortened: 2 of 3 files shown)


jchappelow:
This actually seems to work, but doesn't feel great to add a retry with a heuristic in client just for CI.  @Yaiba What other solution can we try or what might make this client change less of place?  Maybe a `WithAllowConnectRetry`  option so it's not normal operation?  CI failures from chain_info network failures are intolerable some days.

Yaiba:
> This actually seems to work, but doesn't feel great to add a retry with a heuristic in client just for CI. @Yaiba What other solution can we try or what might make this client change less of place? Maybe a `WithAllowConnectRetry` option so it's not normal operation? CI failures from chain_info network failures are intolerable some days.

Yeah agree. Plus if we do want to enable retry, then retry logic should applied for all request, not just for chain_info.

The solution I'm thinking is something like `timedClient` in stress tools , which only affects tests, but seems won't work on cli driver without making more changes.

`WithAllowConnnectRetry` seems like a solid approach.

brennanjl:
I do agree with `WithAllowConnectRetry`, or some other auto `Retry` option that can be applied to all rpc calls. It seems easier to stomach than just something for CI

jchappelow:
I'm gonna rework this with an option today.  CI is so flakey, again.

jchappelow:
> Yeah agree. Plus if we do want to enable retry, then retry logic should applied for all request, not just for chain_info.

This I don't agree with.  To clarify, this PR does not apply it to an explicit chain_info request.  What it does is apply it in the constructor around the "handshake" logic, which happens to use chain_info.

Say an application makes a Client instance, do we really want every single RPC to have a retry?  I only think we want this for the handshake.  I think it's actually fairly dangerous to retry RPCs that involve a transaction or signed message.

Yaiba:
> This I don't agree with. To clarify, this PR does not apply it to an explicit chain_info request. What it does is apply it in the constructor around the "handshake" logic, which happens to use chain_info.

I wasn't thinking through, forget we reused the connection. So yeah more accurately, retry on connection/handshake.

> The solution I'm thinking is something like timedClient in stress tools , which only affects tests, but seems won't work on cli driver without making more changes.

So this is not the way to go.

jchappelow:
I don't really feel like adding cli flags for this.  Giving up on it.  Open to other CI minded resolutions.

PR #645, opened 2024-04-09
http: strip trailing slashes from "BasePath"

jchappelow:
The generated Swagger http clients require a URL to which an method's endpoint path is appended.  This is a problem if this base path has a trailing slash since the endpoint paths begin with a slash. The result is that kwild's server will report 404 not found, while kgw would try to clean the path but redirect with GET even for POST.

This change fixes our http client constructors to trim trailing slashes from the base URL.


Yaiba (approved):
LGTM

PR #647, opened 2024-04-09
gatewayclient: fix domain compare when match cookie

Yaiba:
kwilteam/kgw#18 made changes to what's get set as domain in cookie, kwil-cli didn't change accordingly. 

So when call `.build/kwil-cli database call --action owner_only --name testdb --kwil-provider http://localhost:8090 --authenticate` multiple times, got prompted every time. 
The issue is that after cookie is loaded, we compared the cookie domain and `--kwil-provider`, it won't match if provider has `:port`

This fixes it.

This should be backported to v0.2 and v0.1


core/gatewayclient/client.go
```diff
@@ -189,7 +190,11 @@ func (c *GatewayClient) GetAuthCookie() (cookie *http.Cookie, found bool) {
 // If the cookie is not valid for the client target, it returns an error.
 // It will overwrite any existing authentication cookie.
 func (c *GatewayClient) SetAuthCookie(cookie *http.Cookie) error {
-	if cookie.Domain != "" && cookie.Domain != c.target.Host {
+	// ref https://stackoverflow.com/a/16328399.
+	// KGW already set the cookie domain without port in
+	// https://github.com/kwilteam/kgw/pull/18/files#diff-5b365c916e8b28d0115f136435a03daa1ef6df8cf0eb49479c556923545b56c7R60
+	targetDomain := strings.Split(c.target.Host, ":")[0]
+	if cookie.Domain != "" && cookie.Domain != targetDomain {
```
jchappelow, line 197: Yeah, good catch.
Fix seems fine.  Only call SetAuthCookie once client use in most cases, so no strong need to store this `targetDomain` in a field.  Also could/should use `strings.Cut` (see docs for `strings.Split`) but just a nit.

jchappelow (approved):
LGTM.  Untested though.

core/gatewayclient/client.go
Yaiba: hmm, if `strings.Cut` then I think we'll need `targetDomain`? As it returns multi values

Or `if cookie.Domain != "" && cookie.Domain != strings.Split(c.target.Host, ":")[0]` ?

jchappelow: It's good as-is

PR #662, opened 2024-04-19
core/client: add example app and developer guide

jchappelow:
This change:
- adds a README to core/client that provides a walkthrough for developers. See rendered markdown [here](https://github.com/kwilteam/kwil-db/blob/02577bde53868d8dd9293b083eb5b9ab71cabbcd/core/client/README.md).
- adds a complete but minimal example app demonstrating use of `core/client.Client` (and related packages)
- updates lacking error / error code handling in txapp, abci, and transactions


core/client/example/go.mod
```diff
@@ -0,0 +1,50 @@
+module mykwilapp
+
+go 1.22.1
+
+require (
+	// this kuneiform rev uses the update sqlparser on kwil-db's main:
+	github.com/kwilteam/kuneiform v0.6.1-0.20240418203939-18f750d293bf
+	github.com/kwilteam/kwil-db/core v0.1.2
```
jchappelow, line 8: These two requires for a basic app got me thinking about the `parse` module again.  If an app has to import the _main_ module as well as the `core` module to work with .kf files, that could be unfortunate for an app developer.

brennanjl: I'm not opposed to that, however what is the difference between a developer importing `core` and `main` vs `core` and `parse`?

core/client/README.md
```diff
@@ -0,0 +1,364 @@
...
+action tag($msg) public {
+    INSERT INTO "tags" (ident, msg) VALUES (@caller, $msg);
+}
+```
+
+We called the `"tag"` action with arguments `[][]any{{"jon was here"}}`. This is
+a `[][]any` to support batched action execution. In this example, we execute the
+action once, using `"jon was here"` as the `$msg` argument.
+
+NOTE: To execute an action with no input arguments, provide `nil`.
```
MicBun, line 323: > batched action execution

could you also provide an example of this `batched action execution`?

> using `"jon was here"` as the `$msg` argument.

So, I assume, that if there were more than one argument, would it always be in order? i.e.:
```
action tag($msg,$msg1)
```
will have
```
args := [][]any{{"jon was here"},{"ben was here"}}
```
where
msg1 -> `jon was here`
msg2 -> `ben was here`

core/client/README.md
```diff
@@ -0,0 +1,364 @@
...
+	"github.com/kwilteam/kwil-db/core/crypto/auth"
+)
+
+func genKey() *crypto.Secp256k1PrivateKey {
+	key, _ := crypto.GenerateSecp256k1Key()
+	return key // fmt.Println(key.Hex())
+}
+
+func makeSigner(keyHex string) auth.Signer {
+	key, err := crypto.Secp256k1PrivateKeyFromHex(testKeyHex)
```
MicBun, line 118:
```suggestion
	key, err := crypto.Secp256k1PrivateKeyFromHex(keyHex)
```

core/client/README.md
```diff
@@ -0,0 +1,364 @@
+# Kwil Go Client
+
+This folder contains the Go language client for interacting with a Kwil RPC
+provider. Package `client` may be used to build a third-party application with
+the ability to:
+
+- Retrieve the status of a Kwil network.
+- List and retrieve Kuneiform schema deployed on a Kwil network.
+- Deploy and drop Kwil schema.
+- Execute mutative actions defined in a scheme.
```
brennanjl, line 10: nit: schema, instead of scheme

core/client/example/go.mod
jchappelow: There are some considerations I hadn't really though about clearly until doing this:

- Independent versioning is more important here since the main module version tags refer to our product version, which is not a package.  A package version following semantic versioning is more important.
- sequestering irrelevant module dependencies is a "good thing".  e.g. writing an app that uses kwil that has postgres and cometbft packages in it's go.mod is confusing and potentially frustrating

These were reasons when we made the  `parse` module initially, but we also had the kuneiform circular require to defeat.

I think it would be better to have `parse` module with `parse/kuneiform` packages.

core/client/README.md
```diff
@@ -0,0 +1,364 @@
...
+
+func main() {
+	ctx := context.Background()
+	signer := makeSigner(privKey)
+	acctID := signer.Identity()
+
+	opts := &ctypes.Options{
+		Logger:  klog.NewStdOut(klog.InfoLevel),
+		ChainID: chainID, // ensure the provider matches
+		Signer:  signer,  // required for transactions and auth
```
KwilLuke, line 143: Does the signature type need to be specified anywhere? Is this the same process for Ed25519 / or other custom signers?

core/client/README.md
```diff
@@ -0,0 +1,364 @@
...
+records, err := cl.CallAction(ctx, dbid, "get_all", nil)
+if err != nil {
+	log.Fatal(err)
+}
+```
+
+The `CallAction` method returns the data in the `core/types/client.Records`
+type. See the [godocs](https://pkg.go.dev/github.com/kwilteam/kwil-db/core/types/client)
+for this type to see the methods available for accessing the records.
+
```
KwilLuke, line 343: Do we need an example for authenticating with KGW? Can I just pass a signer to `cl.CallAction`?

brennanjl: Kwil has a gateway client that can do that. It is an entirely separate package, which is not what this is documenting/

core/client/README.md
```diff
@@ -0,0 +1,364 @@
...
+functions on the command line. Go applications may use the package directly.
+
+## Get the `core` Go Module
+
+The `client` package is part of the `core` Go sub-module of the `kwil-db` repository. To use the package in your Go application, add it as a `require` in your project's `go.mod`:
+
+```sh
+$ go get github.com/kwilteam/kwil-db/core
+go: downloading github.com/kwilteam/kwil-db/core v0.1.2
+go: downloading github.com/kwilteam/kwil-db v0.7.2
```
Yaiba, line 26: how come main module is downloaded? maybe it's only for download?

jchappelow: I wondered the same, concluded the same -- probably due to version control constraints it must download the main module when resolving submodules.  Maybe?

Yaiba: yeah I guess.

core/client/README.md
jchappelow: Good question.  The signer itself conveys this.  When `core/transactions.(*Transaction).Sign` call's the signer's `Sign` method, It creates a `Signature` with `Type` set correctly so that it may be verified with the corresponding authenticator.

So when we use `makeSigner`, it creates a `*auth.EthPersonalSigner` that will set it's type as above.  To use ed25519, we'd use this instead to make the `Signer`:

```go
func makeEdSigner(keyHex string) auth.Signer {
	key, err := crypto.Ed25519PrivateKeyFromHex(keyHex)
	if err != nil {
		panic(fmt.Sprintf("bad private key: %v", err))
	}
	return &auth.Ed25519Signer{*key}
}
```

jchappelow:
Mmm, I think I need to quick write some godocs for the `core/types/client.Records` type and methods.

Before merging, I'll squash related commits, but keep the txapp/abci commit separate.

core/client/README.md
jchappelow: Oh, yes.  I had read the example incorrectly.  This is the revision to clarify: https://github.com/kwilteam/kwil-db/pull/662/commits/02577bde53868d8dd9293b083eb5b9ab71cabbcd#diff-037cd674919f54a98f50e6ee053384586c751a823ab95cb38e7cf948b125301cR326

core/client/example/go.mod
```diff
@@ -0,0 +1,50 @@
+module mykwilapp
+
+go 1.22.1
+
+require (
+	// this kuneiform rev uses the update sqlparser on kwil-db's main:
+	github.com/kwilteam/kuneiform v0.6.1-0.20240418203939-18f750d293bf
```
jchappelow, line 7: When this commit goes on a release branch, this will change to the tagged version.  I had to use this unreleased revision of kuneiform for the `main` PR since there was an incompatible sqlparser change.

jchappelow:
I believe all comments are now addressed.

Yaiba (approved):
LGTM

core/client/README.md
KwilLuke: Ah perfect - yeah I don't think we need to add that to the docs for now.

KwilLuke (approved):

core/client/README.md
MicBun: thanks!

brennanjl (approved):

PR #663, opened 2024-04-19
[v0.7.3] core readme and example

jchappelow:
This brings the core/client readme and example app back to `release-v0.7` from https://github.com/kwilteam/kwil-db/pull/662

Also:
- add Record godocs
- add new error codes to core/types/transactions

This PR includes **only** the changes to the `core` module.  After merging, we would need to tag `core/v0.1.3` and update the main module's `core` `require` (no replaces to help on release branches).  Then the fixed codes can be used by txapp etc. as in https://github.com/kwilteam/kwil-db/commit/13132e4470b241e3799e47597d307abb02b8322a


brennanjl (approved):

KwilLuke:
Does this need to go to main too? @brennanjl @jchappelow

PR #664, opened 2024-04-19
re-introduced parse module

brennanjl:
As discussed this morning, this PR re-introduces the parse module. It actually moves quite a bit of the engine to the parse module (such as most of the procedural code and sqlanalyzer). I've made it a separate PR to procedures so that it is easy to tell what is new vs https://github.com/kwilteam/kwil-db/pull/608, since #608 has already been reviewed and for the most part approved.

Keep in mind that there are also some half-baked additions for foreign calls that I included here from a separate branch. Making this change would've totally screwed the foreign call branch, so I decided to include it here.

`parse` now features a top-level `ParseKuneiform` function, which will parse a kuneiform string, as well as all actions and procedures. It also performs cartesian join checks, type checking, etc..

Ideally, we would only let consumers import the top-level package in `parse`, however I am not sure if this is possible. I tried adding an `internal` package, but the main Kwil module was unable to import from there, so I had to leave it as-is.


KwilLuke:
FYI @brennanjl, I think we need to add `foreign_calls` to proto before I can test this with kwil-js. I think it makes sense for me to wait for this and #669 to go to main before implementing the changes needed to support foreign_calls in kwil-js (assuming this will get merged at least a week before release).

internal/engine/ddl/foreign_procedure.go
```diff
@@ -0,0 +1,340 @@
...
+			str.WriteString(" ")
+			str.WriteString(pgStr)
+			outList = append(outList, name)
+		}
+	}
+
+	str.WriteString(`)`)
+
+	// If the return type is a table, we need to format the returns as a table.
+	if proc.Returns != nil && proc.Returns.IsTable {
```
Yaiba, line 144: I think this logic could use a bit refactor? Not sure if I get this logic right

```go
// if proc.Returns == nil
// ...
// logic for !Returns.IsTtable
// ...
// if proc.Returns.IsTable {
//   if len(proc.Returns.Fields) == 0{
//      str.WriteString(" RETURNS VOID")
//   }
//   str.WriteString(" RETURNS TABLE(")
//   ...
// }


```

Yaiba (changes requested):
In general looks to me, there are some new json tags which are not using snake case and we shoud enforce it

core/types/schema.go
```diff
@@ -818,8 +842,8 @@ func (p *Procedure) IsOwner() bool {
 // ProcedureReturn holds the return type of a procedure.
 // EITHER the Type field is set, OR the Table field is set.
 type ProcedureReturn struct {
-	IsTable bool         `json:"isTable,omitempty"`
-	Fields  []*NamedType `json:"returnTypes,omitempty"`
+	IsTable bool         `json:"isTable"`
```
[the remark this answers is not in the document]
brennanjl: Good point. I will make that change. @KwilLuke this will be a breaking change for you

core/types/schema.go
```diff
@@ -761,7 +785,7 @@ type Procedure struct {
 	Body string `json:"body"`
 
 	// Returns is the return type of the procedure.
-	Returns *ProcedureReturn `json:"returnTypes,omitempty"`
+	Returns *ProcedureReturn `json:"returnTypes"`
```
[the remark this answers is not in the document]
brennanjl: @KwilLuke another breaking change for you

parse/go.mod
```diff
@@ -0,0 +1,23 @@
+module github.com/kwilteam/kwil-db/parse
+
+go 1.21.0
+
+replace github.com/kwilteam/kwil-db/core => ../core
+
+require (
+	github.com/antlr4-go/antlr/v4 v4.13.0
+	github.com/kwilteam/action-grammar-go v0.1.1
```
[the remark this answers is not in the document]
brennanjl: ! good catch

parse/parse.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/kwilteam/kwil-db/parse/kuneiform"
+	procedures "github.com/kwilteam/kwil-db/parse/procedures"
+)
+
+// ValidateSchema validates the syntax of the action and procedure bodies
+// in a schema.
+func ValidateSchema(schema *types.Schema) error {
+	// to validate, we will simply call the analyzers and discard
+	// the results. If there is an error, we will return it.
+	_, err := actions.AnalyzeActions(schema, schema.DBID())
```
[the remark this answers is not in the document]
brennanjl: We actually do test this for procedures (we can't for actions).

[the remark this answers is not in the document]
brennanjl: See `parse/metadata/functions.go`. Function definitions have a `ValidateArgs` func.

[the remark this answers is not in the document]
Yaiba: Those are for functions right?
I mean to validate table column attributes, before it's send to server.  The example I mentioned is for something like 

```
database testdb;
table user {
 id int primary,
 name text min(3);
}
```

The table def is wrong

scratch
```diff
@@ -1,9 +0,0 @@
-CREATE OR REPLACE FUNCTION ds_x975a5d686cb685d60bed9d99a0d3cb80e4f712dbf8d2700297085752.get_owned_users(_param_address TEXT)
```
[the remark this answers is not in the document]
brennanjl: Oddly, I do not have this file locally

parse/parse.go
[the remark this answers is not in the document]
brennanjl: Oh oh oh I see.

Will add that, good catch

core/types/schema.go
```diff
@@ -785,7 +797,7 @@ type Procedure struct {
 	Body string `json:"body"`
 
 	// Returns is the return type of the procedure.
-	Returns *ProcedureReturn `json:"returnTypes"`
+	Returns *ProcedureReturn `json:"return_types"`
```
Yaiba, line 800: There are some tags, for example `isArrary`, should change those as well

core/types/schema.go
```diff
@@ -278,7 +278,19 @@ type Attribute struct {
...
+	case MIN, MAX:
+		if !col.Type.Equals(IntType) {
+			return fmt.Errorf("attribute %s is only valid for int columns", a.Type)
+		}
+	case MIN_LENGTH, MAX_LENGTH:
+		if !col.Type.Equals(TextType) {
+			return fmt.Errorf("attribute %s is only valid for text columns", a.Type)
+		}
+	}
+
```
Yaiba, line 293: This is good. 
Maybe also add index and foreign_key validation? i.e. make sure referred column exists. 
That's so far the validations I can think of that we can do for table definition.

Yaiba:
This pr looks good to me now.  There is one last thing,  we might want to get rid of the `replace` in go mod before we merge?

@jchappelow are you reviewing this?

Yaiba (approved):

jchappelow:
> This pr looks good to me now. There is one last thing, we might want to get rid of the `replace` in go mod before we merge?
> 
> @jchappelow are you reviewing this?

Yeah I'm switching to it now. Will just give it an hour.  O gather this is holding up other things

jchappelow:
> This pr looks good to me now. There is one last thing, we might want to get rid of the `replace` in go mod before we merge?

I don't see how we can get rid of `replace`s now. require on `github.com/kwilteam/kwil-db/core v0.1.2` from parse.go isn't adequate.  eventually we'll switch that to require `core/v0.2.0` but not now

parse/go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/kwilteam/kwil-db/parse
+
+go 1.21.0
+
+replace github.com/kwilteam/kwil-db/core => ../core
+
+require (
+	github.com/antlr4-go/antlr/v4 v4.13.0
+	github.com/kwilteam/kwil-db v0.7.2
```
jchappelow, line 9: This almost defeats the whole purpose of making a parse module.  `parse` consumers don't want to require the main module, and we don't want a circular `parse` <-> main module require.

jchappelow: I see it used in parse/sql/sqlanalyzer/clean/clean.go for `validation.IsKeyword`.

I guess you wanted to use `core/types/validation.IsKeyword` instead?

I also see the only other users of `validation.IsKeyword` are `validation.ValidateIdentifier` and `validation.ValidateDBID`, and both of those seem to be unused?  (there's a also now a `core/types/validation.ValidateIdentifier` that seems to provide this now, so I bet the common/validation is now obsolete?

jchappelow: Ah, yeah, that's all it was.

`import "github.com/kwilteam/kwil-db/core/types/validation"

in parse/sql/sqlanalyzer/clean/clean.go

seems to remove this require.

parse/kuneiform/gen/kuneiform_lexer.go
```diff
@@ -0,0 +1,483 @@
+// Code generated from KuneiformLexer.g4 by ANTLR 4.13.1. DO NOT EDIT.
```
jchappelow, line 1: Was a git rm from kuneiform/gen/kuneiform_lexer.go?

jchappelow: Oh, the diff must have exceeded it's threshold for matching.  All good.

parse/go.mod
brennanjl: Oh wow total mistake on my end. That old validation package shouldn't be there, definitely meant to use core

Yaiba:
> > This pr looks good to me now. There is one last thing, we might want to get rid of the `replace` in go mod before we merge?
> 
> I don't see how we can get rid of `replace`s now. require on `github.com/kwilteam/kwil-db/core v0.1.2` from parse.go isn't adequate. eventually we'll switch that to require `core/v0.2.0` but not now

right, forgot we dont' have a tag on main works

parse/go.mod
brennanjl: TIL you can have circular dependencies between submodules

parse/kuneiform/gen/kuneiform_lexer.go
brennanjl: Ah I was not actually sure what was going on there

internal/engine/ddl/foreign_procedure.go
```diff
@@ -0,0 +1,340 @@
...
+// of the new parts of the engine are still in flux.
+
+// GenerateForeignProcedure generates a plpgsql function that allows the schema to dynamically
+// call procedures in other schemas, expecting certain inputs and return values. It will prefix
+// the generated function with _fp_ (for "foreign procedure").
+func GenerateForeignProcedure(proc *types.ForeignProcedure) (string, error) {
+	str := strings.Builder{}
+
+	// first write the header
+	str.WriteString(fmt.Sprintf(`CREATE OR REPLACE FUNCTION _fp_%s(_dbid TEXT, _procedure TEXT`, proc.Name))
```
jchappelow, line 106: I think this function is not a hot path, called only on deploy, but for simple string concatenation it's often a good idea to go with `+`.  The `fmt` printers are relatively heavy duty, using lots of reflection and generally useless operations for simple tasks like this.

```go
str.WriteString(`CREATE OR REPLACE FUNCTION _fp_` + proc.Name + `(_dbid TEXT, _procedure TEXT`)
```

https://dev.to/pmalhaire/concatenate-strings-in-golang-a-quick-benchmark-4ahh

Again, probably not a worry here, but something to be aware of in code that is called a lot.

internal/engine/ddl/foreign_procedure.go
```diff
@@ -0,0 +1,340 @@
...
+		for i, out := range proc.Returns.Fields {
+			str.WriteString(", OUT ")
+			pgStr, err := out.Type.PGString()
+			if err != nil {
+				return "", err
+			}
+
+			str.WriteString("_out_")
+			name := fmt.Sprintf("_out_%d", i+1)
+			str.WriteString(name)
```
jchappelow, line 134: Maybe I'm reading this wrong because it's Friday afternoon, but this looks like it's going to write `_out__out_X`.

internal/engine/ddl/foreign_procedure.go
brennanjl: Ah thats good to know. None of the string formatting are hot paths as you mentioned, but I will note that for the future.

I probably won't bother to change this right now, since I actually think we will want to re-evaluate how procedures work under the hood after this release. translating to plpgsql is very hard, and extremely error prone. It will be near impossible to add gas to; I think it will require making it a postgres extension after this initial version.

internal/engine/ddl/foreign_procedure.go
brennanjl: Lol yes this is a bug

parse/sql/sqlanalyzer/typing/visitor.go
```diff
@@ -73,21 +74,23 @@ func (t *typeVisitor) VisitRelationSubquery(p0 *tree.RelationSubquery) any {
...
 			Name:     p0.Alias, // this can be ""
 			Relation: r,
 		})
 	})
 }
 
-func findProcedure(name string, procedures []*types.Procedure) (*types.Procedure, error) {
+// findProcedure finds a procedure by name in a list of procedures.
+// It returns false if the procedure is not found.
+func findProcedure(name string, procedures []*types.Procedure) (proc *types.Procedure, found bool) {
```
jchappelow, line 86: This unexported function appears to be unused.

parse/wasm/wasm.go
```diff
@@ -8,11 +8,11 @@ import (
 	"fmt"
 	"syscall/js"
 
-	"github.com/kwilteam/kwil-db/kuneiform"
+	"github.com/kwilteam/kwil-db/parse"
```
jchappelow, line 11: Up a few lines, on the `//go:generate` line, there appears to be a typo, lacking a space between `netgo` and `wasm.go` (it's passing the tag `netgowasm.go` and the source is implied).

```diff
diff --git a/parse/wasm/wasm.go b/parse/wasm/wasm.go
index d3ba98ce..19413e1e 100644
--- a/parse/wasm/wasm.go
+++ b/parse/wasm/wasm.go
@@ -1,6 +1,6 @@
 //go:build js && wasm
 
-//go:generate env GOOS=js GOARCH=wasm go build -o kuneiform.wasm -ldflags "-s -w" -trimpath -tags netgowasm.go
+//go:generate env GOOS=js GOARCH=wasm go build -o kuneiform.wasm -ldflags "-s -w" -trimpath -tags netgo wasm.go
 package main
```

Probably good to add ` CGO_ENABLED=0` too.

internal/engine/ddl/foreign_procedure_test.go
```diff
@@ -0,0 +1,209 @@
...
+			// fmt.Printf("\n\n\nWANT:")
+			// fmt.Println(test.want)
+			// panic("")
+
+			require.Equal(t, removeWhitespace(test.want), removeWhitespace(got))
+		})
+	}
+}
+
+// func Test_ForeignProcedureReturnsTable(t *testing.T) {
```
jchappelow, line 95: Need to keep this one?

internal/engine/execution/procedure.go
```diff
@@ -71,173 +67,128 @@ type preparedAction struct {
...
+						calledAction = p
 						break
 					}
 				}
-				if action == nil {
-					return nil, fmt.Errorf(`procedure "%s" not found`, stmt.Method)
-				}
 
-				for _, mod := range action.Modifiers {
+				for _, mod := range calledAction.Modifiers {
```
jchappelow, line 161: Any way this can fail to fall the `calledAction`, leaving it nil?

internal/engine/execution/procedure.go
```diff
@@ -445,73 +396,24 @@ type evaluatable func(ctx context.Context, exec dbQueryFn, values map[string]any
...
-						},
-					},
-				},
-			},
-		}
-
-		stmt, err := tree.SafeToSQL(selectTree)
-		if err != nil {
-			return nil, err
+func makeExecutables(params []*actions.InlineExpression) ([]evaluatable, error) {
```
jchappelow, line 399: `makeExecutables` doesn't have a possible error return, although the `evaluatables` do

internal/engine/execution/procedure.go
```diff
@@ -445,73 +396,24 @@ type evaluatable func(ctx context.Context, exec dbQueryFn, values map[string]any
...
+	var evaluatables []evaluatable
+
+	for _, param := range params {
+		// copy the param to avoid loop variable capture
+		param2 := &actions.InlineExpression{
+			Statement:     param.Statement,
+			OrderedParams: append([]string{}, param.OrderedParams...),
```
jchappelow, line 406: Is this append for a one line slice copy?  Maybe `slices.Clone` would be more straightforwared, but I think a deep copy of the fields is more than what's needed to avoid the loop variable capture.

internal/engine/execution/queries.go
```diff
@@ -96,7 +98,7 @@ var (
 	// sqlBackfillSchemaTableV1 adds the owner and name to all existing schemas,
 	// and updates the version to 1.
 	sqlBackfillSchemaTableV1 = fmt.Sprintf(`
-	UPDATE %s.kwil_schemas SET owner = $1, name = $2, version = 1;
+	UPDATE %s.kwil_schemas SET owner = $1, name = $2, version = 1 WHERE dbid = $3;
```
jchappelow, line 101: Oh yeah, I thought this was in the last PR.

internal/engine/integration/schema_test.go
```diff
@@ -27,23 +28,171 @@ func Test_Schemas(t *testing.T) {
...
 	}
 
+	// the tests rely on three schemas:
+	// users: a table of users, which maps a wallet address to a human readable name
+	// social_media: a table of posts and post_counts. posts contains posts, and post_counts contains the number of posts a user has made.
+	// video_game: a table of scores tracks users high scores in a video game.
+	// posts and video+game also have admin commands for setting the dbid and procedure names.
 	testCases := []testCase{
+		// {
+		// 	name: "simple schema",
```
jchappelow, line 41: uncomment or delete?

parse/kuneiform/grammar/KuneiformLexer.g4
```diff
@@ -100,19 +102,26 @@ fragment DIGIT: [0-9];
 // STMT_MODE captures actions and procedures.
 // it is used to allow us to correctly parse action / procedure bodies
 mode STMT_MODE;
+    COMMENT:   ( '//' ~[\r\n]* '\r'? '\n'
+        | '/*' .*? '*/'
```
jchappelow, line 106: In `.*?`, what is the `?` for?  At least in regexp, `*` means zero or more.

jchappelow (changes requested):
Looks pretty good.  Need a few updates to CI and tooling though:

In .golangci.yml:


```diff
diff --git a/.golangci.yml b/.golangci.yml
index 42d61f4b..4c2ddafd 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,6 +3,10 @@ run:
   skip-dirs:
     - core/rpc/protobuf
     - core/rpc/http
+    - parse/sql/gen
+    - parse/procedures/gen
+    - parse/kuneiform/gen
+    - parse/actions/gen
```

Taskfile.yml lint and fmt updates:


```diff
diff --git a/Taskfile.yml b/Taskfile.yml
index 480b4f71..b38ed18f 100644
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -47,8 +47,12 @@ tasks:
       # so we have to list all the folders that are adjacent to protobuf.
       - |
         goimports -format-only -w ./cmd ./internal ./test ./core/client \
-          ./core/crypto ./core/log ./core/rpc/client ./core/rpc/conversion \
-          ./core/rpc/transport ./core/types ./core/utils
+          ./core/crypto ./core/log ./core/rpc/client \
+          ./core/rpc/transport ./core/types ./core/utils \
+          ./parse/parse.go ./parse/metadata ./parse/kuneiform/*.go \
+          ./parse/procedures/*.go ./parse/procedures/parser ./parse/procedures/visitors \
+          ./parse/sql/*.go ./parse/sql/postgres ./parse/sql/sqlanalyzer ./parse/sql/tree \
+          ./parse/actions/*.go ./parse/actions/parser \
 
   tidy:
     desc: go mod tidy each module
@@ -64,7 +68,7 @@ tasks:
     desc: Lint with golangci-lint
     cmds:
       # skip-dirs only works in .yaml
-      - golangci-lint run ./... ./core/... ./test/... -c .golangci.yml
+      - golangci-lint run ./... ./core/... ./test/... ./parse/... -c .golangci.yml
```

Also, `git rm internal/parse/sql/tree/functions.md`?

internal/engine/execution/procedure.go
brennanjl: I don't think so, but I will add a check just in case

internal/engine/execution/procedure.go
brennanjl: I don't think so, but I will add a check just in case

internal/engine/execution/procedure.go
brennanjl: Did not know about `slices.Clone`.

However you are right, I do not think we actually have to copy the `OrderedParams` slice.

internal/engine/execution/queries.go
brennanjl: Could've sworn it was. A lot of this has sort've been continual work, as the first few days of foreign calls were being worked on while procedures was in review.

Probably just got my wires crossed here

parse/kuneiform/grammar/KuneiformLexer.g4
brennanjl: This seems to be the way of doing block comments in antlr. I noticed Gavin had used it before, chatgpt used it, others use it: https://stackoverflow.com/questions/12898052/antlr-how-to-skip-multiline-comments

parse/wasm/wasm.go
```diff
@@ -1,18 +1,18 @@
 //go:build js && wasm
 
-//go:generate env GOOS=js GOARCH=wasm go build -o kuneiform.wasm -ldflags "-s -w" -trimpath -tags netgowasm.go
+//go:generate env GOOS=js GOARCH=wasm go build -o kuneiform.wasm -ldflags "-s -w" -trimpath -tags netgo wasm.go CGO_ENABLED=0
```
jchappelow, line 3: `CGO_ENABLED=0` would go with the other `env` settings like `GOOS` rather.

brennanjl: whoops thanks. Sorry, been a long day

jchappelow (approved):

PR #667, opened 2024-04-22
Kwil-admin setup testnet command fix

charithabandi:


cmd/kwil-admin/nodecfg/toml_test.go
```diff
@@ -74,34 +75,62 @@ func Test_IncrementingPorts(t *testing.T) {
...
 		want   string
+		fail   bool
 	}
 
 	testcases := []testcase{
 		{
 			input:  "localhost:26656",
 			amount: 1,
-			want:   "localhost:26657",
+			want:   "tcp://localhost:26657",
```
jchappelow, line 85: Hmmm this seems undesirable for our service settings, right?  I'm surprised it's works for the http setting for example. 

"tcp://" is not a real scheme, just a strange cometbft thing. Seems unfortunate to adopt that for our http rpc service. 

Seem weird or ok?

charithabandi: Yeah, agreed. Let me fix that

charithabandi: Yeah, agreed. Let me fix that

cmd/kwil-admin/nodecfg/generate.go
```diff
@@ -546,7 +556,7 @@ func incrementPort(incoming string, amt int) (string, error) {
 	// Reconstruct and return the new URL
 	newUrl := net.JoinHostPort(host, strconv.Itoa(port))
 
-	if res.Scheme != "localhost" {
+	if schemaExists {
```
jchappelow, line 559: Nice.  I like this solution way better.  The scope of the changes logically should be limited to nodecfg (other than exporting `HasScheme`).
Will test in the a.m., but LGTM.

jchappelow (approved):

PR #671, opened 2024-04-23
docker: use 16.2 in compose

jchappelow:
With the `kwildb/postgres` docker image updated in https://github.com/kwilteam/kwil-db/pull/605, this updates the CI to use that version explicitly.

Also use `postgres:16.2` in the compose service files, and update the command flags to match `kwildb/postgres:16.2-1`.


Yaiba (approved):
LGTM

PR #672, opened 2024-04-23
VoteBodyTx fixes

charithabandi:
There is an issue with `FilterExistsNoBody` which is supposed to take list of event IDs and filter out the events/resolutions to which we already received a voteBody. But the API filters out only the ids for which the resolution and body exists. It doesn't return the event ids to which the resolutions are not created yet. This becomes more pronounced in a single-node deployments as we only do voteBody Transactions, so resolutions without body are never created, Essentially the extensions will not work in a single-node deployments.

### DIFF
--- a/internal/txapp/txapp.go
+++ b/internal/txapp/txapp.go
@@ -654,11 +654,13 @@ func (r *TxApp) ProposerTxs(ctx context.Context, txNonce uint64, maxTxsSize int6
 		ids = append(ids, event.ID())
 	}
 
+	// Filter out events which doesn't have resolutions or resolution body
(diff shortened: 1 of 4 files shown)


brennanjl:
Do you think you could add a specification to the single-node acceptance tests that tests this? I think you could probably re-use a lot of what already exists in the integration tests to make it quick.

charithabandi:
> Do you think you could add a specification to the single-node acceptance tests that tests this? I think you could probably re-use a lot of what already exists in the integration tests to make it quick.

Yeah, will do.

brennanjl (approved):

PR #679, opened 2024-04-25
json-rpc 2.0 server

jchappelow:
This gives us an JSON-RPC 2.0 server on port 8484, endpoint path `/rpc`.  I've chosen a non root path to keep it well separated from future paths on this port that we'd want for REST.

"Ready for review" in the sense that we can consider this as a proposal.  Needs a bit of work to actually merge, although it's fully functional and integration tests pass.

## JSON-RPC 2.0

Spec: https://www.jsonrpc.org/specification

> JSON-RPC is a lightweight remote procedure call protocol. It's designed to be simple!
>
> The general mechanism consists of two peers establishing a data connection. During the lifetime of a connection, peers may invoke methods provided by the other peer. To invoke a remote method, a request is sent. Unless the request is a notification it must be replied to with a response.

- HTTP POST requests on one endpoint
- types of JSON-RPC 2.0 objects
    * request with "jsonrpc", "method", "params", "id"
    * response with "result", "error", "id"
    * error that goes in the response.error with "code", "message", "data"
- structured objects for the "params" field i.e. the args for a given "method"
- conventional and custom error "code"
- arbitrary structured error "data" types

## packages

 - `core/rpc/json` contains `package jsonrpc`, containing the type definitions used by both client and server.  The handful of types defined by JSON-RPC 2.0 are defined in jsonrpc.go, while the kwil requests and responses are in commands.go and responses.go.  Error codes and structured error types are in errors.go.  The known method names are defined in methods.go
 - `internal/services/jsonrpc` contains the http `Server` and the "tx service" used by it.
 - `core/rpc/client/user/jsonrpc` contains the `Client`, which satisfies the `user.TxSvcClient` interface required by `core/client.Client`

## changes to defaults for testing

For sake of testing, I've quickly rigged everything to use the JSON-RPC service's endpoint (port 8484 and path "/rpc").  Acceptance and integration tests both pass.

`core/client.Client`'s default `user.TxService` implementation is now the JSON-RPC one from `core/rpc/client/user/jsonrpc` instead of `core/rpc/client/user/http`.

`kwil-cli`'s default provider is `"http://localhost:8484/rpc"`, also for quick testing.  

`kwild` listens on `0.0.0.0:8484`, hard coded for now.

## other changes

A couple types in `core/types/transactions` have json tags now since they are used in `jsonrpc` request/response types directly. Potentially breaking for the http gateway on 8080, may revert, or clone these types in `jsonrpc` so they are more idiomatic (not capitalized `CamelCase`).

A handful of other json tags are renamed, such as in `core/crypto/auth/signer.Signature`.  Potentially breaking for the http gateway on 8080, may revert.

## Next

- http at first, ws later for notifications.
- Just "tx" service now.  Add services for "admin" and "functions" so we can deprecate those gRPC services now too.
- Currently supports only structured/named arguments (object), not positional (array).  Named looks like `"params: {"arg1": "xyz", "arg3": 123}`, while positional looks like `"params": ["xyz", null, 123]`.  Positional is JSON-RPC 1.0, while named is 2.0.  Named is much simpler and more flexible, especially with optional parameters.  Technically 1.0 still includes positional support, but it's not important for us right now.
- Emulate the openAPI endpoints so we can remove all the generated code while keeping the legacy port 8080 gateway functional.
- Add proper json-rpc configs instead of hijacking the http gateway settings all over.


jchappelow:
If we define an ["OpenRPC" spec](https://spec.open-rpc.org/) for our service, then the public playground can can provide both docs and an a way to interact with it.  For example: https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/eth1.0-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false

jchappelow:
To demonstrate how JSON-RPC generally works:

```
$ curl -H "content-type: application/json" -d '{"jsonrpc":"2.0","method":"chain_info","params":null,"id":1}' 127.0.0.1:8484/rpc
{"jsonrpc":"2.0","id":1,"result":{"chain_id":"kwil-chain-zk8NDyl2","block_height":8,"block_hash":"d5f281c5bf3302564a70ac2b4684d4879daddd3653196df60d547610d3f99580"}}
```

Basically set `"method"` and `"params"`.  The method is a string and the params are object like `{"account": "asvw==", "status": 1}`

This is the standard in just about all nodes.  Ethereum, Bitcoin Core, CometBFT, etc.

cmd/common/display/message_test.go
```diff
@@ -143,20 +143,20 @@ func Example_respTxQuery_json() {
 	//     "hash": "31303234",
 	//     "height": 10,
 	//     "tx": {
-	//       "Signature": {
```
Yaiba, line 146: Didn't realize it has not json tag.. 🤦‍

brennanjl:
Taking another scan and it looks pretty good to me. The only feedback I really have is regarding the type duplicates that include json tags. I think I remember you mentioning this, but is there a reason why you didn't just tag the original structs?

AFAIK, the only json tags we currently rely on are defined on `Schema` and its children (`Table`, `Action`, etc.). The rest of them I am pretty confident we do not use.

core/rpc/json/errors.go
```diff
@@ -0,0 +1,23 @@
...
+
+// General error codes
+const (
+	ErrorUnspecified = iota // don't use
+	ErrorInternal
+	ErrorNotFound
+	ErrorUnknownMethod
+	ErrorRequestEncoding
+	ErrorBadArgument
+	ErrorInvalidSignature // maybe
```
Yaiba, line 11: maybe for `function` service

core/rpc/json/errors.go
```diff
@@ -0,0 +1,23 @@
...
+	ErrorUnauthorized
```
Yaiba, line 12: Not sure about this. Seems kwildb won't return this error?

core/rpc/json/errors.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// More detailed errors use a structured error type in the "data" field of the
+// responses "error" object. These may include a code field for domain-specific
+// codes.
+
+type BroadcastError struct {
```
Yaiba, line 19: We probably will need more of this for other Methods?

core/rpc/json/errors.go
```diff
@@ -0,0 +1,23 @@
+package jsonrpc
+
+// General error codes
+const (
+	ErrorUnspecified = iota // don't use
+	ErrorInternal
```
Yaiba, line 6: Should we comply to SPEC since those are reserved error code?

internal/services/jsonrpc/service.go
```diff
@@ -0,0 +1,364 @@
...
+	if err != nil {
+		logger.Error("failed to broadcast tx", log.Error(err))
+		return nil, jsonrpc.NewError(jsonrpc.ErrorInternal, "failed to broadcast transaction", nil)
+	}
+
+	code, txHash := res.Code, res.Hash.Bytes()
+
+	if txCode := transactions.TxCode(code); txCode != transactions.CodeOk {
+		errData := &jsonrpc.BroadcastError{
+			Code:    code, // e.g. invalid nonce, wrong chain, etc.
```
Yaiba, line 111: I think we probably should have our own error code associate to know errors, like for example we can have some engine error code pop up here, so we know there is some deeper issue when engine execute. not sure if it's better than error message like "e1: e2: e3: e4: ...".

Yaiba (commented):
Over all looks good to me, this is solid proposal.  json-rpc is really simple.

To merge, I guess you have to change default endpoint back to grpc-gateway.

core/rpc/json/errors.go
jchappelow: Ah, so for example, use -32601 for `ErrorUnknownMethod`, yeah, should do that.

image

That what you had in mind? Seems like a good idea.

internal/services/jsonrpc/service.go
jchappelow: I think that's what this does?

We have defined:

```go
type TxCode uint32

const (
	CodeOk                  TxCode = 0
	CodeEncodingError       TxCode = 1
	CodeInvalidTxType       TxCode = 2
	CodeInvalidSignature    TxCode = 3
	CodeInvalidNonce        TxCode = 4
	CodeWrongChain          TxCode = 5
	CodeInsufficientBalance TxCode = 6
	CodeInsufficientFee     TxCode = 7
	CodeInvalidAmount       TxCode = 8
	CodeInvalidSender       TxCode = 9

	// engine-related error code
	CodeInvalidSchema  TxCode = 100
	CodeDatasetMissing TxCode = 110
	CodeDatasetExists  TxCode = 120

	CodeUnknownError TxCode = math.MaxUint32
)
```

So because the client decodes a `BroadcastError`, they know the context in which the code should be interpreted.

This is facilitated by the recent change to txApp that added this: https://github.com/kwilteam/kwil-db/blob/98daaf16af14e5c498eb466c9136a149f44b2e3a/internal/txapp/routes.go#L94-L109

So we have codes that correspond to well defined engine errors.

core/rpc/json/errors.go
Yaiba: Yes you're right

internal/services/jsonrpc/service.go
jchappelow: Really good point.  Moving that way now in https://github.com/kwilteam/kwil-db/pull/679/commits/81d10de31e808446e0be6c8953dfe974ca49ad25

Yaiba: cool, we'll see how this approach pans out

jchappelow:
For a separate PR, I've already started refactoring to separate server from service in the way that gRPC servers worked.  This is a step for admin service whether we use the same server (port and/or endpoint) or not because the server code is reusable with different services.
https://github.com/jchappelow/kwil-db/commit/3f6ade38ec07f18341c9619301293959e5bfd7e5

Anyway, I think this PR is good to go.  Type duplication minimized, and RPC error handling and codes are in OK shape.  It took a while to deal with the details of config, integ test updates, tagging types, etc

I'll make the issues for the next steps soon.

jchappelow:
kwil-db issues for next steps: https://github.com/kwilteam/kwil-db/labels/json-rpc
kgw issue: https://github.com/kwilteam/kgw/issues/60

test/integration/kwild_test.go
```diff
@@ -18,7 +18,7 @@ var dev = flag.Bool("dev", false, "run for development purpose (no tests)")
 
 var spamTest = flag.Bool("spam", false, "run the spam test that requires a special docker image to be built")
 
-var drivers = flag.String("drivers", "http,cli", "comma separated list of drivers to run")
+var drivers = flag.String("drivers", "jsonrpc,http,cli", "comma separated list of drivers to run")
```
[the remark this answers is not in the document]
jchappelow: Yeah.  :(

cmd/kwild/config/flags.go
```diff
@@ -14,6 +14,7 @@ func AddConfigFlags(flagSet *pflag.FlagSet, cfg *KwildConfig) {
 
 	// General APP flags:
 	flagSet.StringVar(&cfg.AppCfg.PrivateKeyPath, "app.private-key-path", cfg.AppCfg.PrivateKeyPath, "Path to the node private key file")
+	flagSet.StringVar(&cfg.AppCfg.JSONRPCListenAddress, "app.jsonrpc-listen-addr", cfg.AppCfg.JSONRPCListenAddress, "kwild JSON-RPC listen address")
 	flagSet.StringVar(&cfg.AppCfg.GrpcListenAddress, "app.grpc-listen-addr", cfg.AppCfg.GrpcListenAddress, "kwild gRPC listen address")
```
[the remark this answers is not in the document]
jchappelow: I'm down to remove the grpc settings.  Will update.

internal/services/jsonrpc/server.go
```diff
@@ -0,0 +1,218 @@
...
+// var routes = map[string]methodHandler{
+// 	jsonrpc.MethodAccount:   handleAccount,
+// 	jsonrpc.MethodChainInfo: nil,
+// }
+
+// func registerMethod[T any](method string, req *T, handler func(context.Context, *T) (any, *jsonrpc.Error)) {}
+
+// handleRequest sends the request to the correct handler function if able.
+func (s *Server) handleRequest(ctx context.Context, req *jsonrpc.Request) *jsonrpc.Response {
+	if req.JSONRPC != "2.0" || req.ID == 0 {
```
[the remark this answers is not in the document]
jchappelow: Ah, good catch. This was not smart.  It's only true if the any is an `int(0)` but not an `int64(0)`.  Basically and unset or zero numeric for `ID` is discouraged in requests since in responses a null/unset ID means it's a notification, so the server should reject that malformed request.  Will fix this.

core/rpc/json/responses.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/kwilteam/kwil-db/core/types"
+	"github.com/kwilteam/kwil-db/core/types/transactions"
+)
+
+// This file defines the response types. There is one for each request type.
+// As with the request types, many are based on types in the core module.
+
+// AccountResponse contains the response object for MethodAccount.
+type AccountResponse struct {
+	Identifier []byte `json:"identifier,omitempty"`
```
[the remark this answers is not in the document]
jchappelow: Potentially.  If we make it `HexBytes` in the requests, then we should make it here.  I'm leaning toward using hex for the account identifiers since in the majority of cases it's an ethereum address, which is hex.  Yeah, will make the change now.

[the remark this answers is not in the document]
Yaiba: cool, I was not sure about this, just think maybe we should.

cmd/kwild/config/flags.go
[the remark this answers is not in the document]
Yaiba: ah you're right... It would be problem for testing.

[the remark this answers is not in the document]
jchappelow: I think I have a solution to this.  Seems like I can have it listen on `127.0.0.1:0` and then get the actual port from the listener and give that to `buildGatewayServer`.  Testing that out now.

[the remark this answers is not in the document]
Yaiba: does using `0` as port will just pick up a random port?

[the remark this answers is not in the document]
jchappelow: Yep, it works like this: https://github.com/kwilteam/kwil-db/pull/679/commits/0226f5bd4187344a13a72e722685a16ef155982b#diff-d8f52f5a516a5d0335c26047aceb97265c4c7fdc6d5cc9ac7a4dbf42cd580af2L509
After you create the `net.Listener`, you use it's `Addr()` method to get the hostport that it ended up selecting.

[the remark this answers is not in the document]
Yaiba: never thought port `0` is reserved for this purpose 🤣 , TIL

jchappelow:
Took a while, again, but I think I hit everything there @Yaiba.  Cleaned out all gRPC config flags/settings, fixed the ID check, and converted owner/sender fields to HexBytes so that is marshals/unmarshals as a hex string and looks like an address.  I kept all the pubkey fields in the validator types as []byte though, for now.

build/package/docker/kwild.debug.dockerfile
```diff
@@ -30,5 +30,5 @@ COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=build /app/dist/kwild ./kwild
 COPY --from=build /app/dist/kwil-admin ./kwil-admin
 COPY --from=build /app/dist/kwil-cli ./kwil-cli
-EXPOSE 40000 50151 50051 8080 8484 26656 26657
+EXPOSE 40000 50151 8080 8484 26656 26657
```
Yaiba, line 33: 👍

Yaiba (approved):
LGTM

PR #681, opened 2024-04-25
[v0.6.8] Introduce broadcast tx commit timeout

charithabandi:


cmd/kwild/config/flags.go
```diff
@@ -44,6 +44,7 @@ func AddConfigFlags(flagSet *pflag.FlagSet, cfg *KwildConfig) {
 
 	// Chain RPC flags
 	flagSet.StringVar(&cfg.ChainCfg.RPC.ListenAddress, "chain.rpc.listen-addr", cfg.ChainCfg.RPC.ListenAddress, "Chain RPC listen address")
+	flagSet.Var(&cfg.ChainCfg.RPC.BroadcastTxTimeout, "chain.rpc.broadcast-tx-timeout", "Chain RPC broadcast transaction timeout")
```
jchappelow, line 47: We've been using `Var` for the other durations, so no objection, but I wonder why we don't use `DurationVar`.  Seems to exist.  Maybe just doesn't make help anything.

jchappelow:
Will you make this PR for `main` as well?

charithabandi:
> Will you make this PR for `main` as well?

Yup, done. https://github.com/kwilteam/kwil-db/pull/684

cmd/kwild/config/flags.go
charithabandi: not sure why we switched to Var, I am sure we used to use DurationVar before.

jchappelow (approved):

PR #682, opened 2024-04-25
tx broadcast debugging

jchappelow:
This is https://github.com/kwilteam/kwil-db/pull/680, but for `main` and also with warning logged in `prepareBlockTransactions` if a transaction is held back for exceeding the block size limit (which is only likely in the case of lots of validator inserted txns).

### DIFF
--- a/cmd/kwil-cli/cmds/account/balance.go
+++ b/cmd/kwil-cli/cmds/account/balance.go
@@ -15,11 +15,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
-	confirmed bool
-)
-
 func balanceCmd() *cobra.Command {
+	var confirmed bool
 	cmd := &cobra.Command{
 		Use:   "balance",
 		Short: "Gets an account's balance and nonce",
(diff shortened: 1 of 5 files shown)


cmd/kwil-cli/cmds/account/balance.go
```diff
@@ -15,6 +15,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	confirmed bool
```
brennanjl, line 19: Pretty minor nit, but there's a convention I have seen in a lot of cobra repos where flags get declared within the command. So here, it would be:

```go
func balanceCmd() *cobra.Command {
    var confirmed bool
	cmd := &cobra.Command{
		Use:   "balance",
```

brennanjl (approved):

cmd/kwil-cli/cmds/account/balance.go
jchappelow: Oh yeah that's a good point.  It doesn't need global scope.  Sec

brennanjl (approved):

PR #685, opened 2024-04-25
return schema from http server

brennanjl:
Fixes the bug mentioned here: https://github.com/truflation/tsn-data-provider/issues/58#issuecomment-2077994594

The Truflation team noticed that schema extensions were not properly being returned from the txsvc


jchappelow (approved):
lgtm

PR #688, opened 2024-04-30
change kwil-cli config and flags 

Yaiba:
BREAKING CHANGE:
- change flags `kwil-provider` to `provider`
- change config field `grpc_url` to `provider`
- change default config folder from `~/.kwil_cli` to `~/.kwil-cli`

### DIFF
--- a/cmd/kwil-cli/cmds/common/roundtripper.go
+++ b/cmd/kwil-cli/cmds/common/roundtripper.go
@@ -35,8 +35,8 @@ func DialClient(ctx context.Context, cmd *cobra.Command, flags uint8, fn RoundTr
 		return err
 	}
 
-	if conf.GrpcURL == "" {
-		return fmt.Errorf("kwil provider url is required")
+	if conf.Provider == "" {
+		return fmt.Errorf("rpc provider url is required")
 	}
 
 	needPrivateKey := flags&WithoutPrivateKey == 0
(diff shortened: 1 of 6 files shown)


jchappelow (approved):
LGTM.  If the folder name is changing, I think we can just make this provider setting switch without deprecating or detecting an old grpc_url setting.

Yaiba:
> LGTM. If the folder name is changing, I think we can just make this provider setting switch without deprecating or detecting an old grpc_url setting.

you mean like this ?
```
type kwilCliPersistedConfig struct {
        ...
	// Deprecated: use `Provider` instead
	GrpcURL  string `mapstructure:"grpc_url" json:"grpc_url,omitempty"`
	Provider string `mapstructure:"provider" json:"provider,omitempty"`
        ...
}
```

jchappelow:
> you mean like this ?

Right.  I had done just that in #679.  That also required a bunch of code to support backward compat, so I decided there's little point looking for the old field, particularly because it's already looking in a different folder (`.kwil-cli` instead of `.kwil_cli`)

Yaiba:
> > you mean like this ?
> 
> Right. I had done just that in #679. That also required a bunch of code to support backward compat, so I decided there's little point looking for the old field, particularly because it's already looking in a different folder (`.kwil-cli` instead of `.kwil_cli`)

Alright cool. 
 @brennanjl do you have concerns on this breaking changes?

brennanjl:
Yeah this seems fine to me. I don't think many people rely on the persisted config anyways. Fractal seems to pass everything with flags, and Truflation has built their own Go server for writing to Kwil.

brennanjl (approved):

PR #689, opened 2024-04-30
fix kgw client call action

Yaiba:
This pr 
- adds `Call` function to `GatewayClient`, so GatewayClient will do proper authn if required
- update KGWTest to test against procedure

### DIFF
--- a/core/gatewayclient/client.go
+++ b/core/gatewayclient/client.go
@@ -120,12 +120,18 @@ func NewClient(ctx context.Context, target string, opts *GatewayOptions) (*Gatew
 	return g, nil
 }
 
-// CallAction call an action. It returns the result records.  If authentication is needed,
-// it will call the gatewaySigner to sign the authentication message.
+// CallAction calls an action. It returns the result records.  If authentication is needed,
+// Deprecated: Use Call instead.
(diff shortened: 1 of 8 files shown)


test/integration/kgw_test.go
```diff
@@ -50,7 +50,7 @@ func TestKGW(t *testing.T) {
 
 			// When user deployed a database
 			// invalid kuneiform
-			specifications.DatabaseDeployInvalidSql1Specification(ctx, t, creatorDriver)
+			//specifications.DatabaseDeployInvalidSql1Specification(ctx, t, creatorDriver)
```
jchappelow, line 53: No need for this test now?

Yaiba: ah forgot I wanted to fix this, will try to include that test

jchappelow (approved):

core/gatewayclient/client.go
```diff
@@ -120,12 +120,18 @@ func NewClient(ctx context.Context, target string, opts *GatewayOptions) (*Gatew
 	return g, nil
 }
 
-// CallAction call an action. It returns the result records.  If authentication is needed,
-// it will call the gatewaySigner to sign the authentication message.
+// CallAction calls an action. It returns the result records.  If authentication is needed,
```
jchappelow, line 123: Oh, comment is funny now.

PR #691, opened 2024-05-01
core: kgw cookie on multi providers

Yaiba:
This pr change the saved cookie `key`, to support multi providers.

Current


Yaiba:
@KwilLuke I couldn't find how kwil-js handle this in nodejs, can check that?

cmd/kwil-cli/cmds/common/roundtripper.go
```diff
@@ -98,14 +98,19 @@ func DialClient(ctx context.Context, cmd *cobra.Command, flags uint8, fn RoundTr
 		return fn(ctx, client, conf)
 	}
 
-	cookie, err := LoadPersistedCookie(KGWAuthTokenFilePath(), clientConfig.Signer.Identity())
+	providerDomain, err := getDomain(conf.Provider)
```
Yaiba, line 101: When backport, `conf.Provider` should be `conf.GrpcUrl`

cmd/kwil-cli/cmds/common/authinfo.go
```diff
@@ -77,6 +78,26 @@ func convertToHttpCookie(c cookie) *http.Cookie {
 	}
 }
 
+// getDomain returns the domain of the URL.
+func getDomain(target string) (string, error) {
+	if target == "" {
+		return "", fmt.Errorf("target is empty")
+	}
+
+	parsedTarget, err := url.Parse(target)
```
jchappelow, line 87: One quirk of `url.Parse` is how it handles a string like `localhost:8080`.  It does not error.  It returns a URL with no host and "Opaque":

`&url.URL{Scheme:"localhost", Opaque:"8080", User:(*url.Userinfo)(nil), Host:"", Path:"", RawPath:"", OmitHost:false, ForceQuery:false, RawQuery:"", Fragment:"", RawFragment:""}`

We'll probably be rejecting or erroring on such a target _before_ we reach `getDomain`, but it's still probably an unintended result for `getDomain`:
- input `localhost:8080`
- output `localhost://`

jchappelow (commented):
All looks, fine, but I'm not certain why we have to use this computed domain (actually scheme://domain[:port]) when doing `SaveCookie`.  Does `http.Cookie.Domain` not persist in the cookie jar?  Or would that not be the exactly right key, missing scheme and probably port?

cmd/kwil-cli/cmds/common/authinfo.go
Yaiba: ah I got screwed every time I use `url.Parse` 🤦‍♂️

Yaiba:
> All looks, fine, but I'm not certain why we have to use this computed domain (actually scheme://domain[:port]) when doing `SaveCookie`. Does `http.Cookie.Domain` not persist in the cookie jar? Or would that not be the exactly right key, missing scheme and probably port?

`http.Cookie.Domain` will be empty if https + cookie.SameSite=None, see https://github.com/kwilteam/kgw/blob/53159c27d08286bb8bd463cc64c5cbb3abdbfe9a/cookie.go#L82.  So we cannot rely on it

cmd/kwil-cli/cmds/common/authinfo.go
```diff
@@ -84,6 +85,10 @@ func getDomain(target string) (string, error) {
 		return "", fmt.Errorf("target is empty")
 	}
 
+	if !(strings.HasPrefix(target, "http://") || strings.HasPrefix(target, "https://")) {
```
Yaiba, line 88: For 0.7 or v0.6, this should work, since we require scheme https://docs.kwil.com/docs/kwil-cli/configuration#available-configs. 

For v0.8, seems this could break?

Yaiba: you got a pretty handy util function here https://github.com/kwilteam/kwil-db/pull/703/files#diff-7a8b6c70656b28a8bba3d014ad27d1a48efe27f72f184939a4b9d7b5d2313a84R72, I think it will help.

jchappelow: With latest update to that PR, `cfg.Provider` must be a proper URL now.  Another reason to keep that convention I figure.

Yaiba: The persistened `cfg.Provider` could still be without scheme, am I right?

jchappelow: Not any more.  It's an error:

```
$  ./kwil-cli database list
Error: parse url: parse "127.0.0.1:8484": first path segment in URL cannot contain colon
parse url: parse "127.0.0.1:8484": first path segment in URL cannot contain colon
$  cat ~/.kwil-cli/config.json 
{"provider":"127.0.0.1:8484"}
```

I think we had that same constraint in v0.7?

jchappelow (approved):

cmd/kwil-cli/cmds/common/authinfo.go
Yaiba: We have some unpleasant constraint, like this :
```
.build/kwil-cli database list --kwil-provider localhost:8080                                                   
Error: wrap client: chain_info: Get "localhost:8080/api/v1/chain_info": unsupported protocol scheme "localhost"
wrap client: chain_info: Get "localhost:8080/api/v1/chain_info": unsupported protocol scheme "localhost"
```

It'll get error when create client, that's before `getDomain`, so I guess we're fine

Yaiba:
eh, backport workflow didn't run

PR #697, opened 2024-05-02
parse,tasks: reenable parse unit tests and tidy

jchappelow:
This re-enables unit tests and mod tidy for the reborn `parse` module.

On doing so, there is a failing unit test (`Test_Parse/all_possible_constraints` in `parse/kuniform`):

```
--- FAIL: Test_Parse (0.00s)
    --- FAIL: Test_Parse/all_possible_constraints (0.00s)
        parser_test.go:505: 
            	Error Trace:	/home/jon/kwil/git/kwil-db/parse/kuneiform/parser_test.go:505
            	Error:      	Received unexpected error:
            	            	parent table other_uses not found
            	Test:       	Test_Parse/all_possible_constraints
```

I've updated parse/kuneiform/parser_test.go with a make-it-pass mindset, but this is a very naive resolution.  I'm not sure why this test passed before or what the correct resolution is.  I do recall that the `Clean` methods now take a list of tables, which I think is for typing, but I'm not positive.  @brennanjl Suggested changes to fix this unit test, assuming what I did is incorrect?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -69,8 +69,8 @@ jobs:
 
       - name: Compile packages, apps, and specs
         run: | # enter workspace mode to do this on one line
-          go work init && go work use . ./test ./core
-          go build -mod=readonly ./... ./core/... ./test/specifications/
+          go work init . ./test ./core ./parse
+          go build -mod=readonly ./... ./core/... ./parse/... ./test/specifications/
 
       - name: Lint
         uses: golangci/golangci-lint-action@v4.0.0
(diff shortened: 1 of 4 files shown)


brennanjl:
Ah yeah I ran into this failing test, and it is fixed on my current branch.

The only nit is that in the `other_uses` table, age has to be unique, as foreign keys cannot reference non-unique columns. We do not currently test for this, however I can add this logic in this PR.

Once added, I think we can merge this

brennanjl:
Never mind, because this is a remote branch I cannot add to it.

We can merge as-is, and I will handle the logic (and negative tests for foreign keys) in a subsequent PR.

jchappelow:
Did it reject your push?  I have this set:

image

That normally lets maintainers push to the branch regardless of the repo.

jchappelow:
Oh, it is actually "uses" not "users".  I see

brennanjl:
Ah I didn't actually try, I just assumed that was not even supported.

brennanjl (approved):

PR #698, opened 2024-05-02
added better error handling for wasm parsing

brennanjl:
This PR adds better error handling in the parser, which makes it easier to build language tooling (e.g. using WASM or language servers). It allows us to not only identify syntax errors, but also type errors, invalid table, column, and function references, cartesian joins, and more by identifying the position where the error occurred. The positions are a bit rough sometimes; the ANTLR go target is pretty rough around the edges, and isn't always perfect.

# Parser Error Handling

This was implemented by marking each AST node (procedure, action, or sql ast nodes) with a position. For the top-level Kuneiform parser, I make a map that tracks the start and end position of each block (e.g. a table block, action block, procedure block). The AST positions are relative to where that statement appears in it's specific parser, and then composed within the error listener to find the final result. Take for example:

```go
database mydb;

procedure my_proc() public { // line 2 of database, line 0 of procedure
   $id int; // <- line 1 of procedure body + line 2 of db = line 3
   INSERT INTO ... // line 0 of SQL + line 2 of proc + line 2 of db = line 4
}
```

In this example, you can see that at each place where we enter a new parser, we remember where we left off in the old one. This is all encapsulated in the error listener, so that visitor implementations do not have to worry about returning the correct location. All they have to do is return the error to an error listener and it will properly handle it.

_You may notice that indexing starts at 0, and parsers / ides usually start at 1. For some reason, antlr-go indexes columns at 0 but lines at 1. I decided to normalize to 0 and then add 1 when errors are read because it is easier to track positions._


jchappelow:
> For some reason, antlr-go indexes columns at 0 but lines at 1.

:exploding_head:

internal/engine/integration/deployment_test.go
```diff
@@ -228,7 +228,7 @@ func Test_Deployment(t *testing.T) {
 				TxID:   "test",
 			})
 			if tc.err != nil {
-				require.ErrorIs(t, err, tc.err)
+				require.ErrorAs(t, err, &tc.err)
```
jchappelow, line 231: This will pass any `tc.err` value because it's just a regular `error err` and anything can be put in there.

The semantics of `errors.Is` that make sure `tc.err` has the correct error *value* to that in `err` is what we need.  The problem is that `CombineParseErrors` untypes any errors that were in the slice. `ParseErrors` `Err` method is a related issue as it really shouldn't exist since `ParseErrors` itself is an `error`.  The missing piece is implementing either `Unwrap() []error` or `Unwrap() error` for `ParseErrors`.

I'll suggest a revision shortly.

jchappelow (commented):
`ParseErrors` isn't working quite as intended.  Will suggest changes shortly.

internal/engine/integration/deployment_test.go
jchappelow: There's a little more to change than expected.  There are two listeners that need modification.  But also to keep wrapped errors detectable, `ParseError.Err` needs to become an `error` instead of a `string`, and then methods like this need to take an `error`:

```go
func (e *ErrorListener) NodeErr(node *Node, errType ParseErrorType, msg string) {
	e.Errs = append(e.Errs, &ParseError{
		ParserName: e.name,
		Type:       errType,
		Err:        msg,
		Node:       e.adjustNode(node),
	})
}
```

Because here's where the error's identity gets thrown out:

```go
t.errs.NodeErr(p0.GetNode(), types.ParseErrorTypeSemantic, fmt.Sprintf(`%s: "%s"`, types.ErrUnknownForeignProcedure.Error(), p0.Name))
```

Looking for a neat resolution, but I think every call to `NodeErr` also needs to change.

jchappelow: Pushed a resolution: https://github.com/kwilteam/kwil-db/pull/698/commits/820f6b6255f04b3e68182e30c842ad4cf9d122ca
Feel free to pop that off if you want to go a different way.  I think it's best to wrap the errors.

brennanjl: I appreciate the help here, thanks. Took a look at your commit, will keep it.

I always assumed that error typing was based on whether the error message contained the correct string.

jchappelow (approved):
LGTM.  Need a review from @Yaiba ?

brennanjl:
If he wants to, could be good. I think @KwilLuke should probably take a quick look, just at the error info returned to make sure it works for him.

Yaiba:
> LGTM. Need a review from @Yaiba ?

I'm gonna review it

KwilLuke:
@brennanjl, when i commit a SQL error like this:
image

It does not return this error in the json.errors part of the `parseKuneiform` response. Instead, it prevents any json from being returned, and instead returns on the old `error` field. E.g.: 
image

KwilLuke:
Another parser bug. This:
image

throws this:
image

It would be reasonable for this bug to be thrown in `json.errors`

parse/procedures/visitors/typing/typing.go
```diff
@@ -579,49 +628,68 @@ func (t *typingVisitor) analyzeProcedureCall(p0 parser.ICallExpression) *types.P
...
+			t.errs.NodeErr(p0.GetNode(), types.ParseErrorTypeSemantic, types.ErrVariableAlreadyDeclared)
+			return nil
 		}
 
 		t.declarations[p0.Variable] = r
+
+		// check that we are iterating over an array
+		dataType, ok := t.declarations[target.Variable.Name]
+		if !ok {
+			t.errs.NodeErr(p0.GetNode(), types.ParseErrorTypeSemantic,
```
Yaiba, line 658: I see this is not consistent with VisitExpressionVariable, you returned type error there

parse/procedures/visitors/typing/typing.go
```diff
@@ -718,73 +800,82 @@ func (t *typingVisitor) VisitStatementReturn(p0 *parser.StatementReturn) any {
...
 
 	if t.currentProcedure.Returns == nil {
-		panic("procedure does not return anything")
+		t.errs.NodeErr(p0.GetNode(), types.ParseErrorTypeSemantic, errors.New("procedure does not return anything"))
+		return nil
 	}
 
 	if !t.currentProcedure.Returns.IsTable {
-		panic("RETURN NEXT can only be used in procedures that return a table")
+		t.errs.NodeErr(p0.GetNode(), types.ParseErrorTypeSemantic, types.ErrReturnNextUsedInNonTableProc)
```
Yaiba, line 813: I feel like this is 'type' error?

parse/sql/sqlanalyzer/clean/walker.go
```diff
@@ -7,38 +7,67 @@ import (
...
-func NewStatementCleaner(schema *types.Schema) *StatementCleaner {
+func NewStatementCleaner(schema *types.Schema, errorListener parseTypes.NativeErrorListener) *StatementCleaner {
 	return &StatementCleaner{
 		AstListener: tree.NewBaseListener(),
 		schema:      schema,
+		errs:        errorListener,
 	}
 }
 
+var ErrTableNotFound = errors.New("table not found")
```
Yaiba, line 23: move this to errors.go ?

parse/sql/sqlanalyzer/typing/visitor.go
```diff
@@ -26,103 +28,85 @@ type typeVisitor struct {
...
-		if err != nil {
-			return err
-		}
+	return evalFunc(func(e *evaluationContext) {
+		relation := p0.Select.Accept(t).(returnFunc)(e)
 
 		_, ok := t.commonTables[p0.Table]
 		if ok {
-			return fmt.Errorf("common table expression conflicts with existing table %s", p0.Table)
+			t.options.ErrorListener.NodeErr(p0.GetNode(), parseTypes.ParseErrorTypeSemantic, fmt.Errorf("common table expression conflicts with existing table %s", p0.Table))
```
Yaiba, line 41: seems nbd, but `t.options.ErrorListener` feels a bit off to me, semantically.  It's option, should be used to overwrite your default `t.ErrorListener` .

parse/sql/sqlanalyzer/typing/visitor.go
```diff
@@ -562,260 +486,241 @@ func (t *typeVisitor) VisitExpressionColumn(p0 *tree.ExpressionColumn) any {
...
 			esc := p0.Escape.Accept(t).(attributeFn)
-			et, err := esc(ev)
-			if err != nil {
-				return nil, err
-			}
+			et := esc(ev)
 
 			if !et.Type.Equals(types.TextType) {
-				return nil, fmt.Errorf("%w: string comparison expected text. Received: %s", ErrInvalidType, et.Type.String())
+				t.options.ErrorListener.NodeErr(p0.GetNode(), parseTypes.ParseErrorTypeType, fmt.Errorf("string comparison escape expected text. Received: %s", et.Type.String()))
```
Yaiba, line 696: escape?

test/specifications/utils.go
```diff
@@ -25,8 +26,8 @@ type FileDatabaseSchemaLoader struct {
 
 // parseKuneiformWithoutValidation parses a Kuneiform schema without performing
 // any validation checks.
-func parseKuneiformWithoutValidation(kf string) (*types.Schema, error) {
-	schema, err := kuneiform.Parse(kf)
+func parseKuneiformWithoutValidation(kf string) (*parse.ParseResult, error) {
```
Yaiba, line 29: This logic doesn't look right now, `ParseKuneiform` will perform all checks

Yaiba (commented):
Over all looks good to me

KwilLuke:
Another one - 

This: 
image

Gives this:
image

KwilLuke:
Another - 

Schema:
image

Error:
image

KwilLuke:
> Another -
> 
> Schema: image
> 
> Error: image

FYI - this happens whenever there is a type in a table that is invalid.

KwilLuke:
This: 
image

Gives:
image

KwilLuke:
This:
image

Gives this:
image

KwilLuke:
This: 
image

Throws:
image

KwilLuke:
Not sure if this is smthn you can fix:

In this procedure, I am missing a comma here, but it throws a confusing error on $address:
image

image

Adding the comma after `age int` makes the error go away, but the misplacement is confusing.

KwilLuke:
Also not sure if this is something you can fix.

Here, my error is that there there is no column called post_name (it should be post_num):

image

However, the error is thrown one line too high:

image

KwilLuke:
This:

image


Gives:
image

parse/procedures/parser/visitor.go
```diff
@@ -55,12 +45,14 @@ func (p *proceduralLangVisitor) VisitForeign_call(ctx *gen.Foreign_callContext)
 
 	dbid := ctx.GetDbid()
 	if dbid == nil {
-		panic("missing dbid")
+		// this should get caught by the parser
```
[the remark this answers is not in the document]
brennanjl: yeah, just a sanity check

parse/procedures/parser/visitor.go
```diff
@@ -144,24 +146,29 @@ func (p *proceduralLangVisitor) VisitExpr_boolean_literal(ctx *gen.Expr_boolean_
...
+	switch v := e.(type) {
+	case *ExpressionCall:
+		v.Set(ctx)
+		v.TypeCast = tc
+	case *ExpressionForeignCall:
+		v.Set(ctx)
+		v.TypeCast = tc
+	default:
+		// should never happen
+		panic(fmt.Sprintf("invalid type cast for %T", e))
```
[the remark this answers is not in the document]
brennanjl: No, if we never expect them to happen I put a panic, since it is an internal error. Normally I would just return an error, but visitors cannot take errors, so I panic and catch it

cmd/kwil-cli/cmds/utils/parse.go
```diff
@@ -0,0 +1,60 @@
...
+			file, err := os.ReadFile(args[0])
+			if err != nil {
+				return display.PrintErr(cmd, err)
+			}
+
+			res, err := parse.ParseKuneiform(string(file))
+			if err != nil {
+				return display.PrintErr(cmd, err)
+			}
+
```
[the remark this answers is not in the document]
brennanjl: No, we check the `res.Err()` when we format. Known parsing errors (syntax/type errors) should get returned, but should not brick the whole parser. We format them so that consumers can highlight it.

parse/sql/sqlanalyzer/typing/visitor.go
brennanjl: The options struct sets a default if none are set. Will make it more obvious

KwilLuke:
This:
image

Throws this:
image

parse/sql/sqlanalyzer/typing/visitor.go
brennanjl: yeah, SQL string comparisons allow escape clauses for escaping variables from the comparison

KwilLuke:
If i start typing the database declaration, the parser thinks that is the database name:
image

image

It would be better if nothing is returned in the `schema.name` field while i am declaring the database.

KwilLuke:
This:
image

Gives this:
image

jchappelow (approved):
LGTM.  Sounds like bugs can be squashed in followup.  Squash and merge when ready!  :rocket:

KwilLuke:
This:
image

gives this:
image

PR #699, opened 2024-05-02
config flag to specify the Kwil cli persistent settings file

charithabandi:


brennanjl (approved):

cmd/kwil-cli/config/flags.go
```diff
@@ -45,9 +49,11 @@ func BindGlobalFlags(fs *pflag.FlagSet) {
 	fs.String(globalPrivateKeyFlag, cliCfg.PrivateKey, "the private key of the wallet that will be used for signing")
 	fs.String(globalProviderFlag, cliCfg.Provider, "the Kwil rpc provider HTTP endpoint")
 	fs.String(globalChainIDFlag, cliCfg.ChainID, "the expected/intended Kwil Chain ID")
+	fs.StringVar(&configFile, globalConfigFileFlag, defaultConfigFile, "the path to the Kwil CLI persistent global settings file")
 
 	// Bind flags to viper, named by the flag name
 	viper.BindPFlag(viperPrivateKeyName, fs.Lookup(globalPrivateKeyFlag))
 	viper.BindPFlag(viperProviderName, fs.Lookup(globalProviderFlag))
 	viper.BindPFlag(viperChainID, fs.Lookup(globalChainIDFlag))
+	viper.BindPFlag(viperConfigFile, fs.Lookup(globalConfigFileFlag))
```
jchappelow, line 58: Do we still need to bind a viper setting to the flag?  I figured defining `viperConfigFile` wouldn't be necessary as I think those are for settings that appear in the config file?  Haven't tested; perhaps I misunderstand.

cmd/kwil-cli/config/config.go
```diff
@@ -152,11 +152,11 @@ func LoadCliConfig() (*KwilCliConfig, error) {
 		return nil, err
 	}
 
-	// NOTE: defaultConfigFile is set in init() in flags.go
-	// read default config file if it exists
-	// and override viper values from config file
-	if fileExists(defaultConfigFile) {
-		viper.SetConfigFile(defaultConfigFile)
+	// If the config file exists, override viper values from the config file
+	// The config file is set thrugh the flag and has the default value as
```
jchappelow, line 156: thrugh => through,
but only if you have another change to push, otherwise nbd

cmd/kwil-cli/config/flags.go
charithabandi: let me verify this

PR #700, opened 2024-05-03
Feature/kgw jsonrpc

### LINKED ISSUE #694: json-rpc: gateway client support
In addition to [updates to `kgw`](https://github.com/kwilteam/kgw/issues/60), the `gatewayclient` in `kwil-db` needs to be updated with a pattern like with `WrapClient` used in the main client so that the underlying implementation can be swapped between the new json-rpc client and the old openapi http client.

Yaiba:
This resolves #694 

This add a new kgw jsonrpc client, which will do all interactions with GKW using json rpc.


Yaiba:
After some dev on this issue, I realized that we might not want to support change the impl between jsonrpc and REST in gatewayclient, as it will require two set of configurations(kgw API can support both, but it requires different configuration of kwil-db REST/jsonrpc backend)

core/rpc/client/gateway/jsonrpc/client.go
```diff
@@ -0,0 +1,188 @@
...
+	"reflect"
+	"strconv"
+	"sync/atomic"
+
+	rpcClient "github.com/kwilteam/kwil-db/core/rpc/client"
+	"github.com/kwilteam/kwil-db/core/rpc/client/gateway"
+	jsonrpc "github.com/kwilteam/kwil-db/core/rpc/json"
+)
+
+const (
```
Yaiba, line 19: Maybe should move to `core/rpc/json` ?

core/rpc/client/gateway/jsonrpc/client.go
```diff
@@ -0,0 +1,188 @@
...
+	// Then we will first check if we recognize the error(KGW) code, if not,
+	// we use client.clientError
+
+	rpcErr := &rpcClient.RPCError{
+		Msg:  jsonRPCErr.Message,
+		Code: int32(jsonRPCErr.Code),
+	}
+	err := errors.Join(jsonRPCErr, rpcErr)
+
+	switch jsonRPCErr.Code {
```
Yaiba, line 177: It seems we can delete this block(in every client)? IMO the `jsonRPCErr.Code` is very specific, and more useful for app? @jchappelow

core/rpc/client/gateway/jsonrpc/client.go
Yaiba: maybe not, admin and function methods are in `core/rpc/json` because it's also used in JSON rpc service server.

core/gatewayclient/client.go
```diff
@@ -137,10 +144,18 @@ func (c *GatewayClient) Call(ctx context.Context, dbid string, action string, in
 	if err == nil {
 		return res, nil
 	}
-	if !errors.Is(err, rpcClient.ErrUnauthorized) {
-		return nil, err
+
+	var jsonRPCErr *jsonrpc.Error
```
Yaiba, line 148: should prefer this way of handling error or below?

core/rpc/client/gateway/jsonrpc/client.go
jchappelow: I think that the highest level clients want more abstraction i.e. shouldn't have to know about the jsonrpc package and codes at all.  Also `errors.Is(err, <some error kind in core/client package>)` is better for a user of `core/client.Client` I think.

jchappelow: Codes could move up to a package that's not jsonrpc specific?  Or core/client could give a error code getter to extract one from this opaque `error`?

core/rpc/client/gateway/jsonrpc/client.go
```diff
@@ -0,0 +1,186 @@
...
+// GetAuthnParameter returns the auth parameter for the client.
+func (g *Client) GetAuthnParameter(ctx context.Context) (*gateway.AuthnParameterResponse, error) {
+	res := &gateway.AuthnParameterResponse{}
+	err := g.call(ctx, gateway.MethodAuthnParam, &gateway.AuthnParameterRequest{}, res)
+	return res, err
+}
+
+// call sends a JSON-RPC request to the gateway.
+// NOTE: this is basically a copy of core/rpc/client/user/jsonrpc/client.call
+func (cl *Client) call(ctx context.Context, method string, cmd, res any) error {
```
jchappelow, line 76: In the other client we can export and call this `CallMethod`, so that gateway client can embed it perhaps?   We just want to add the two auth related methods right?

core/rpc/client/gateway/jsonrpc/client.go
jchappelow: Oh. this is in the jsonrpc package.  Hmm.  I guess core/client should be checking codes an translating to Go error types...

core/gatewayclient/client.go
```diff
@@ -137,10 +144,18 @@ func (c *GatewayClient) Call(ctx context.Context, dbid string, action string, in
...
+	if errors.As(err, &jsonRPCErr) {
+		if jsonRPCErr.Code != jsonrpc.ErrorKGWNotAuthorized {
```
jchappelow, line 150: Ah.  Yup! `core/gatewayclient` can indeed interpret based on the jsonrpc code.

core/gatewayclient/client.go
Yaiba: ah just saw you commit

core/rpc/client/gateway/jsonrpc/client.go
Yaiba: You mean in core/client and core/gatewayclient ? yeah maybe, I just see those errors are not used much really, most of the time we just pop it up

core/rpc/client/gateway/jsonrpc/client.go
Yaiba: yeah I thought about it and totally forgot about it. I think we should do it. I copied bcz that's when we have two different branch , and my branch is based on another branch..

core/gatewayclient/client.go
Yaiba: Yeah, this is why I have the question https://github.com/kwilteam/kwil-db/pull/700#discussion_r1594780356.

core/rpc/client/gateway/jsonrpc/client.go
Yaiba: Alright I'll make the change

core/rpc/client/gateway/jsonrpc/client.go
Yaiba: Haven't gotten a solid solution yet, I'll leave it for now

core/rpc/client/jsonrpc.go
```diff
@@ -0,0 +1,214 @@
...
+//
+//  - the structs marshal into json objects, set in "params" field of request.
+//    with the old http gateway, endpoints were defined with associated
+//    message (pb) types sent in POST, or just GET with endpoint implying method.
+//  - the "method" in the outer request type is instead of the endpoint, all POST
+
+// JSONRPCClient is a JSON-RPC client that handles sending and receiving
+// JSON-RPC requests and responses. It is a low-level client that does not
+// care about the specifics of the JSON-RPC methods or responses.
+type JSONRPCClient struct {
```
Yaiba, line 44: Well the naming is kind awkward, I tempted to name it like `BaseRPCClient`.  The reason is in the future say we support WS, we might have another base client

core/gatewayclient/client.go
```diff
@@ -12,25 +12,25 @@ import (
...
-	gatewayTypes "github.com/kwilteam/kwil-db/core/types/gateway"
 )
 
 // GatewayClient is a client that is made to interact with a kwil gateway.
 // It inherits the functionality of the main Kwil client, but also provides
 // authentication cookies to the gateway.
 // It automatically handles the authentication process with the gateway.
 type GatewayClient struct {
-	client.Client
+	client.Client // core client
```
Yaiba, line 28: GatwayClient still needs this type embeded, I cannot move it to `core/rpc/gateway/`, as `core/rpc` are just client for handling exact input/output of APIs, but `core/client`  deals with user input/outputs.

core/types/client/opts.go
```diff
@@ -45,13 +49,18 @@ func (c *Options) Apply(opts *Options) {
 		c.ChainID = opts.ChainID
 	}
 
+	if opts.Conn != nil {
```
Yaiba, line 52: Just a thought on the style.

TBH, I don't like this way of passing options, i.e. `Client(xx, optionObj)`, I prefer `Client(xx, WithOptA, WithOptB...)`, and I kind of want to be able to apply those options to all our clients.
Currently we can reuse options though embed(like core/gatewayclient/opts.GatewayOptions`) , I'm not sure how to do it in `Client(xx, WithOptA...)` style, maybe using interface.

core/rpc/http/function/client.go
```diff
@@ -52,7 +52,7 @@ type service struct {
 }
 
 // NewAPIClient creates a new API client. Requires a userAgent string describing your application.
-// optionally a custom http.Client to allow for advanced features such as caching.
+// optionally a custom http.JSONRPCClient to allow for advanced features such as caching.
```
Yaiba, line 55: ehh sometime my IDE suggests too much refactor... Will change it back

core/rpc/client/user/jsonrpc/client.go
```diff
@@ -1,211 +1,238 @@
...
-//	- unmarshal the "result" (json.RawMessage) into the method's result type
-
-// note compared to http grpc gateway:
-//
-//  - the structs marshal into json objects, set in "params" field of request.
-//    with the old http gateway, endpoints were defined with associated
-//    message (pb) types sent in POST, or just GET with endpoint implying method.
-//  - the "method" in the outer request type is instead of the endpoint, all POST
-
+// Client is a JSON-RPC client for the Kwil user service. It use the JSONRPCClient
```
jchappelow, line 21: If we keep this file named "methods.go" (and squash all commits in this PR), git detects the rename of and the diff for both that file and methods.go is minimized.

i.e.

```
-------------------------- core/rpc/client/jsonrpc.go --------------------------
similarity index 81%
rename from core/rpc/client/user/jsonrpc/client.go
rename to core/rpc/client/jsonrpc.go
index cecb7df8..2059a708 100644
```

and core/rpc/client/user/jsonrpc/methods.go will have about a dozen lines changed.

Yaiba: cool, I was totally not thinking about this. will change .

core/types/client/opts.go
jchappelow: I don't have strong feelings about this, but the functional options pattern is generally more cumbersome and time consuming to use compared to a config struct.  Yet most of our code does functional options, so might as well here.

In general, I think the main benefit of functional options is on the highest level exported APIs because a config struct changing is technically a breaking change to the constructor while new option functions are not.   With how unstable our Go APIs are all around, this benefit is almost non-existent.

Ironically, the most benefit for functional options is in `core/client.NewClient`, but the transport-aware clients deeper in core/rpc/client/... benefit less from doing it that way.

Anyway, I don't mind any options refactoring you'd like to do, but the more internal things get, the less important it is to use functional options.

core/rpc/client/jsonrpc.go
jchappelow: Meh.  But if we support websocket, that's still json-rpc, just using websocket.  Anyway, I think `JSONRPCClient` is appropriate, at least while it is defined in `core/rpc/client`.

jchappelow (approved):
I think this "base client" / `JSONRPCClient` approach does the trick nicely.
I think you're still working on some changes though?

core/rpc/client/user/jsonrpc/client.go
Yaiba: Thanks for reminding me this aspect as well

core/types/client/opts.go
Yaiba: Yeah, just have this thought so it's consistent everywhere.  
I can work with current approach, at least all exposed clients are using config struct, it's consistent.

Yaiba:
> I think this "base client" / `JSONRPCClient` approach does the trick nicely. I think you're still working on some changes though?

I've done what I wanted. Will squash and merge.

core/gatewayclient/client.go
jchappelow: In this package, this makes sense.  For the _consumer_ of core/client or core/gatewayclient, they don't _need_ know json rpc codes.  If they want to introspect to see if it's a  json-rpc error that's fine, but it's preferable for external developers to be able to do a simple errors.Is with one of our predefined types.  At least that's how I feel for now.

PR #703, opened 2024-05-06
rpc: admin+functions client and server

jchappelow:
Quite half baked still, but waned to put up the basic structure of things for @Yaiba .  Function service added too.

Will write a PR proper description and clean up the code shortly.


jchappelow:
Enabling for CI, but need to clean up a bunch before review.

cmd/kwil-admin/cmds/node/version.go
```diff
@@ -12,7 +12,7 @@ var (
 	versionLong = `Print the node's version information. The version is the Kwil's version string, set at compile time.`
 
 	versionExample = `# Print the node's version information
-kwil-admin node version --rpcserver localhost:50151 --authrpc-cert "~/.kwild/rpc.cert"`
+kwil-admin node version --rpcserver /tmp/kwild.socket`
```
Yaiba, line 15: nbd, seems you gave `--rpcserver` to different example values in other cmds.  Maybe use same example value?

internal/services/jsonrpc/methods.go
```diff
@@ -7,63 +7,43 @@ import (
...
+// MethodHandler is a type of function that returns an interface containing a
+// pointer to a handler's input arguments, and a handler function that captures
+// the arguments pointer. The handler function returns its result type in an
+// interface, and a *jsonrpc.Error. A simple MethodHandler would instantiate a
+// new concrete instance of the parameters type and define a function that uses
+// that instance to perform some operations.
+type MethodHandler func(ctx context.Context, s *Server) (argsPtr any, handler func() (any, *jsonrpc.Error))
+
+// Svc is a type that enumerates it's handler functions by method name. To
+// handle a method, the Server:
```
Yaiba, line 19: 👍

Yaiba (approved):
LGTM

core/types/admin/types.go
```diff
@@ -39,6 +39,8 @@ type ValidatorInfo struct {
 	Power  int64          `json:"power"`
 }
 
+// type ValidatorInfo = types.Validator
+
```
jchappelow, line 43: I need to sort through the duplicated types here an in core/types.  I don't recall why we have this core/types/admin package while many of the types exist in core/types already.

.golangci.yml
```diff
@@ -1,6 +1,8 @@
 run:
-  deadline: 10m
-  skip-dirs:
+  timeout: 10m
+
+issues:
+  exclude-dirs:
```
jchappelow, line 5: The golangci-lint config fields changed a bunch.  If you run it on your local machine, make sure to update it to at least 1.58.

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -118,6 +118,16 @@ pg_db_pass = "{{ .AppCfg.DBPass }}"
 # PostgreSQL database name (override database name, default is "kwild")
 pg_db_name = "{{ .AppCfg.DBName }}"
 
+# The admin RPC server can require a password, if set. Ensure the connection is
+# encrypted since the password is sent unencrypted in the HTTP Authorization
+# header. Not needed if client authentication is done with mutual TLS (clients.pem).
+# admin_pass = "{{ .AppCfg.AdminRPCPass }}"
```
jchappelow, line 124: Oh, I haven't tested this password based auth.  Was just an idea I wanted to try since it's simpler than TLS key pairs if a deployment happens to have a secured network or using TLS with a root CA (not self-signed) an terminated at the edge of a secure network.

cmd/kwild/config/config.go
```diff
@@ -464,6 +467,7 @@ func LoadConfigFile(configPath string) (*KwildConfig, error) {
 // LoadEnvConfig loads a config from environment variables.
 func LoadEnvConfig() (*KwildConfig, error) {
 	viper.SetEnvPrefix("KWILD")
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_")) // --app.output-paths => KWILD_APP_OUTPUT_PATHS
```
jchappelow, line 470: Whoops, was just testing this.  Will remove.

jchappelow:
@Yaiba @KwilLuke I think that @Yaiba talked me into just requiring the URL like before, no hostport excpetions.  I've hidden the `/rpc/v1` business away from top level clients.  https://github.com/kwilteam/kwil-db/pull/703/commits/e4012c3848014df633617cbdc47e92c59f107dca  I think this will work all around:

```
./kwil-cli database list --provider http://127.0.0.1:8484
```

Basically just like before.  Only awareness of the `/rpc/v1` path part are the jsonrpc-specific packages on both server and client.  This means that it's enforced, and clients will always append that so there's no hiding it with deployment and load balancer tricks.  This is no different from before with the `/api/v1` part of the path.  It will just pass through a load balancer.  

This does however mean:
1. if a load balancer/proxy config wants to selectively pass ONLY RPC stuff, they have to recognize /rpc/v1 only and block everything else (as opposed to forwarding `/` to `/rpc/v1` on the backeneds)
2. raw use with say `curl` must always include that `/rpc/v1` on account of (1) above

Yaiba:
> @Yaiba @KwilLuke I think that @Yaiba talked me into just requiring the URL like before, no hostport excpetions. I've hidden the `/rpc/v1` business away from top level clients. [e4012c3](https://github.com/kwilteam/kwil-db/commit/e4012c3848014df633617cbdc47e92c59f107dca) I think this will work all around:
> 
> ```
> ./kwil-cli database list --provider http://127.0.0.1:8484
> ```
> 
> Basically just like before. Only awareness of the `/rpc/v1` path part are the jsonrpc-specific packages on both server and client. This means that it's enforced, and clients will always append that so there's no hiding it with deployment and load balancer tricks. This is no different from before with the `/api/v1` part of the path. It will just pass through a load balancer.
> 
> This does however mean:
> 
> 1. if a load balancer/proxy config wants to selectively pass ONLY RPC stuff, they have to recognize `/rpc/v1` only and block everything else (as opposed to forwarding `/` to `/rpc/v1` on the backeneds)
> 2. raw use with say `curl` must always include that `/rpc/v1` on account of (1) above

I'm fine with what it was,  this new changes also works for kgw.
For the context of KGW, it routes `/rpc/v1` to its JSONRPC handler; the configuration of backend(of upstream) though, will require no path, i.e. `http://host:port`

Yaiba (approved):
LGTM

core/rpc/json/responses.go
```diff
@@ -8,11 +8,20 @@ import (
...
+	Patch       uint32 `json:"patch"`
+	KwilVersion string `json:"kwil_ver"`
+}
+
 // AccountResponse contains the response object for MethodAccount.
 type AccountResponse struct {
-	Identifier []byte `json:"identifier,omitempty"`
-	Balance    string `json:"balance"`
-	Nonce      int64  `json:"nonce"`
+	Identifier types.HexBytes `json:"identifier,omitempty"`
```
KwilLuke, line 22: I just noticed that calling user.account only returns `{balance, nonce}`, now, and no longer includes the identifier. Is this intentional?

E.g. calling with postman:
`data: { jsonrpc: '2.0', id: 1, result: { balance: '0', nonce: 0 } }`

jchappelow: For an account that does not exist, yes, that was intentional.  Can change though.  I also though it was pointless to return identifier that you provided in the request?

jchappelow: Raises other questions.  Discussing in https://kwilteam.slack.com/archives/C05RZNFKNJD/p1715285006767559?thread_ts=1715284990.944059&cid=C05RZNFKNJD

Regardless of this response, I think now is a good time to change anything for the better since we're breaking quite a bit.  Only reason to minimize changes is to keep kwil-cli with `--output json` showing similar responses, but that's fairly niche I think.

jchappelow:
Should first merge https://github.com/kwilteam/proto/pull/35 and update the git submodule to the new proto main after that.  Any objections to removing unused and old proto service definitions?

Yaiba:
> Should first merge [kwilteam/proto#35](https://github.com/kwilteam/proto/pull/35) and update the git submodule to the new proto main after that. Any objections to removing unused and old proto service definitions?

Yes we should remove them.  it's already not used even for v0.6 I think

Yaiba (approved):

jchappelow:
OK, gonna merge.  But two things shaking out:
- the `TransactionBody.Fee *big.Int` field marshals/unmarshals to a number, not a string.  My minunderstanding.  Will find a resolution that fixes json but also doesn't break RLP
- I added a basic auth mechanism with a password to the admin service (for use with secure connections) but need to make ensure that's working right or remove that functionality.

PR #678, opened 2024-04-24
[v0.7.5]  update gomod for core module version bump

Yaiba:
This pr is going to update go.mod which deps on core/v0.1.3(to tag, I'll tag it once I'm sure no more changes will be made to core module).

PR #705, opened 2024-05-07
# Load env variables into KwildConfig

charithabandi:
Env variables are not supported since we renamed the cmd line flags to have hyphens or dashes instead of underscores. Whereas the struct tags remained to use dashes to support config file parsing. Therefore manual binding is needed to map the env variables correctly to the structures.

### DIFF
--- a/cmd/kwild/config/config.go
+++ b/cmd/kwild/config/config.go
@@ -463,15 +463,28 @@ func LoadConfigFile(configPath string) (*KwildConfig, error) {
 
 // LoadEnvConfig loads a config from environment variables.
 func LoadEnvConfig() (*KwildConfig, error) {
-	viper.SetEnvPrefix("KWILD")
-	viper.AutomaticEnv()
+	// Manually bind environment variables to viper keys.
+	for _, key := range viper.AllKeys() {
+		// Replace dashes with underscores in the key to match the flag name.
+		// This is required because there is inconsistency between our flag names
+		// and the struct tags. The struct tags use underscores, but the flag names
+		// use dashes. Viper uses the flag names to bind environment variables
+		// and this conversion is required to map it to the struct fields correctly.
+		bindKey := strings.ReplaceAll(key, "-", "_")
+		envKey := "KWILD_" + strings.ToUpper(strings.ReplaceAll(bindKey, ".", "_"))
+		viper.BindEnv(bindKey, envKey)
+	}
 
-	var cfg KwildConfig
-	if err := viper.Unmarshal(&cfg); err != nil {
+	// var cfg KwildConfig, won't work because, viper won't be able to extract
+	// the heirarchical keys from the config structure as fields like cfg.app set to nil.
+	// It can only extract the first level keys [app, chain, log] in this case.
+	// To remedy that, we use DefaultEmptyConfig with all the sub fields initialized.
+	cfg := DefaultEmptyConfig()
+	if err := viper.Unmarshal(cfg); err != nil {
 		return nil, fmt.Errorf("decoding config: %v", err)
 	}
 
-	return &cfg, nil
+	return cfg, nil
 }
 
 var ErrConfigFileNotFound = fmt.Errorf("config file not found")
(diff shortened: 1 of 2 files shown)


cmd/kwild/config/config.go
```diff
@@ -466,12 +466,23 @@ func LoadEnvConfig() (*KwildConfig, error) {
...
-	var cfg KwildConfig
-	if err := viper.Unmarshal(&cfg); err != nil {
+	for _, key := range viper.AllKeys() {
+		// Replace dashes with underscores in the key to match the flag name.
+		// This is required because there is inconsistency between our flag names
+		// and the struct tags. The struct tags use underscores, but the flag names
+		// use dashes. Viper uses the flag names to bind environment variables
+		// and this conversion is required to map it to the struct fields correctly.
+		bindKey := strings.ReplaceAll(key, "-", "_")
+		envKey := "KWILD_" + strings.ToUpper(strings.ReplaceAll(bindKey, ".", "_"))
```
jchappelow, line 476: Bummer, so is this above now pointless?

```
viper.SetEnvPrefix("KWILD")
viper.AutomaticEnv()
```

charithabandi: Just pasting the comment here for reference: 
`/*
		Lots of Viper magic here, but the gist is:
		We want to be able to set config values via
			-  flags
			-  environment variables
			-  config file
			-  default values
		for env variables support:
		Requirement is, we need to be able to config from env variables with a prefix "KWILD_"
		It can be done 2 ways:
		1. AutomaticEnv: off mode
			- This will not bind env variables to config values automatically
			- We need to manually bind env variables to config values (this is what we are doing currently)
			- As we bound flags to viper, viper is already aware of the config structure mapping,
				so we can explicitly call viper.BindEnv() on all the keys in viper.AllKeys()
			- else we would have to reflect on the config structure and bind env variables to config values
		2. AutomaticEnv: on mode
			- This is supposed to automatically bind env variables to config values
				(but it doesn't work without doing a bit more work from our side)
			- One way to make this work is add default values using either viper.SetDefault() for all the config values
			  or can do viper.MergeConfig(<serialized config>)
			- Serializing is really painful as cometbft has a field which is using map<interface><interface> though its deprecated.
				which prevents us from doing the AutomaticEnv binding
		Issues referencing the issues (or) correct usage of AutomaticEnv: https://github.com/spf13/viper/issues/188
		For now, we are going with the first approach
		Note:
		The order of preference of various modes of config supported by viper is:
		explicit call to Set > flags > env variables > config file > default values
	*/`

jchappelow: I thought that `map[string]any` issue is gone since we made all of our own structs instead of using cometbft's.

Anyway, I don't understand the viper magic at all, so whatever works is fine.  Remove those two lines and comment on why they don't have any purpose?

cmd/kwild/config/config.go
```diff
@@ -481,6 +492,22 @@ func fileExists(path string) bool {
...
+		},
+		ChainCfg: &ChainConfig{
+			P2P:       &P2PConfig{},
+			RPC:       &ChainRPCConfig{},
+			Mempool:   &MempoolConfig{},
+			StateSync: &StateSyncConfig{},
+			Consensus: &ConsensusConfig{},
+		},
+		Logging: &Logging{},
+	}
```
jchappelow, line 508: Will you comment on why this is necessary instead of `var cfg KwildConfig`?

cmd/kwild/config/config.go
charithabandi: Oh yeah! I totally forgot about that, Yeah, should be doable with automatic env, with serialization. But its still broken bcz of the inconsistency in the tags.

jchappelow: I think your solution is fine, but have a look at `viper.SetEnvKeyReplacer`.  I think it's built for this purpose like `viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))`.  But I really am just guessing here, not a test!

charithabandi: umm, tried it out, didn't work for me.

jchappelow:
After this goes in a backport PR for v0.7.5 we can re-open https://github.com/kwilteam/kwil-db/pull/678 

I'll do a quick test on this PR now but it looks fine.

jchappelow (approved):
LGTM.  Tested with `KWILD_APP_PG_DB_NAME=wrong`

PR #707, opened 2024-05-08
ci: add backport workflow

Yaiba:
This add backport workflow.

Requirements:

Screenshot 2024-05-08 at 2 29 48 PM

How to: tag the PR you want to backport, it always starts with `backport-to-BRANCH`. So if we want to backport to v0.7, the tag will be `backport-to-release-v0.7`. After tagged PR is merged, a backport pr will be created automatically.

---

UPDATE: a demo

https://github.com/Yaiba/test-relase-based-workflow/pull/9, this is the original pr;
after applying label backport-to-release-v0.1 to the pr,  once it’s merged, it creates a new pr https://github.com/Yaiba/test-relase-based-workflow/pull/10

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,25 @@
+name: backport
+
+on:
+  pull_request_target:
+    types:
+      - closed
+
+jobs:
+  backport:
+    name: Backport PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Backport Action
+        uses: sorenlouv/backport-github-action@v9.3.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          auto_backport_label_prefix: backport-to-
+
+      - name: Info log
+        if: ${{ success() }}
+        run: cat ~/.backport/backport.info.log
+
+      - name: Debug log
+        if: ${{ failure() }}
+        run: cat ~/.backport/backport.debug.log
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Backport Action
-        uses: sorenlouv/backport-github-action@v9.3.0
+        uses: sorenlouv/backport-github-action@f7073a2287aefc1fa12685eb25a712ab5620445c #v9.3.1
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           auto_backport_label_prefix: backport-to-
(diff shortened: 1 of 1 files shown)


Yaiba:
@jchappelow is this `workflow permissions` ok to you? I'm positive, as this `backport` workflow only will create PR, as for the `write permission`, as long as we control what workflow get merge in to main, we're good.

jchappelow:
> @jchappelow is this `workflow permissions` ok to you? I'm positive, as this `backport` workflow only will create PR, as for the `write permission`, as long as we control what workflow get merge in to main, we're good.

Yeah totally good.

jchappelow (approved):

.github/workflows/backport.yml
```diff
@@ -0,0 +1,25 @@
...
+    types:
+      - closed
+
+jobs:
+  backport:
+    name: Backport PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Backport Action
+        uses: sorenlouv/backport-github-action@v9.3.0
```
jchappelow, line 14: Can you bump to v9.3.1?  The node 20 update is there.

.github/workflows/backport.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
jchappelow, line 16: Do we have to worry about sorenlouv (hypothetically) stealing or secret token?

Yaiba: lol that's a valid concern, maybe we should fork it?

jchappelow: Looks fairly trusted.  Meh.  I think we should have 
`uses: sorenlouv/backport-github-action@f7073a2287aefc1fa12685eb25a712ab5620445c # v9.3.0`
instead to pin to a commit then.  Less prone to unwitting hijacks.

Yaiba: It seems this `github_token` will be whoever merges this pr?

jchappelow: Oh, I meant to use the commit hash if we don't fork.  If we fork then tags are under our control  :) NBD
I'm just talking through security concerns since I don't really know how the action works.
I suspect this guy's project is fine if we pin by commit hash instead of something changeable like a tag or latest or whatever.

Yaiba: haha, just incase the repo could be gone...

jchappelow (approved):

Yaiba:
According to https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#label , it's correct that we removed `labeled`

PR #708, opened 2024-05-09
Stats monitor for stress tool

charithabandi:
Stats monitor tool, that can connect to list of chain rpc servers and gather Block and mempool stats and does basic analysis on it. 

```bash

./stats -h
Usage of ./stats:
  -output string
        stats file name (default "stats.json")
  -rpcservers string
        comma separated list of  chain rpc servers to query stats from (default "http://localhost:26657")

```


test/stress/load-test/main.go
```diff
@@ -42,6 +42,8 @@ var (
 	nonceChaos          int
```
jchappelow, line 42: I kinda want the stress tool to compile to a binary named `stress`.
`test/stress/stats` doesn't really fit.

Maybe
`test/stress` -- stress tool
`test/statsmon` -- stats collection tool.

Or put everything under `test/cmd`: `test/cmd/stress` and `test/cmd/stats`?

test/stress/stats/main.go
```diff
@@ -0,0 +1,70 @@
...
+	statsMonitor, err := newStatsMonitor(addresses, statsFile, logger)
+	if err != nil {
+		logger.Error("failed to create stats monitor", log.Error(err))
+		os.Exit(1)
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	go statsMonitor.Start(ctx)
+	defer statsMonitor.Stop()
```
jchappelow, line 47: A couple things to note: `defer`s will not run after an `os.Exit`.  That's a hard program kill.  This is why I tend to have `main` be a thin wrapper around some other `func realMain(context.Context) error {` function that can defer.

Second thing is that when you implement Stop, be sure that it is capable of waiting for Start to actually return.  Otherwise make a waitgroup here.

test/stress/load-test/main.go
charithabandi: okay

PR #711, opened 2024-05-09
ci: backport with condition

Yaiba:
This add a condition on backport action, so it only runs when pr is merged and has a label like `backport-to-`

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -9,6 +9,7 @@ jobs:
   backport:
     name: Backport PR
     runs-on: ubuntu-latest
+    if: github.event.pull_request.merged && contains(github.event.pull_request.labels.*.name, 'backport-to-')
     steps:
       - name: Backport Action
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   backport:
     name: Backport PR
     runs-on: ubuntu-latest
+    if: github.event.pull_request.merged && contains(github.event.pull_request.labels.*.name, 'backport-to-')
```
jchappelow, line 12: So was the example yml file no good because "labeled" in the `type` causes the action to run when the label is added, not just merging a PR with the label?  That's screwy, but oh well.

jchappelow (approved):

.github/workflows/backport.yml
Yaiba: yes, here https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#label it explains it, I think we don't want that behavior. 
But in any case, we cannot avoid this workflow got triggerred every time there is a pr got merged

PR #712, opened 2024-05-09
skip CI on some file changes

Yaiba:
As our CI running time is now taking average 25 mins, we need to offload some unnecessary CI runs.

This add path filter to skip CI on certain file changes.

We use https://github.com/dorny/paths-filter, as the builtin [path filter](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks) won't fit in our use case

### DIFF
--- a/.github/filters.yaml
+++ b/.github/filters.yaml
@@ -1,0 +1,18 @@
+code:
+  - added|modified: 'go.mod'
+  - added|modified: '.github/filters.yaml'
+  - added|modified: '.github/workflows/ci.yaml'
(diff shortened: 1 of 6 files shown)


Yaiba:
I also modified CI so only jsonrpc,cli drivers are run for PR, http driver will be run after merged
 
I think it's good to go.

brennanjl (approved):
Thank you for doing this

PR #713, opened 2024-05-10
transactions: fix fee marshaling

jchappelow:
The Fee field of `core/types/transactions.TransactionBody` is a `*big.Int`.  By default this marshals to/from an integer (json number).  This should be a string to avoid issues with JavaScript representing all numbers as 64-bit floating point values, which can't store a large int64 (same reason why `json.Number` exists in Go).

This also changes a few field types:
- `TxQueryRequest.TxHash` is now `HexBytes`
- `TxQueryResponse.TxHash` is now `HexBytes`
- `BroadcastResponse.TxHash` is now `HexBytes`
- `VerifySignatureRequest.Msg` is now a `[]byte` (not `HexBytes`)

Finally, remove `core/types.NodeInfo`, which was unused (there's something similar that is used in core/types/admin).


Yaiba (approved):
LGTM

PR #719, opened 2024-05-10
made schema null marshalling consistent

brennanjl:
@KwilLuke commented on inconsistencies for nulls in JSON marshalled schemas. This PR removes `omitempty` tags to make the marshalling consistent.

We sort've used these randomly so I figured it's not a big deal to just ditch them


Yaiba (approved):
If this is a problem for client then I agree the `omitempty` should be removed

PR #723, opened 2024-05-10
Optimize TxQuery

charithabandi:
Performance showed that polling TxQuery frequently can lead to resource saturation due to computation of Hash() for all the transaction in the mempool. This PR avoids this by ABCI computing the TxHash when it first sees the transaction through CheckTx and stores it in its internal structure. This hash would be purged once the tx is included in the block. This avoids repeated computation of the Hash() in the TxQueries. 

Also reduces the repeated hash computations in checkTx making them a tad bit faster. 

One major change in the TXQuery behavior is the result doesn't include the actual transaction if its in the mempool, as we currently don't have a way to retrieve it without computing and checking the Hashes of all the mempool transactions. I am gonna raise a issue with cometBft to expose these APIs on the mempool. TxQuery would just return `{txHash, height=-1 and tx=nil}` for mempool transactions.

The block times have reduced from 60sec to 12 sec with this fix as shown below.
image

### DIFF
--- a/core/rpc/client/user/jsonrpc/methods.go
+++ b/core/rpc/client/user/jsonrpc/methods.go
@@ -212,10 +212,11 @@ func (cl *Client) TxQuery(ctx context.Context, txHash []byte) (*transactions.TcT
 	if err != nil {
 		return nil, err
 	}
+
 	return &transactions.TcTxQueryResponse{
 		Hash:     res.Hash,
 		Height:   res.Height,
-		Tx:       *res.Tx,
+		Tx:       res.Tx,
 		TxResult: *res.TxResult,
 	}, nil
 }
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -10,6 +10,7 @@ import (
 	"math/big"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/kwilteam/kwil-db/core/log"
 	"github.com/kwilteam/kwil-db/core/types"
@@ -48,6 +49,7 @@ func NewAbciApp(cfg *AbciConfig, snapshotter SnapshotModule, statesyncer StateSy
 		log: log,
 
 		validatorAddressToPubKey: make(map[string][]byte),
+		txCache:                  make(map[string]bool),
 	}
 
 	return app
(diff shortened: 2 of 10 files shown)


brennanjl:
Nice!

internal/abci/abci.go
```diff
@@ -96,6 +97,12 @@ type AbciApp struct {
 
 	// validatorAddressToPubKey is a map of validator addresses to their public keys
 	validatorAddressToPubKey map[string][]byte
+
+	// txCache stores hashes of all the transactions currently in the mempool.
+	// This is used to avoid recomputing the hash for all mempool transactions
+	// on every TxQuery request (to mitigate Potential DDOS attack vector).
```
jchappelow, line 103: or just a DoS vector as it turns out :laughing:

internal/abci/abci.go
```diff
@@ -197,6 +202,11 @@ func (a *AbciApp) CheckTx(ctx context.Context, incoming *abciTypes.RequestCheckT
 		return &abciTypes.ResponseCheckTx{Code: code.Uint32(), Log: err.Error()}, nil
 	}
 
+	// Cache the transaction hash
+	if newTx {
+		txHash := cmtTypes.Tx(incoming.Tx).Hash()
+		a.txCache[string(txHash)] = true
```
jchappelow, line 208: We have to worry about data races here.  `TxInMempool` can be called on demand via RPC, correct?

charithabandi: Oh shoot, totally forgot about those. Thanks for bringing that up

core/rpc/client/user/jsonrpc/methods.go
```diff
@@ -212,10 +212,16 @@ func (cl *Client) TxQuery(ctx context.Context, txHash []byte) (*transactions.TcT
...
+	var tx transactions.Transaction
+	if res.Tx != nil {
+		tx = *res.Tx
+	}
+
 	return &transactions.TcTxQueryResponse{
 		Hash:     res.Hash,
 		Height:   res.Height,
-		Tx:       *res.Tx,
+		Tx:       tx,
```
jchappelow, line 224: Looks like we should make this field a pointer.
Seems sloppy to return the zero value and all the nil and undefined fields of this type.

charithabandi: I agree with that

internal/abci/abci.go
jchappelow: steer away from `sync.Map` tho

jchappelow: oh nice you did it already with rwmutex :partying_face:

internal/abci/abci.go
```diff
@@ -197,6 +204,13 @@ func (a *AbciApp) CheckTx(ctx context.Context, incoming *abciTypes.RequestCheckT
 		return &abciTypes.ResponseCheckTx{Code: code.Uint32(), Log: err.Error()}, nil
 	}
 
+	// Cache the transaction hash
+	if newTx {
+		txHash := cmtTypes.Tx(incoming.Tx).Hash()
```
jchappelow, line 209: +1 for hashing before locking

jchappelow (approved):
Slick optimization.  :rocket:

PR #722, opened 2024-05-10
deps: resolve a number of security alerts

jchappelow:
There are a good number of dependabot security alerts we should resolve before beta period.

https://github.com/kwilteam/kwil-db/security/dependabot

Most are false positive, some are not.

PR #726, opened 2024-05-11
auto uuid scanning

brennanjl:
This PR makes it such that our UUID types are automatically scanned from the database. It also adds json marshalling for our UUID types, so that they are sent as strings over json.


core/types/uuid.go
```diff
@@ -62,6 +87,134 @@ func (u *UUID) Scan(src any) error {
...
+
+			uuid, ok := target.(*UUID)
+			if !ok {
+				if len(src) != 16 {
+					return fmt.Errorf("expected 16 bytes, got %d", len(src))
+				}
+
+				uuid := UUID{}
+				copy(uuid[:], src)
+				reflect.ValueOf(target).Elem().Set(reflect.ValueOf(uuid))
```
jchappelow, line 154: What's this doing?  I'm the case where `!ok` above, what's an example with concrete types where this does something sensible?
I'm not sure where `target` comes from or is used later

core/types/uuid.go
```diff
@@ -62,6 +87,134 @@ func (u *UUID) Scan(src any) error {
 	return errors.New("not a byte slice")
 }
 
+func (u *UUID) FormatSupported(format int16) bool {
+	return format == pgtype.TextFormatCode || format == pgtype.BinaryFormatCode
+}
+
+func (u *UUID) PreferredFormat() int16 {
+	return pgtype.TextFormatCode
```
jchappelow, line 95: The UUID type built in to pgype prefers binary.
Will you document some of the decisions in this code and/or like to reference implementations?  There's a lot of logic add with no obvious reasons, although I'm sure there are indeed good ones.

jchappelow:
> This PR makes it such that our UUID types are automatically scanned from the database

..for the UUID OID?  We already could scan/value with our UUID type for BYTEA, columns, so I gather we're now enabling it to work with the special UUID type from the standard postgres extension?

I'm all for controlling the implemention, but to document motivation, what are some of the considerations with making our own scanner vs. one of the existing like https://pkg.go.dev/github.com/jackc/pgx/v5@v5.5.5/pgtype#UUID or https://github.com/jackc/pgx/wiki/UUID-Support?

core/go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/google/go-cmp v0.5.9
 	github.com/google/uuid v1.5.0
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.18.0
+	github.com/jackc/pgx/v5 v5.5.2
```
jchappelow, line 13: Is it OK if we do upgrade pgx to 5.5.5?  We were planning to do so to resolve a security alert in https://github.com/kwilteam/kwil-db/pull/722.  I noticed a "downgrade" commit in this PR and wasn't sure if you hit a problem or were just aiming to keep the changes minimal.

brennanjl: I got some weird build error while upgrading, so I downgraded. Will upgrade though

core/types/uuid.go
brennanjl: I'll document this. When testing, it sometimes did not pass a UUID, but instead passed []byte. I believe this is because of how we pass uuids internally (such as in the vote store), where we convert them to []byte.

core/go.mod
jchappelow: No need, we upgraded in https://github.com/kwilteam/kwil-db/pull/722.
Just wanted to make sure there wasn't something problematic about 5.5.5

brennanjl:
While creating tests for this, I noticed that we actually had a bug for sending arrays via flags, as well as with RLP encoding. I have added into this PR fixes for these. These are breaking changes, but given the other breaking changes in this PR, I figured it was not a big deal. I also could not think of a way to properly support arrays without doing this.

The last outstanding thing on this PR is some ridiculous bug with writing UUIDs to postgres. For some reason, in one of the integration tests, I get a primary key violation for having duplicate primary keys, despite the fact that they certainly aren't.

Will finish this up tomorrow for the beta.

core/types/transactions/payloads.go
```diff
@@ -166,12 +169,175 @@ func (a *ActionExecution) Type() PayloadType {
...
+	}
+
+	// encodeScalar encodes a scalar value into a byte slice.
+	// It also returns the data type of the value.
+	encodeScalar := func(v any) ([]byte, *types.DataType, error) {
+		switch t := v.(type) {
+		case string:
+			return []byte(t), types.TextType, nil
+		case int, int16, int32, int64, int8, uint, uint16, uint32, uint64, uint8:
+			i64, err := conv.Int64(v)
```
jchappelow, line 260: Our `internal/conv` if possible, please.

core/types/transactions/payloads.go
```diff
@@ -166,12 +169,175 @@ func (a *ActionExecution) Type() PayloadType {
...
+		default:
+			return nil, fmt.Errorf("cannot decode type %s", t.Name)
+		}
+	}
+
+	// check if it is an array
+	if e.Type.IsArray {
+		decoded := make([]any, 0)
+		for _, elem := range e.Data {
+			dec, err := decodeScalar(elem, e.Type.toTypes())
```
jchappelow, line 223: Probably spare some conversions and allocations and just pass `e.Type.Name` to `decodeScalar` because that's all it needs.  Array context is already established here.

core/types/transactions/payloads.go
```diff
@@ -166,12 +169,175 @@ func (a *ActionExecution) Type() PayloadType {
...
+		case types.NullType.Name:
+			return nil, nil
+		default:
+			return nil, fmt.Errorf("cannot decode type %s", t.Name)
+		}
+	}
+
+	// check if it is an array
+	if e.Type.IsArray {
+		decoded := make([]any, 0)
```
jchappelow, line 221:
```suggestion
		decoded := make([]any, 0, len(e.Data))
```

jchappelow (commented):
Just looking at the client array encoding/decoding updates now.

core/types/transactions/payloads.go
brennanjl: We can't because this is in core.

jchappelow: Oh, well we definitely don't want to use stockton here.  We can move conv to core or copy/paste what's needed here.

brennanjl: I've got a way around it, will push up

brennanjl:
Closing in favor of https://github.com/kwilteam/kwil-db/pull/761. I have addressed your comments thus far there as well.

Issue #725, opened 2024-05-11
Throwing Away Engine Error


brennanjl:
I noticed that with the addition of JSON RPC, we throw away engine errors if we do not know what they are, and simply return internal error. IMO this is bad; a lot of basic errors get discarded this way, including exceptions raised by the schema itself.

Is there a reason we are throwing them away?

jchappelow:
I agree that's bad.  I didn't think we were throwing them away.

brennanjl:
Here's where they are being thrown away: https://github.com/kwilteam/kwil-db/blob/f2850cfc0b54e2970e142ceec0424e36915e4b5a/internal/services/jsonrpc/usersvc/service.go#L313

jchappelow:
Yeah.  Well, actually here: https://github.com/kwilteam/kwil-db/blob/f2850cfc0b54e2970e142ceec0424e36915e4b5a/internal/services/jsonrpc/usersvc/service.go#L310

PR #727, opened 2024-05-11
services/jsonrpc: do not throw away engine error details

jchappelow:
Resolves https://github.com/kwilteam/kwil-db/issues/725

The following rpc service methods calling into engine are now returning the full error when it is unidentified:
- `Query` (this is the most sensible place to return the full engine error)
- `ListDatabases` (but this error is more likely to be genuinely internal, since an empty slice is the result if none found rather than an error)
- `Call` (like `Query` it is likely that the "internal" error will be important to help the user resolve the issue with their call or the dataset definition itself).

Note that `Schema` also uses the `engineError` helper, but `engine.GetSchema` only returns `ErrDatasetNotFound` or `nil`, and there is no internal error that would have been discarded.

### DIFF
--- a/internal/services/jsonrpc/usersvc/service.go
+++ b/internal/services/jsonrpc/usersvc/service.go
@@ -307,7 +307,7 @@ func checkEngineError(err error) (jsonrpc.ErrorCode, string) {
 		return jsonrpc.ErrorEngineInvalidSchema, execution.ErrInvalidSchema.Error()
 	}
 
-	return jsonrpc.ErrorEngineInternal, "internal engine error"
+	return jsonrpc.ErrorEngineInternal, err.Error()
 }
 
 func engineError(err error) *jsonrpc.Error {


jchappelow:
**Reference** (no error):

```
$  ./kwil-cli database query -i x6f89a49f1bd5fd49b7f4e13fa9cdb76eb05cae29aa776cd88a9da0cf "SELECT * FROM users WHERE age > 25"
| age |         id          | username |                   wallet                   |
+-----+---------------------+----------+--------------------------------------------+
|  42 | 1061840862069797577 | me       | 0x5a9fd9d33B310eA8Fb192818C1016A353e734C41 |
```

Bad query ("agex" instead of "age"):

**Before** 

```
$  ./kwil-cli database query -i x6f89a49f1bd5fd49b7f4e13fa9cdb76eb05cae29aa776cd88a9da0cf "SELECT * FROM users WHERE agex > 25"
error querying database: jsonrpc.Error: code = -300, message = internal engine error, data = null
err code = -300, msg = internal engine error
```

**After**

```
 $  ./kwil-cli database query -i x6f89a49f1bd5fd49b7f4e13fa9cdb76eb05cae29aa776cd88a9da0cf "SELECT * FROM users WHERE agex > 25"
error querying database: jsonrpc.Error: code = -300, message = ERROR: column "agex" does not exist (SQLSTATE 42703), data = null
err code = -300, msg = ERROR: column "agex" does not exist (SQLSTATE 42703)
```

tldr: `msg = internal engine error` => `ERROR: column "agex" does not exist (SQLSTATE 42703)`

Which for an ad hoc query is exactly what I'd want in return.

Yaiba (approved):

brennanjl (approved):

PR #733, opened 2024-05-14
[WIP] ci: fix backport not run

Yaiba:
This fix the issue that `backport` workflow doesn't run.

I've tested in https://github.com/Yaiba/test-relase-based-workflow pr 17-28. It'll backport when put label when create pr, after create pr, after pr is merged. 

Only one issue is that it'll run on every pr even if not labelled, I just cannot make it not run , we can just ignore it.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -4,12 +4,15 @@ on:
   pull_request_target:
     types:
       - closed
+      - labeled # so labels will be updated if PR is labeled after creation
 
 jobs:
   backport:
     name: Backport PR
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jchappelow:
> Only one issue is that it'll run on every pr even if not labelled, I just cannot make it not run , we can just ignore it.

So it will make a backport PR for every PR?

Yaiba:
> every

It'll check if it needs to backport(if so, then create new PRs), and since most of PRs won't have those labels, this 'run' will report error, which we can ignore it.

Yaiba:
Actually, it'll send email reports it fails, which is annoying...

Yaiba:
close for now, will reopen once kwilteam/kgw#66 is resolved

PR #737, opened 2024-05-14
core/types/transactions: rlp nil proc returns

jchappelow:
Quick change.

PR #739, opened 2024-05-15
added parse debug

brennanjl:
@KwilLuke was running into a bug that was fixed by https://github.com/kwilteam/kwil-db/pull/737.

In the process of trying to figure it out, I added to our parse cli command to help with debugging. I also added to the schema cleaning, which should prevent this error from reaching postgres in the first place (since the error is still possible to hit with bad rlp encoding.)

### DIFF
--- a/cmd/kwil-cli/cmds/utils/parse.go
+++ b/cmd/kwil-cli/cmds/utils/parse.go
@@ -4,13 +4,16 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"reflect"
 
 	"github.com/kwilteam/kwil-db/cmd/common/display"
+	"github.com/kwilteam/kwil-db/internal/engine/ddl"
 	"github.com/kwilteam/kwil-db/parse"
 	"github.com/spf13/cobra"
 )
 
 func newParseCmd() *cobra.Command {
+	debug := false
 	cmd := &cobra.Command{
 		Use:   "parse <file_path>",
 		Short: "Parse a Kuneiform schema",
@@ -31,9 +34,54 @@ func newParseCmd() *cobra.Command {
 				return display.PrintErr(cmd, err)
 			}
 
+			if !debug {
+				res.Actions = nil
+				res.Procedures = nil
+			} else {
+				debugSchema := "debug_schema"
+				var tbls [][]string
+				for _, t := range res.Schema.Tables {
+					tbl, err := ddl.GenerateDDL(debugSchema, t)
+					if err != nil {
+						return display.PrintErr(cmd, err)
+					}
+
+					tbls = append(tbls, tbl)
+				}
+
+				var procs []string
+				for _, p := range res.Procedures {
+					proc, err := ddl.GenerateProcedure(p, debugSchema)
+					if err != nil {
+						return display.PrintErr(cmd, err)
+					}
+
+					procs = append(procs, proc)
+				}
+
+				var fprocs []string
+				for _, p := range res.Schema.ForeignProcedures {
+					proc, err := ddl.GenerateForeignProcedure(p)
+					if err != nil {
+						return display.PrintErr(cmd, err)
+					}
+
+					fprocs = append(fprocs, proc)
+				}
+
+				return display.PrintCmd(cmd, &dumpResult{
+					ParseResult:                *res,
+					GeneratedTables:            tbls,
+					GeneratedProcedures:        procs,
+					GeneratedForeignProcedures: fprocs,
+				})
+			}
+
 			return display.PrintCmd(cmd, &schemaDisplay{Result: res})
 		},
 	}
+
+	cmd.Flags().BoolVarP(&debug, "debug", "d", false, "debug mode (show generated code and ASTs)")
 
 	return cmd
 }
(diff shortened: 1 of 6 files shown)


cmd/kwil-cli/cmds/utils/parse.go
```diff
@@ -58,3 +106,34 @@ func (s *schemaDisplay) MarshalText() (text []byte, err error) {
...
+	k := reflect.TypeOf(v).Kind() // ptr or not?
+
+	if k != reflect.Ptr {
+		return fn(v)
+	}
+
+	// dereference pointer
+	v2 := reflect.ValueOf(v).Elem().Interface()
+	return marshalSafe(v2, fn)
+}
```
jchappelow, line 139: The "standard" way to do this is with a type declaration that hides the marshl/unmarshal methods

```go
func (d dumpResult) MarshalJSON() ([]byte, error) {
	type resAlias dumpResult // abandon methods
	return json.Marshal(resAlias(d))
}

func (d *dumpResult) MarshalText() (text []byte, err error) {
	type resAlias dumpResult
	return json.MarshalIndent((*resAlias)(d), "", "  ")
}
```

test

```go
package utils

import (
	"reflect"
	"testing"

	"github.com/kwilteam/kwil-db/core/types"
	"github.com/kwilteam/kwil-db/parse"
	parseTypes "github.com/kwilteam/kwil-db/parse/types"
)

func Test_dumpResult_MarshalJSON(t *testing.T) {
	tests := []struct {
		res     dumpResult
		want    []byte
		wantErr bool
	}{
		{
			dumpResult{
				ParseResult: parse.ParseResult{
					Schema: testSchema,
					// Errs: ,
					SchemaInfo: &parseTypes.SchemaInfo{},
				},
			},
			[]byte(`{"base_result":{"schema":{"name":"test_schema","owner":"75736572","extensions":[{"name":"auth","initialization":[{"name":"token","value":"abc123"}],"alias":"authentication"}],"tables":[{"name":"users","columns":[{"name":"id","type":{"name":"int","is_array":false},"attributes":[{"type":"primary_key","value":"true"}]}],"indexes":[{"name":"index_name","columns":["id","name"],"type":"btree"}],"foreign_keys":[{"child_keys":["child_id"],"parent_keys":["parent_id"],"parent_table":"parent_table","actions":[{"on":"delete","do":"cascade"}]}]}],"actions":[{"name":"get_user","annotations":null,"parameters":["user_id"],"public":true,"modifiers":null,"body":"SELECT * FROM users WHERE id = $user_id"}],"procedures":null,"foreign_calls":null},"schema_info":{"blocks":null}},"generated_tables":null,"generated_procedures":null,"generated_foreign_procedures":null}`),
			false,
		},
	}
	for _, tt := range tests {
		t.Run("", func(t *testing.T) {
			got, err := tt.res.MarshalJSON()
			if (err != nil) != tt.wantErr {
				t.Errorf("dumpResult.MarshalJSON() error = %v, wantErr %v", err, tt.wantErr)
				return
			}
			if !reflect.DeepEqual(got, tt.want) {
				t.Errorf("dumpResult.MarshalJSON() = %v, want %v", string(got), string(tt.want))
			}
		})
	}
}

var testSchema = &types.Schema{
	Owner: []byte("user"),
	Name:  "test_schema",
	Tables: []*types.Table{
		{
			Name: "users",
			Columns: []*types.Column{
				{
					Name: "id",
					Type: types.IntType,
					Attributes: []*types.Attribute{
						{
							Type:  "primary_key",
							Value: "true",
						},
					},
				},
			},
			ForeignKeys: []*types.ForeignKey{
				{
					ChildKeys:   []string{"child_id"},
					ParentKeys:  []string{"parent_id"},
					ParentTable: "parent_table",
					Actions: []*types.ForeignKeyAction{
						{
							On: "delete",
							Do: "cascade",
						},
					},
				},
			},
			Indexes: []*types.Index{
				{
					Name:    "index_name",
					Columns: []string{"id", "name"},
					Type:    "btree",
				},
			},
		},
	},
	Actions: []*types.Action{
		{
			Name:       "get_user",
			Parameters: []string{"user_id"},
			Public:     true,
			Body:       "SELECT * FROM users WHERE id = $user_id",
		},
	},
	Extensions: []*types.Extension{
		{
			Name: "auth",
			Initialization: []*types.ExtensionConfig{
				{
					Key:   "token",
					Value: "abc123",
				},
			},
			Alias: "authentication",
		},
	},
}

```

jchappelow: An alternative might be something like this:

```go
type Marshaller struct {
	any
}

func (d Marshaller) MarshalJSON() ([]byte, error) {
	return json.Marshal(d.any)
}

func (d *Marshaller) MarshalText() (text []byte, err error) {
	return json.MarshalIndent(d.any, "", "  ")
}
```

then 

```go
				return display.PrintCmd(cmd, &Marshaller{&dumpResult{
					ParseResult:                *res,
					GeneratedTables:            tbls,
					GeneratedProcedures:        procs,
					GeneratedForeignProcedures: fprocs,
				}})
```

parse/actions/actions.go
```diff
@@ -149,17 +149,17 @@ func convertStatement(stmt actparser.ActionStmt, schema *types.Schema, pgSchemaN
...
+	Name string `json:"name"`
 	// Public is true if the action is public.
-	Public bool
+	Public bool `json:"public"`
 	// IsView is true if the action is a view.
-	IsView bool
+	IsView bool `json:"is_view"`
 	// OwnerOnly is true if the action is owner-only.
-	OwnerOnly bool
+	OwnerOnly bool `json:"owner_only"`
```
jchappelow, line 158: IMO "view" and "owner" would be preferable, but it doesn't bother me too much with the verbosity to suggest bool

cmd/kwil-cli/cmds/utils/parse.go
```diff
@@ -58,3 +106,34 @@ func (s *schemaDisplay) MarshalText() (text []byte, err error) {
 		return json.MarshalIndent(s.Result, "", "  ")
 	}
 }
+
+type dumpResult struct {
```
Yaiba, line 110: I think #71 is conflict with this PR, and also might misinterpreted your intention.    Maybe fix #71 in this PR?

PR #741, opened 2024-05-15
kwil-cli: fix `utils parse` output

Yaiba:
This fix `kwil-cli utils parse` output, so the output can be used for `kwil-cli database deploy`.  

We have a test kf file:
`$ echo "database test;" > /tmp/test.kf`

Before:
```
$ .build/kwil-cli utils parse /tmp/test.kf                                                                                           
{
  "schema": {
    "name": "test",
    "owner": "",
    "extensions": null,
    "tables": null,
    "actions": null,
    "procedures": null,
    "foreign_calls": null
  }
}

$ .build/kwil-cli utils parse /tmp/test.kf > /tmp/test.json

$ .build/kwil-cli database deploy -p /tmp/test.json -t json
TxHash: 9a1f0200c2c7574f2e5f45c2546554ebb7a346d7b311ae0c3e1f4258f7953a3

$ .build/kwil-cli utils query-tx 9a1f0200c2c7574f2e5f45c2546554ebb7a346d7b311ae0c3e1f4258f7953a37
Transaction ID: 9a1f0200c2c7574f2e5f45c2546554ebb7a346d7b311ae0c3e1f4258f7953a37
Status: failed
Height: 78
Log: name cannot be empty
```

After:
```
$ .build/kwil-cli utils parse /tmp/test.kf 
{
  "name": "test",
  "owner": "",
  "extensions": null,
  "tables": null,
  "actions": null,
  "procedures": null,
  "foreign_calls": null
}

$ .build/kwil-cli utils parse /tmp/test.kf > /tmp/test.json

$ .build/kwil-cli database deploy -p /tmp/test.json -t json 
TxHash: 35ecace2fe1c49fb703ad339dcde3ab7279a62528e8d0367ad235c40be7344fd

$ .build/kwil-cli utils query-tx 35ecace2fe1c49fb703ad339dcde3ab7279a62528e8d0367ad235c40be7344fd
Transaction ID: 35ecace2fe1c49fb703ad339dcde3ab7279a62528e8d0367ad235c40be7344fd
Status: success
Height: 292
Log: success
```

### DIFF
--- a/cmd/kwil-cli/cmds/utils/parse.go
+++ b/cmd/kwil-cli/cmds/utils/parse.go
@@ -6,6 +6,7 @@ import (
 	"os"
 
 	"github.com/kwilteam/kwil-db/cmd/common/display"
+	"github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/parse"
 	"github.com/spf13/cobra"
 )
@@ -31,7 +32,11 @@ func newParseCmd() *cobra.Command {
 				return display.PrintErr(cmd, err)
 			}
 
-			return display.PrintCmd(cmd, &schemaDisplay{Result: res})
+			if res.Err() != nil {
+				return display.PrintErr(cmd, res.Err())
+			}
+
+			return display.PrintCmd(cmd, &schemaDisplay{Result: res.Schema})
 		},
 	}
 
@@ -39,9 +44,8 @@ func newParseCmd() *cobra.Command {
 }
 
 // schemaDisplay is a struct that will be used to display the schema.
-// It includes an error because the parser can return a schema even if there is an error.
 type schemaDisplay struct {
-	Result *parse.SchemaParseResult
+	Result *types.Schema
 }
 
 func (s *schemaDisplay) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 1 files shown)


Yaiba:
The problem is so easy to identify but I went totally wrong direction trying to debug server side...

jchappelow:
Oh, I would expect the issue is you should use `--output json`?  Or at least that MarshalJSON needs a similar fix?

Although it's also debatable if `ParseResult` should be returned and the caller should not expect the `Schema` only.  That is, a savy user would do `jq '.schema'` on the output to get what can be deployed?  Otherwise we should just change

```go
type schemaDisplay struct {
	Result *parse.ParseResult
}
```

to

```go
type schemaDisplay struct {
	Result *types.Schema
}
```

@brennanjl maybe you can clarify the requirements for this command.

Yaiba:
> Oh, I would expect the issue is you should use `--output json`? Or at least that MarshalJSON needs a similar fix?

ah, I forgot `--output` flags 🤦‍♂️

Yaiba:
> ```go
> type schemaDisplay struct {
> 	Result *types.Schema
> }
> ```

I think this is the right changes to make

Yaiba:
I think https://github.com/kwilteam/kwil-db/pull/741/commits/495bb79f0185ca0c3f1fc3390e0241e2b40be519 might be the right changes, if not i'll revert

BTW, I guess @brennanjl want json for `--output text`,  so it's useful for the purpose of `parse` cmd?
```
$ .build/kwil-cli utils parse /tmp/test.kf
{
  "name": "test",
  "owner": "",
  "extensions": null,
  "tables": null,
  "actions": null,
  "procedures": null,
  "foreign_calls": null
}

$ .build/kwil-cli utils parse /tmp/test.kf --output json      
{
  "result": {
    "name": "test",
    "owner": "",
    "extensions": null,
    "tables": null,
    "actions": null,
    "procedures": null,
    "foreign_calls": null
  },
  "error": ""
}
```

jchappelow:
Looks fine still, but you and @brennanjl have to decide about which way to go, I guess.

brennanjl (approved):

brennanjl (approved):

PR #743, opened 2024-05-15
Fix issues with TxQuery for unconfirmed transactions

charithabandi:
Minor fixes related to processing nil Transactions in the TxQuery response for unconfirmed transactions.

### DIFF
--- a/cmd/common/display/message.go
+++ b/cmd/common/display/message.go
@@ -107,11 +107,11 @@ func (r *RespTxQuery) MarshalJSON() ([]byte, error) {
 		TxResult: r.Msg.TxResult,
 	}
 	// Always try to serialize to verify hash, but only show raw if requested.
-	raw, err := r.Msg.Tx.MarshalBinary()
-	if err != nil {
-		out.Warn = "ERROR encoding transaction: " + err.Error()
-	} else {
-		if r.WithRaw {
+	if r.Msg.Tx != nil {
+		raw, err := r.Msg.Tx.MarshalBinary()
+		if err != nil {
+			out.Warn = "ERROR encoding transaction: " + err.Error()
+		} else if r.WithRaw {
 			out.Raw = hex.EncodeToString(raw)
 			hash := sha256.Sum256(raw)
 			if !bytes.Equal(hash[:], r.Msg.Hash) {
@@ -145,6 +145,10 @@ Log: %s`,
 	)
 
 	// Always try to serialize to verify hash, but only show raw if requested.
+	if r.Msg.Tx == nil {
+		return []byte(msg), nil
+	}
+
 	raw, err := r.Msg.Tx.MarshalBinary()
 	if err != nil {
 		msg += "\nERROR encoding transaction: " + err.Error()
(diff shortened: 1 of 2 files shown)


cmd/common/display/message.go
```diff
@@ -107,16 +107,18 @@ func (r *RespTxQuery) MarshalJSON() ([]byte, error) {
...
-			hash := sha256.Sum256(raw)
-			if !bytes.Equal(hash[:], r.Msg.Hash) {
-				out.Warn = fmt.Sprintf("HASH MISMATCH: requested %x; received %x",
-					r.Msg.Hash, hash)
+	if r.Msg.Tx != nil {
+		raw, err := r.Msg.Tx.MarshalBinary()
+		if err != nil {
+			out.Warn = "ERROR encoding transaction: " + err.Error()
+		} else {
+			if r.WithRaw {
```
jchappelow, line 115: So I guess I'm the culprit here, but let's pls revise to `} else if r.WithRaw {`
The indent is getting kind deep now.

cmd/common/display/message.go
```diff
@@ -107,16 +107,18 @@ func (r *RespTxQuery) MarshalJSON() ([]byte, error) {
 		TxResult: r.Msg.TxResult,
 	}
 	// Always try to serialize to verify hash, but only show raw if requested.
-	raw, err := r.Msg.Tx.MarshalBinary()
```
jchappelow, line 110: Pretty wild that this doesn't error with `r.Msg.Tx == nil`, but it's just doing `serialize.Encode(nil)` which RLP is just fine with  :facepalm:

internal/services/grpc/txsvc/v1/tx_query.go
```diff
@@ -28,14 +28,19 @@ func (s *Service) TxQuery(ctx context.Context, req *txpb.TxQueryRequest) (*txpb.
 		return nil, status.Error(codes.Unknown, "failed to query transaction")
 	}
 
-	originalTx := &transactions.Transaction{}
-	if err := originalTx.UnmarshalBinary(cmtResult.Tx); err != nil {
-		logger.Error("failed to deserialize transaction", zap.Error(err))
-		return nil, status.Error(codes.Internal, "failed to deserialize transaction")
+	// cmtResult.Tx can be nil if the transaction is not found
+	var tx *txpb.Transaction
+	var originalTx *transactions.Transaction
```
jchappelow, line 33: Can omit this decl.  In the conditional block, can declare with `:=` since it's scope is limited.  Only `tx` needed out here.

core/rpc/client/user/http/convert.go
```diff
@@ -14,6 +14,10 @@ import (
 
 // convertTx converts a transaction to a httpTx.TxTransaction
 func convertTx(tx *transactions.Transaction) *httpTx.TxTransaction {
+	if tx == nil {
+		return nil
+	}
```
jchappelow, line 19: I have no problem with this, for safety, but is there somewhere this is expected presently?  Can leave regardlesss, but I don't see it being used to handle the nil unconfirmed thing back from server.

jchappelow (changes requested):
Thx.  Minor nits, but looks good.

jchappelow (approved):

PR #744, opened 2024-05-15
pg: always release conn on commit/rollback error

jchappelow:
This fixes a shutdown deadlock and eventual DB reader pool exhaustion while running.

@brennanjl This resolves the hang you observed.  I accidentally reproduced it while testing the timeouts in https://github.com/kwilteam/kwil-db/pull/736

I hate to say it, but this might need to be on `release-0.7`


Yaiba (commented):
I'm not familiar with this pkg, but lgtm

PR #745, opened 2024-05-15
chore: update debug dockerfile

Yaiba:
This makes the kwild debug image work


jchappelow (approved):
LGTM.  Although for the debug variant I guess the ubuntu weight wouldn't matter.  Still, best to be consistent with the production image.

PR #747, opened 2024-05-16
[v0.7.6] kwild,rpc: db and rpc timeout settings

jchappelow:
This backports https://github.com/kwilteam/kwil-db/pull/736, just without the JSON-RPC pieces.

### DIFF
--- a/cmd/kwild/config/test_data/config.toml
+++ b/cmd/kwild/config/test_data/config.toml
@@ -40,6 +40,12 @@ admin_listen_addr="localhost:50151"
 # TCP address for the KWILD App's HTTP server to listen on
 http_listen_addr = "localhost:8080"
 
+# Timeout on requests on the user RPC servers
+rpc_timeout = "45s"
+
+# Timeout on database reads initiated by the user RPC service
+db_read_timeout = "5s"
+
 # List of Extension endpoints to be enabled ex: ["localhost:50052", "169.198.102.34:50053"]
 extension_endpoints = ["localhost:50052", "localhost:50053", "localhost:50054"]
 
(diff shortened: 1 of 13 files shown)


cmd/kwild/config/test_data/config.toml
```diff
@@ -40,6 +40,12 @@ admin_listen_addr="localhost:50151"
 # TCP address for the KWILD App's HTTP server to listen on
 http_listen_addr = "localhost:8080"
 
+# Timeout on requests on the user RPC servers
+rpc_timeout = "45s"
+
+# Timeout on database reads initiated by the user RPC service
+db_read_timeout = "5s"
```
jchappelow, line 47: To reiterate, `rpc_timeout` applies to to just the DB exec in both `Call` and `Query`, just after `BeginReadTx`.

The `db_read_timeout` on the other hand is for the RPC server.  It sets the deadline for the completion of an RPC request once it is initiated.  This applies to all RPC methods.  This also covers the entire process of request to response, including reading the request itself (which may be large if broadcasting a tx that inserts a lot of data), to the actual handling of the request, and finally sending the bytes of the response back to the client.  These are all points where a malicious or discourteous client may utilize excessive resources by holding open connections or doing very expensive operations outside of paid transactions.

Regarding the gRPC approach to RPC timeouts: https://github.com/grpc/grpc-go/issues/5059

Yaiba: I think you mean `db_read_timeout applies to to just the DB exe` and  `rpc_timeout on the other hand is for the RPC server.` ?

Yaiba: For gRPC, I was thinking [gRPC deadlines](https://grpc.io/docs/guides/deadlines/), but after review it, quote:

> "Please note that the server application is responsible for stopping any activity it has spawned to service the request. If your application is running a long-running process you should periodically check if the request that initiated it has been cancelled and if so, stop the processing."

I think the interceptor is the right way to do it.

Yaiba (approved):
LGTM

PR #750, opened 2024-05-16
feature: support basic cors for jsonrpc server

### LINKED ISSUE #749: Bug: CORS does not apply to JSONRPC
Adding issue so we can track.

Full conversation and context here: https://kwilteam.slack.com/archives/C05RZNFKNJD/p1715816980708999

Yaiba:
This resolves #749

This adds a simple basic CORS returns cors headers for all requests

### DIFF
--- a/internal/services/jsonrpc/server.go
+++ b/internal/services/jsonrpc/server.go
@@ -43,7 +43,7 @@ type serverConfig struct {
 	pass       string
 	tlsConfig  *tls.Config
 	timeout    time.Duration
-	enabelCORS bool
+	enableCORS bool
 }
 
 type Opt func(*serverConfig)
@@ -75,7 +75,7 @@ func WithTimeout(timeout time.Duration) Opt {
 
 func WithCORS() Opt {
 	return func(c *serverConfig) {
-		c.enabelCORS = true
+		c.enableCORS = true
 	}
 }
 
@@ -174,7 +174,7 @@ func NewServer(addr string, log log.Logger, opts ...Opt) (*Server, error) {
 	// contexts: https://github.com/golang/go/issues/59602
 	// So, we add a timeout to the Request's context.
 	h = jsonRPCTimeoutHandler(h, cfg.timeout)
-	if cfg.enabelCORS {
+	if cfg.enableCORS {
 		h = corsHandler(h)
 	}
 	h = recoverer(h, log)
@@ -207,7 +207,7 @@ func corsHandler(h http.Handler) http.Handler {
 		w.Header().Set("Access-Control-Allow-Credentials", "true")
 
 		// Preflight request
-		if r.Method == "OPTIONS" {
+		if r.Method == http.MethodOptions {
 			return
 		}
 
--- a/internal/services/jsonrpc/server.go
+++ b/internal/services/jsonrpc/server.go
@@ -40,9 +40,10 @@ type Server struct {
 }
 
 type serverConfig struct {
-	pass      string
-	tlsConfig *tls.Config
-	timeout   time.Duration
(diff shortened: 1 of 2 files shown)


internal/services/jsonrpc/server.go
```diff
@@ -182,6 +192,30 @@ func jsonRPCTimeoutHandler(h http.Handler, timeout time.Duration) http.Handler {
 	return http.TimeoutHandler(h, timeout, string(respMsg))
 }
 
+// corsHandler adds CORS headers to the response. We don't need sophisticated
+// cors handling here (not really kwild's concern, there should be other services
+// like LBs or KGW do that), so we just allow them.
+// NOTE: if this server is served behind KGW, those headers will be stripped.
+func corsHandler(h http.Handler) http.Handler {
+	allowMethods := "GET, POST, OPTIONS"
```
Yaiba, line 200: Since we might support REST, I have 'GET' here. maybe we should add this only needed.

internal/services/jsonrpc/server.go
```diff
@@ -182,6 +192,30 @@ func jsonRPCTimeoutHandler(h http.Handler, timeout time.Duration) http.Handler {
...
+	allowHeaders := "Accept, Content-Type, Content-Length, Accept-Encoding, Authorization, ResponseType, Range"
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Access-Control-Allow-Origin", r.Header.Get("Origin"))
+		w.Header().Set("Access-Control-Allow-Methods", allowMethods)
+		w.Header().Set("Access-Control-Allow-Headers", allowHeaders)
+		w.Header().Set("Access-Control-Allow-Credentials", "true")
+
+		// Preflight request
+		if r.Method == "OPTIONS" {
```
jchappelow, line 210: Best to use `http.MethodOptions` here.

internal/services/jsonrpc/server.go
Yaiba: This also remind me that we might want to return error if handlerV1 get a non-POST request?

internal/services/jsonrpc/server.go
```diff
@@ -40,9 +40,10 @@ type Server struct {
 }
 
 type serverConfig struct {
-	pass      string
-	tlsConfig *tls.Config
-	timeout   time.Duration
+	pass       string
+	tlsConfig  *tls.Config
+	timeout    time.Duration
+	enabelCORS bool
```
jchappelow, line 46: `enabelCORS` => `enableCORS`

jchappelow (approved):
LGTM.  Just two minor tweaks.  If it works for you can @KwilLuke , then it's good to go.

Yaiba:
We have https://github.com/kwilteam/kwil-db/blob/0ee4c8a2e07a26f04704c6a084fa735186966e56/internal/services/jsonrpc/server.go#L275 (and few other places in same function) just return vanilla error, should we return jsonrpc.Response?

KwilLuke:
Just tested - works well when making browser requests 👍

jchappelow:
> @jchappelow We have
> 
> https://github.com/kwilteam/kwil-db/blob/0ee4c8a2e07a26f04704c6a084fa735186966e56/internal/services/jsonrpc/server.go#L275
> (and few other places in same function) just return vanilla error, should we return jsonrpc.Response?

That reminds me I have to work on that authSHA stuff.  It's half-baked.  But regarding the jsonrpc.Response, I was thinking to use a basic response until I can decode the jsonrpc.Request.  Will revisit.

jchappelow (approved):

PR #753, opened 2024-05-16
httpConvert and logging fixes

charithabandi:
- Handles `nil` transactions while decoding TxQuery responses
- Adjusting the logging in snapshots module
- Record correct spends in txApp (minor)


internal/txapp/txapp.go
```diff
@@ -942,7 +942,7 @@ func (r *TxApp) checkAndSpend(ctx TxContext, tx *transactions.Transaction, price
 				return nil, transactions.CodeUnknownError, err2
 			}
 
-			return nil, transactions.CodeInsufficientBalance, fmt.Errorf("transaction tries to spend %s tokens, but account only has %s tokens", amt.String(), tx.Body.Fee.String())
+			return account.Balance, transactions.CodeInsufficientBalance, fmt.Errorf("transaction tries to spend %s tokens, but account only has %s tokens", amt.String(), tx.Body.Fee.String())
```
jchappelow, line 945: Ah, for abci's `ExecTxResult`.  Yah, seems alright given `FinalizeBlock` is written with this flow in mind.

I'm assuming this altered result does not affect apphash in any way, just logging and meta data obtainable from cometbft via e.g. their RPC server?  Regardless, it's alright for 0.8.  Just curious.

jchappelow (approved):

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -159,12 +159,23 @@ hostname = "{{ .AppCfg.Hostname }}"
...
 [app.extensions]
 
+# Oracle extensions can be enabled by adding the following configuration
+# Each oracle extension configuration is defined under a subsection identified by the 
+# oracle extension name [app.extensions.<oracle_extension-name>]
+# The configuration options for each oracle extension are defined as key-value pairs under the subsection.
+# Only string values are supported for these configuration options.
+# For example, to enable the Ethereum listener extension, the configuration would look like:
+# [app.extensions.eth_listener]
+# rpc_provider = "https://mainnet.infura.io/v3/YOUR_INFURA_API_KEY"
```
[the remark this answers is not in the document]
charithabandi: Yeah, it's still based on whether the config is added. Need to think a bit if we explicilty needs enabled flag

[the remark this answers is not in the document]
jchappelow: It's fine however it is, I just couldn't recall if that was part of the system or not.

internal/txapp/txapp.go
charithabandi: umm, yeah. should be okay for 0.8, but this would be a consensus change.

jchappelow: It does affect app hash?  I had an idea that not everything does, but wasn't sure what.   e.g. Code obviously would, but "logs" and "events" etc. I wasn't sure about.  Let's assume this fix can only be on 0.8.  :)

jchappelow: Oh wait, we had talked about this with @brennanjl before.

Although we have 

```go
		abciRes.Code = txRes.ResponseCode.Uint32()
		abciRes.GasUsed = txRes.Spend
```

we thought that we should keep cometbft blind to gas.  There were threads about this:

- https://github.com/cosmos/cosmos-sdk/discussions/18894
- ... (I know there was a cometbft one, maybe and adr, looking...)

What's your take on these @charithabandi ?

jchappelow: Ah, this is old and in tendermint context: https://github.com/cometbft/cometbft/blob/main/docs/references/rfc/tendermint-core/rfc-011-delete-gas.md
Perhaps it's not relevant in cometbft anymore, but the more recent cosmos discussion suggests it's still the thinking. No?

jchappelow: Ok, disregard all that.  Until and if cometbft gets rid of gas recording as part of the recorded transaction result, I think there's a strong case to keep it.  Namely, we don't have our own spend journal that can be used to report how much a transaction spent or the change of balance over time.

charithabandi: Yeah, makes sense

PR #756, opened 2024-05-17
kwil-admin: fix setup testnet with admin address

jchappelow:
This fixes the `kwil-admin setup testnet` command, which was broken when the format for `admin_listen_addr` changed from a full URL to either host:port or a file path (for unix) when JSON-RPC was introduced.

The integration tests were not broken because they no not use the `UniquePorts` option of `nodecfg` as each node is run in a container with no need to avoid port conflicts.

This also changes the port increment for the user rpc server to -1 instead of +1 to avoid collision with the admin server at 8485 (although the default is unix socket for admin anyway).  I initially changed the admin port default to 8585 to avoid this entirely, but I liked 8485 better.  Its a sorta soft default anyway since the real default is a unix socket path of `/tmp/kwild.socket`


cmd/kwild/server/server.go
```diff
@@ -115,18 +116,20 @@ func New(ctx context.Context, cfg *config.KwildConfig, genesisCfg *config.Genesi
 
 func (s *Server) Start(ctx context.Context) error {
 	defer func() {
+		s.log.Info("Closing server resources...")
 		err := s.closers.closeAll()
 		if err != nil {
 			s.log.Error("failed to close resource:", zap.Error(err))
 		}
+		s.log.Info("Server is now shut down.")
```
jchappelow, line 124: I didn't mean to sneak this in, but it would be helpful in cases like today with fractal.

charithabandi:
Thanks for doing this.👍🏻

charithabandi (approved):
LGTM!

PR #757, opened 2024-05-20
kwild: streamline db is-initialized check

jchappelow:
Genesis data restoration has a check for an already-initialized database on start up.  This was creating a `pg.DB` instance, which is the heavyweight type that does much more than just provide a DB connection.  We don't want it to create the replication slot, triggers, or any other stuff required by kwild to run.

This change instead uses a `pg.Pool`, which is a barebones DB connection  so we can do the `GetValidators`  query and nothing else.

This also rearranges the conditions in the early return from `restoreDB` so that it is more likely to short-circuit and not bother doing the actual `isDbInitialized` check.  For instance, if state sync were disabled or `DataAppHash` were empty, the `isDbInitialized` check doesn't need to be done.

The effect is faster startup, with less confusing logs that showed seemingly redundant DB connection setup.

### DIFF
--- a/cmd/kwild/server/build.go
+++ b/cmd/kwild/server/build.go
@@ -412,7 +412,7 @@ func buildDB(d *coreDependencies, closer *closeFuncs) *pg.DB {
 	// If yes, restore the database from the snapshot
 	restoreDB(d)
 
-	db, err := d.dbOpener(d.ctx, d.cfg.AppCfg.DBName, 11)
+	db, err := d.dbOpener(d.ctx, d.cfg.AppCfg.DBName, 24)
 	if err != nil {
 		failBuild(err, "kwild database open failed")
 	}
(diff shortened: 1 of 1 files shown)


cmd/kwild/server/build.go
```diff
@@ -428,10 +428,10 @@ func buildDB(d *coreDependencies, closer *closeFuncs) *pg.DB {
 // DB restoration from snapshot is skipped in the following scenarios:
 //   - If the DB is already initialized (i.e this is not a new node)
 //   - If the genesis apphash is not specified
-//   - If statesync is enabled. Statesync will take care of rapildly syncing the database
+//   - If statesync is enabled. Statesync will take care of syncing the database
 //     to the network state using statesync snapshots.
 func restoreDB(d *coreDependencies) {
-	if isDbInitialized(d) || d.genesisCfg.DataAppHash == nil || d.cfg.ChainCfg.StateSync.Enable {
+	if d.cfg.ChainCfg.StateSync.Enable || len(d.genesisCfg.DataAppHash) == 0 || isDbInitialized(d) {
```
jchappelow, line 434: @charithabandi This is the short-circuit behavior I mentioned in a recent review.  The boolean operations  are evaluated left-to-right, but not necessarily all of them.  With OR, it stops when it hits the first `true`.  With AND, it stops when it hits the first `false`.  Sorry if you were already aware; just wanted to be explicit.

cmd/kwild/server/build.go
```diff
@@ -412,7 +412,7 @@ func buildDB(d *coreDependencies, closer *closeFuncs) *pg.DB {
 	// If yes, restore the database from the snapshot
 	restoreDB(d)
 
-	db, err := d.dbOpener(d.ctx, d.cfg.AppCfg.DBName, 11)
+	db, err := d.dbOpener(d.ctx, d.cfg.AppCfg.DBName, 24)
```
jchappelow, line 415: Eh, I was actually just fooling with this, but I think this his a righteous and safe change for the main DB to have a larger reader pool.  The default `max_connections` setting in every stock postgresql.conf I have encountered for 8 years has been 100.   Further `kwild` code requires this `postgres` setting to be at least 50:

https://github.com/kwilteam/kwil-db/blob/8176ce17d408ddbf611ee8595e7593b9bd7b146b/internal/sql/pg/system.go#L116

So we could possibly go even higher, but let's be cautious.

charithabandi (approved):

cmd/kwild/server/build.go
charithabandi: lol, Can't believe I wrote it this way

PR #758, opened 2024-05-20
jsonrpc: request logging, cleanup, join BroadcastError

jchappelow:
This adds request logging with elapsed times to the JSON-RPC server.

This also cleans up quite a bit of ugly logging in various packages, including `internal/sql/pg`.

This also returns the decoded `BroadcastError` (joined) to the user of `core/client.Client`.  This provides the details and message for the broadcast failure.  We were already decoding this typed error, but only for the integer error _code_, which we translated into a concrete `error` type for Go clients.

Finally, this removes two unused gRPC-related files from `core/rpc/transport`


Yaiba (approved):
LGTM

PR #759, opened 2024-05-20
log,kwild: allow increase log thresh on child loggers

jchappelow:
This adds the ability to raise the log level threshold for: RPC, cometbft, and DB (pg).  This feature is helpful if you want to do things like:
- run kwild with *debug* level logging, but don't want one of these systems spamming so you can set those to *info*
- run kwild with default *info* level logging, but only want *warnings* (or worse) from cometbft, etc.

I would like the ability to set all subsystem loggers independently, including lowering the threshold, but zap only permits increasing the level for child loggers.  This restriction seems to be a consequence of the zap core architecture.

### DIFF
--- a/cmd/kwil-admin/nodecfg/toml.go
+++ b/cmd/kwil-admin/nodecfg/toml.go
@@ -77,6 +77,15 @@ const defaultConfigTemplate = `
 # Output level for logging, default is "info". Other options are "debug", "error", "warn", "trace"
 level = "{{ .Logging.Level }}"
 
+# RPC systems' logging level. Must be higher than log.level.
+rpc_level = "{{ .Logging.RPCLevel }}"
+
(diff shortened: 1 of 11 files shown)


charithabandi:
https://docs.cometbft.com/v0.37/core/running-in-production#logging 


Cometbft already supports module level log specifications, Should we use that?

jchappelow:
> https://docs.cometbft.com/v0.37/core/running-in-production#logging
> 
> Cometbft already supports module level log specifications, Should we use that?

I don't have any particular cometbft module log level variations in mind.

But if you mean switch to a logger that is more usable, I'm all for that.  cometbft uses https://github.com/go-kit/log I believe.  The [stdlib `slog`](https://go.dev/blog/slog) is also likely a good choice.  We'd have to do a survey.

charithabandi (approved):

PR #761, opened 2024-05-22
Type encoding

brennanjl:
This is https://github.com/kwilteam/kwil-db/pull/726, but renamed to encapsulate further changes.

Will write a description later.


core/types/uuid.go
```diff
@@ -15,9 +16,9 @@ type UUID [16]byte
 
 // NewUUIDV5 generates a uuidv5 from a byte slice.
 // This is used to deterministically generate uuids.
-func NewUUIDV5(from []byte) UUID {
-	u := uuid.NewSHA1(namespace, from)
-	return UUID(u)
+func NewUUIDV5(from []byte) *UUID {
+	u := UUID(uuid.NewSHA1(namespace, from))
+	return &u
```
jchappelow, line 21: Why the change to pointer?  a 16 byte array is preferable to copy by value in general, but when it comes to the Valuer, that should really be a value receiver.

core/types/uuid.go
```diff
@@ -42,22 +43,48 @@ func (u UUID) String() string {
...
 	return u[:]
 }
 
-var _ driver.Valuer = UUID{}
+// Over json, we want to send uuids as strings
+func (u UUID) MarshalJSON() ([]byte, error) {
+	return json.Marshal(u.String())
+}
+
+func (u UUID) UnmarshalJSON(b []byte) error {
```
jchappelow, line 59: Unmarshal/Scan type methods should almost always be pointer receiver on the other hand.

core/types/uuid.go
brennanjl: Ah ok. I'll keep passing uuids by pointer, but make the valuer concrete

core/types/uuid.go
```diff
@@ -42,22 +43,48 @@ func (u UUID) String() string {
 	return uuid.UUID(u).String()
 }
 
-func (u UUID) Value() (driver.Value, error) {
-	return u[:], nil // []byte works for sql
+func (u *UUID) Value() (driver.Value, error) {
+	return u.String(), nil
```
jchappelow, line 47: My understanding of this change is that this type no longer works with bytea, only a uuid or text in pg.  When used with a uuid column type, this won't actually be stored as a string in the DB storage, right?  This is just how we communicate it?
Even so, I'm a little uncertain why this has to be text instead of 16 bytes.

core/types/uuid.go
jchappelow:
```go

func Test_UUIDJSON(t *testing.T) {
	seed := []byte("test")

	uuid := types.NewUUIDV5(seed)

	b, err := json.Marshal(uuid)
	if err != nil {
		t.Fatal(err)
	}

	var uuid2 types.UUID
	err = json.Unmarshal(b, uuid2)
	if err != nil {
		t.Fatal(err)
	}
}
```

```
=== RUN   Test_UUIDJSON
    /home/jon/kwil/git/kwil-db/core/types/uuid_test.go:33: json: Unmarshal(non-pointer types.UUID)
--- FAIL: Test_UUIDJSON (0.00s)
FAIL
coverage: 0.8% of statements
FAIL	github.com/kwilteam/kwil-db/core/types	0.012s
```

core/types/uuid.go
brennanjl: For some reason when I was testing, I found that an error was fixed by making it string. Now that I am trying again, it seems like it works either way, so I will make it return `[]byte` again. I've tried so many different permutations of different things over the last 2 days I've forgotten why.

core/types/uuid.go
jchappelow:
```go

func Test_UUIDJSON(t *testing.T) {
	seed := []byte("test")

	uuid := types.NewUUIDV5(seed)

	b, err := json.Marshal(uuid)
	if err != nil {
		t.Fatal(err)
	}

	t.Log(uuid)

	var uuid2 types.UUID
	err = json.Unmarshal(b, uuid2) // call of Unmarshal passes non-pointer as second argumentunmarshaldefault
	if err != nil {
		t.Error(err)
	}

	var uuid3 types.UUID
	err = json.Unmarshal(b, &uuid3)
	if err != nil {
		t.Fatal(err)
	}
	t.Log(uuid3) // 00000000-0000-0000-0000-000000000000
}
```

```
=== RUN   Test_UUIDJSON
    /home/jon/kwil/git/kwil-db/core/types/uuid_test.go:30: 24aa70cf-0e18-57c9-b449-da8c9db37821
    /home/jon/kwil/git/kwil-db/core/types/uuid_test.go:35: json: Unmarshal(non-pointer types.UUID)
    /home/jon/kwil/git/kwil-db/core/types/uuid_test.go:43: 00000000-0000-0000-0000-000000000000
--- FAIL: Test_UUIDJSON (0.00s)
FAIL
coverage: 2.8% of statements
FAIL	github.com/kwilteam/kwil-db/core/types	0.013s
```

core/types/uuid.go
```diff
@@ -43,21 +44,47 @@ func (u UUID) String() string {
...
 
-func (u UUID) Bytes() []byte {
+func (u *UUID) Bytes() []byte {
 	return u[:]
 }
 
-var _ driver.Valuer = UUID{}
+// Over json, we want to send uuids as strings
+func (u *UUID) MarshalJSON() ([]byte, error) {
+	return json.Marshal(u.String())
```
jchappelow, line 56: Some trouble here now.  You get different results if you marshal a value vs a pointer.   Here's where a value receiver for the marshal method is normally the way:

```go

func Test_UUIDJSON(t *testing.T) {
	seed := []byte("test")

	uuidPtr := types.NewUUIDV5(seed)
	uuidVal := *uuidPtr

	b, err := json.Marshal(uuidPtr)
	if err != nil {
		t.Fatal(err)
	}

	t.Log(string(b)) // "24aa70cf-0e18-57c9-b449-da8c9db37821"

	var uuid2 types.UUID
	err = json.Unmarshal(b, &uuid2)
	if err != nil {
		t.Fatal(err)
	}
	t.Log(uuid2) // 24aa70cf-0e18-57c9-b449-da8c9db37821

	b, err = json.Marshal(uuidVal)
	if err != nil {
		t.Fatal(err)
	}

	t.Log(string(b)) // [36,170,112,207,14,24,87,201,180,73,218,140,157,179,120,33] (oops, didn't use MarshalJSON method!)

	var uuid3 types.UUID
	err = json.Unmarshal(b, &uuid3)
	if err != nil {
		t.Fatal(err)
	}
	t.Log(uuid3) // json: cannot unmarshal array into Go value of type string
}
```

core/types/uuid.go
brennanjl: I'm not sure why I said this. I am trying again, and confirmed using Bytes does not work. I get:

```
invalid input syntax for type uuid: "\xc430dbfc28a9508aaee9e6a2e1d81bb1" (SQLSTATE 22P02)
```

jchappelow: Yeah, dang: https://github.com/jackc/pgx/blob/24c0a5e8ff740778bbed9810629cb83c9ea18bf9/pgtype/uuid.go#L91-L96
I guess that's just how `uuid_ossp` demands it.

Are we using any of the extension functions? 
 https://www.postgresql.org/docs/current/uuid-ossp.html#UUID-OSSP-FUNCTIONS-SECT
If not I'm wondering what this extension's `uuid` type on the backend gets us over bytea.  Strong typing in procedures i suppose doesn't work unless?

brennanjl: >  Are we using any of the extension functions

Yeah, `uuid_generate_v5` is exposed to users. It's quite convenient, because it allows users to not have to guess on having available primary keys any more: https://github.com/kwilteam/kwil-db/blob/3a67ad32c8df6c8ba5d3c279992dfae43293d470/test/acceptance/test-data/users.kf#L88

jchappelow: oh, yeah, nvm. This is hash based and txid is obv deterministic.

brennanjl: Yeah, I'm just not quite sure how I could achieve that. This felt easier, and overall wasn't concerned with collisions.

jchappelow: I'm just lost on why we need UUID everywhere.  Having a primary key column defined as `i SERIAL PRIMARY KEY` and letting it autoincrement is the easiest solution of all.
UUID caries risk, from performance to guarantees about the extension that provides it's functionality.
I don't much mind if schema authors really want it, but for internal tables it doesn't make a lot of sense to me.

brennanjl: Auto-increment is not guaranteed to be deterministic. It's only guaranteed that it is greater than a previous value. Randal pointed this out to us, as he had seen people hit issues with the same assumption.

jchappelow: I think this review comment can be resolved, but linking to a new discussion about the SERIAL issue should we decide to revisit: https://github.com/kwilteam/kwil-db/discussions/766

cmd/kwil-cli/cmds/database/call.go
```diff
@@ -104,51 +105,183 @@ func callCmd() *cobra.Command {
...
 	return nil, fmt.Errorf("procedure/action not found")
 }
 
-func buildActionInputs(a *types.Action, inputs []map[string]any) [][]any {
+func decodeManyB64(inputs []string) ([][]byte, bool) {
+	b64Arr := [][]byte{}
+	b64Ok := true
+	for _, s := range inputs {
+		// in the CLI, if data has suffix ;b64, it is base64 encoded
+		if strings.HasSuffix(s, ";b64") {
```
Yaiba, line 134: surprisingly shell doesn't treat what's following ; as a separate cmd on you machine? I just tried, it's treated as a separated cmd , `zsh: command not found: 6b4`

brennanjl: Yeah I have to wrap it in quotes. I'm open to changing it to something else (like SQL's typecast) so that it doesn't need to be wrapped.

Yaiba: Yeah I'm down for `::`, it's kind of consistent.  
 `;` is special in shell, generally.

brennanjl: I guess would that get confusing with SQL? Since we would be doing `::base64`, which isn't really a type.

brennanjl: For completeness: As per Jon's rec, I made it `<value_string>[b64]`

Yaiba: That's a good point.  I don't have good suggestions on other symbols :/

Issue #762, opened 2024-05-23
Bug: unexpected type in for loops


outerlook:
### Version / Commit

current `main` branch

### Operating System

Ubuntu 22.04.4 LTS

### Configuration

none

### Expected Behavior

while trying to parse

```
database test;

procedure test() public {
    $values text[] := ['a', 'b'];

    $new_val text;
    for $row in $values {
        $new_val := $row;
    }
}
```

should work just fine

### Actual Behavior

I get 
```
(procedure) type error: text[] != text
start 8:9 end 8:25
```
it's specific to this line

```
$new_val := $row;
```

if you use `$new_val text[];` it works

### Steps to Reproduce

```
kwil-cli utils parse ./test.kf
```

### Additional Information

we know it's not even released yet :sweat_smile:

outerlook:
Also, for additional context, the more real use case was this:
```
...
    $readonly_keys text[] := [
        'type',
        'stream_owner',
        'readonly_key',
        'taxonomy_version'
    ];

    for $key in $readonly_keys {
        $current_uuid := uuid_generate_v5($current_uuid, @txid);
        INSERT INTO metadata (row_id, metadata_key, value_s, created_at)
            VALUES ($current_uuid, 'readonly_key', $key, $current_block);
    }
...
```

brennanjl:
Fixed: https://github.com/kwilteam/kwil-db/pull/765

PR #765, opened 2024-05-24
Fix loops

brennanjl:
Fixes a type bug for loops over arrays. See more on the bug here: https://github.com/truflation/tsn/issues/257#issuecomment-2128351110

This is based on https://github.com/kwilteam/kwil-db/pull/764, and therefore should go in after it. Most of the diff is due to this.


brennanjl:
Also solves: https://github.com/kwilteam/kwil-db/issues/762

Yaiba:
This fix looks good to me

Yaiba (approved):

Yaiba (approved):

Issue #763, opened 2024-05-23
Bug: unexpected result in procedure null comparison


outerlook:
### Version / Commit

main branch

### Operating System

Ubuntu 22.04.4 LTS

### Configuration

none

### Expected Behavior

```
procedure test2() public view returns (result bool) {
    $value int;
    return $value == null;
}
```

Should return true, right? 

### Actual Behavior

I get 

```
| result |
+--------+
| <nil>  |
```

Or if I use this it's falsy

### Steps to Reproduce

```
kwil-cli utils parse ./test.kf
kwil-cli database deploy -n=test -p=./test_contract.kf
kwil-cli database call -n=test -a=test2
```

### Additional Information

this code gets compiled to the routine:
```
create function test2(OUT _out_0 boolean) returns boolean
	language plpgsql
as $$
DECLARE
_param_value INT8;
BEGIN
_out_0 := _param_value = NULL;
RETURN;

END;
$$;

alter function test2(out boolean) owner to kwild;
```

but `IS NULL` should be used instead of `= NULL`, right? 

Also, workarounds are easy, but create a hard to spot bug because it deploys successfully

brennanjl:
Yes, you are correct. I'm a bit surprised it doesn't properly evaluate `_out_0 := _param_value = NULL;`. Will get this addressed.

brennanjl:
So using `is` definitely works for null, but if we were doing any other data types (e.g. `_param_value = 0`), it gives the same behavior of an undefined result. Very annoying, working on a solution.

brennanjl:
Fix is up here: https://github.com/kwilteam/kwil-db/pull/767

PR #767, opened 2024-05-24
Fix null comparison

brennanjl:
Fixes https://github.com/kwilteam/kwil-db/issues/763.

Equality operators now use `IS DISTINCT`. See here for explanation: https://stackoverflow.com/questions/36508815/not-equal-and-null-in-postgres.

This pr also fixes an issue where procedure calls that do not receive the correct number of arguments give a confusing postgres error. It now gives a more descriptive error.

Should be merged after https://github.com/kwilteam/kwil-db/pull/765

### DIFF
--- a/internal/engine/generate/plpgsql.go
+++ b/internal/engine/generate/plpgsql.go
@@ -264,9 +264,8 @@ func (s *sqlGenerator) VisitExpressionIs(p0 *parse.ExpressionIs) any {
 		str.WriteString("NOT ")
 	}
 	if p0.Distinct {
-		str.WriteString("DISTINCT ")
+		str.WriteString("DISTINCT FROM ")
 	}
-	str.WriteString("FROM ")
 	str.WriteString(p0.Right.Accept(s).(string))
 	// cannot be typecasted
 	return str.String()
--- a/internal/engine/generate/plpgsql.go
+++ b/internal/engine/generate/plpgsql.go
@@ -190,6 +190,7 @@ func (s *sqlGenerator) VisitExpressionComparison(p0 *parse.ExpressionComparison)
 	str.WriteString(p0.Left.Accept(s).(string))
 	str.WriteString(" ")
 	str.WriteString(string(p0.Operator))
+	str.WriteString(" ")
 	str.WriteString(p0.Right.Accept(s).(string))
 	// compare cannot be typecasted
 	return str.String()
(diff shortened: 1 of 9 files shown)


parse/ast.go
```diff
@@ -246,8 +246,10 @@ func (e *ExpressionComparison) Accept(v Visitor) any {
 type ComparisonOperator string
 
 const (
-	ComparisonOperatorEqual              ComparisonOperator = "="
-	ComparisonOperatorNotEqual           ComparisonOperator = "!="
+	// We use distinct for equality because it can handle nulls.
+	// see https://stackoverflow.com/questions/36508815/not-equal-and-null-in-postgres
+	ComparisonOperatorEqual              ComparisonOperator = "IS NOT DISTINCT FROM"
+	ComparisonOperatorNotEqual           ComparisonOperator = "IS DISTINCT FROM"
```
jchappelow, line 252: What if they wanted the expected behavior of `=` and `!=`?  My first though about the linked but  is that the procedure was written incorrectly.

Comparison with null in sql has similar semantics to other languages, and if we break from that then others more familiar might make an issue with the exact opposite "bug"?
https://kwilteam.slack.com/archives/C05RZNFKNJD/p1705595383455259?thread_ts=1705594086.484569&cid=C05RZNFKNJD

brennanjl: Yeah that is fair. I guess we can add in `is distinct` expressions to procedures

Yaiba: Yeah agree with @jchappelow, it's different behavior and we should keep it same as postgres

internal/engine/generate/plpgsql.go
```diff
@@ -265,6 +266,7 @@ func (s *sqlGenerator) VisitExpressionIs(p0 *parse.ExpressionIs) any {
 	if p0.Distinct {
 		str.WriteString("DISTINCT ")
 	}
+	str.WriteString("FROM ")
```
Yaiba, line 269: Should this be in same block as `str.WriteString("DISTINCT ")` ?

Yaiba (approved):
lgtm

internal/engine/generate/plpgsql.go
brennanjl: Good catch

Yaiba (approved):

Issue #755, opened 2024-05-17
Problem: Reader Starvation Impacts Consensus


brennanjl:
@pkoch saw consensus halt while running v0.7.2. We are fairly confident that this is because he also experienced reader starvation, which was due to failure to release read-only connections. @jchappelow fixed the release failure bug in https://github.com/kwilteam/kwil-db/pull/744, and it was included in v0.7.6.

While the root cause of the issue has been solved, it still highlights the potential for readers to slow down consensus, which should never be the case. In order to solve for this, all database connections in the consensus path should be readwrite connections. Below, I have linked all of the connections I can find in the consensus / proposal path that use reader connections:

- https://github.com/kwilteam/kwil-db/blob/553bb6f39c8ce6278c505d55ac08a80512e5ef4d/internal/txapp/txapp.go#L765
- https://github.com/kwilteam/kwil-db/blob/553bb6f39c8ce6278c505d55ac08a80512e5ef4d/internal/abci/abci.go#L231
- https://github.com/kwilteam/kwil-db/blob/553bb6f39c8ce6278c505d55ac08a80512e5ef4d/internal/abci/abci.go#L719
- https://github.com/kwilteam/kwil-db/blob/553bb6f39c8ce6278c505d55ac08a80512e5ef4d/internal/abci/abci.go#L849
- https://github.com/kwilteam/kwil-db/blob/553bb6f39c8ce6278c505d55ac08a80512e5ef4d/internal/voting/broadcast/broadcast.go#L92
- https://github.com/kwilteam/kwil-db/blob/553bb6f39c8ce6278c505d55ac08a80512e5ef4d/internal/voting/broadcast/broadcast.go#L137

These are pretty sneaky. I'm open to suggestions on how we could make the connection handling more obvious; the best thing I can think of is to handle creation of transactions in ABCI.

jchappelow:
> These are pretty sneaky. I'm open to suggestions on how we could make the connection handling more obvious; the best thing I can think of is to handle creation of transactions in ABCI.

Right, we solved this sneakiness with explicit db transactions are function arguments, but we held onto the old design pattern with how ABCI uses TxApp.  I don't know if we want to make ABCI the transaction manager though.  Thinking about it.

>  Below, I have linked all of the connections I can find in the consensus / proposal path that use reader connections:

While the ones in proposal prepare or process aren't holding up block processing on non-validators, this is a problem too.

jchappelow:
[...]

PR #768, opened 2024-05-24
abci,txapp: on DB read pool use on consensus conn

jchappelow:
Resolves https://github.com/kwilteam/kwil-db/issues/755

The gist of the reserved connections is:

```go
// BeginReservedReadTx starts a read-only transaction using a reserved reader
// connection. This is to allow read-only consensus operations that operate
// outside of the write transaction's lifetime, such as proposal preparation and
// approval, to function without contention on the reader pool that services
// user requests.
```

So, some stuff can use `r.currentTx` (the consensus write connection) but in other place like `ProcessProposal` we are not in block execution but we don't want to fight for a read connection from RPC users.

### DIFF
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -228,7 +228,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		return nil, fmt.Errorf("begin tx commit failed: %w", err)
 	}
 
-	initialValidators, err := a.txApp.GetValidators(ctx)
+	initialValidators, err := a.txApp.ConsensusValidators(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get current validators: %w", err)
 	}
@@ -672,7 +672,7 @@ type indexedTxn struct {
 // This also includes the proposer's transactions, which are not in the mempool.
 // The transaction ordering is as follows:
 // MempoolProposerTxns, ProposerInjectedTxns, MempoolTxns by other senders
-func (a *AbciApp) prepareBlockTransactions(txs [][]byte, log *log.Logger, maxTxBytes int64, proposerAddr []byte) [][]byte {
+func (a *AbciApp) prepareBlockTransactions(ctx context.Context, txs [][]byte, log *log.Logger, maxTxBytes int64, proposerAddr []byte) [][]byte {
 	// Unmarshal and index the transactions.
 	var okTxns []*indexedTxn
 	var i int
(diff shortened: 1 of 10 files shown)


internal/abci/abci.go
```diff
@@ -228,7 +228,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		return nil, fmt.Errorf("begin tx commit failed: %w", err)
 	}
 
-	initialValidators, err := a.txApp.GetValidators(ctx)
+	initialValidators, err := a.txApp.ConsensusValidators(ctx)
```
jchappelow, line 231: Please see revised `TxApp interface`: https://github.com/kwilteam/kwil-db/pull/768/files#diff-933cda2247bf956073a324e8191848e3f62f7a8a5aa984a13d43933db7345c4aR39

and the docs on the actual methods e.g. https://github.com/kwilteam/kwil-db/pull/768/files#diff-29b2bc6069630247b4e88988ee7f8c5a0c4fa090867fc9d1e5d050c64da892d1R299

Short of merging AbciApp and TxApp, or bringing DB into the purview of AbciApp, we just need to solve a couple things with docs.

brennanjl:
This all looks fine to me. Curious though: what was the reason for not just handling creation of db txs in abci, instead of txapp, and then just passing around the same writer everywhere?

jchappelow:
> This all looks fine to me. Curious though: what was the reason for not just handling creation of db txs in abci, instead of txapp, and then just passing around the same writer everywhere?

Lots more upheaval is all.

brennanjl (dismissed):

jchappelow:
But the need wasn't tremendously strong to make that change.  The only awkwardness in that regard is the getters for validators and account info.  I passed around the tx in other spots where it should have from within TxApp though.

Presently testing a change to `pg` code that uses pools even for the writer as I think that might help keep alive the connections idle connections.

jchappelow:
OK.  Sorry for the large post-approval changes in https://github.com/kwilteam/kwil-db/pull/768/commits/8ebebaf760f845460e06306c023b6c0e19aa2500

Turns out that a pgpool `Pool` has some benefits even if you don't use more than one connection at a time.  These include auto reconnnect and connection health checks.  This was a concern for me with a mostly idle connection (the reserved one just added) since it would go unused on a sentry not until they became a validator.

It also occurs to me that having a writer acquired from a "pool" when `Begin` is called for block execution makes use of `pgpool` reconnect logic and it could help with some dropped connections.

I'll be looking at this for the replication connection, which is quite different and low level, making reconnect much more difficult.

For now, it's working as expected.  Probably good to get it in beta in case there are gotchas to discover.

jchappelow:
BTW, it's dead simple to [eliminate this `reserved` pool](https://github.com/kwilteam/kwil-db/commit/7fa0c410bf6157d44f96402410fc52d1ff8d35aa) and just grab from the `writer` "pool" (should not be concurrent use of it given ABCI connection concurrency promises) but I don't really wanna risk a deadlock right now in case something is missed.

brennanjl (approved):

Issue #616, opened 2024-03-26
improvement: String() method on AstNode


Yaiba:
### Describe the Feature

having this method could make testing mush easier, for example, sqlparser

jchappelow:
Is this still relevant?  Need to be in 0.8.0 milestone?

Yaiba:
Maybe the title is not accurate, not actually a feature, just improvement. Better to have, I think it's a lot helpful for testing

PR #769, opened 2024-05-24
add missing return detection

brennanjl:
This PR adds several tests, including:

- tests (and bug fixes) that ensure that all procedures that need a return have one. Without this check, users can hit unexpected plpgsql errors.
- tests for default ordering for aggregate and compound selects.
- tests for non-determinism checks with aggregates and group bys

### DIFF
--- a/parse/parse_test.go
+++ b/parse/parse_test.go
@@ -1198,6 +1198,231 @@ func Test_Procedure(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "missing return values",
+			returns: &types.ProcedureReturn{
+				Fields: []*types.NamedType{
+					{
+						Name: "id",
+						Type: types.IntType,
+					},
+					{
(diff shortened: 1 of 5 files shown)


parse/parse_test.go
```diff
@@ -1613,6 +1789,152 @@ func Test_SQL(t *testing.T) {
 			INNER JOIN (SELECT id as uid FROM users WHERE id = 1) ON p.author_id = uid;`,
 			err: parse.ErrUnnamedJoin,
 		},
+		{
+			name: "compound select",
+			sql:  `SELECT * FROM users union SELECT * FROM users;`,
+			want: &parse.SQLStatement{
```
Yaiba, line 1817: do you consider add a `format() string` or similiar funcion, so in testing, you can just compare to a string? I can be really handy.    That's #616

Yaiba (approved):

Issue #770, opened 2024-05-27
Bug: foreign call that returns tables doesn't correctly check value and names


outerlook:
### Version / Commit

main branch (`327a34f`)

### Operating System

Ubuntu 22.04.4 LTS

### Configuration

None

### Expected Behavior

```
foreign procedure ext_get_record($date_from text, $date_to text, $frozen_at text) returns table(
    date_value text,
    value int
)
```

should work, I guess (or is there a syntax error?)

### Actual Behavior

```
procedure get_unsafe_record($dbid text, $date_from text, $date_to text, $frozen_at text) public view returns table(
    date_value text,
    value int
) {
    for $row in SELECT * FROM ext_get_record[$dbid, 'get_record']($date_from, $date_to, $frozen_at) {
        return next $row.date_value, $row.value;
    }
}
```

Produces error

```
ERROR: Foreign procedure definition "ext_get_record" expects return name "date_value" at column position 1, but procedure "get_record" requires text (SQLSTATE P0001)
```

### Steps to Reproduce

- deploy contract with a foreign function that returns tables
- try getting dataa 

### Additional Information

while searching for the cause, I found that this was the generated code for the foreign call:
```
create function _fp_ext_get_record(_dbid text, _procedure text, _arg1 text, _arg2 text, _arg3 text) returns TABLE(date_value text, value bigint)
	language plpgsql
as $$ DECLARE
    _schema_owner BYTEA;
    _is_view BOOLEAN;
    _is_owner_only BOOLEAN;
    _is_public BOOLEAN;
	_returns_table BOOLEAN;
    _expected_input_types TEXT[];
    _expected_return_types TEXT[];
BEGIN
	SELECT p.param_types, p.return_types, p.is_view, p.owner_only, p.public, s.owner, p.returns_table
	INTO _expected_input_types, _expected_return_types, _is_view, _is_owner_only, _is_public, _schema_owner, _returns_table
	FROM kwild_internal.procedures as p INNER JOIN kwild_internal.kwil_schemas as s
	ON p.schema_id = s.id
	WHERE p.name = _procedure AND s.dbid = _dbid;

	IF _schema_owner IS NULL THEN
		RAISE EXCEPTION 'Procedure "%" not found in schema "%"', _procedure, _dbid;
	END IF;

	IF _is_view = FALSE AND current_setting('is_read_only') = 'on' THEN
		RAISE EXCEPTION 'Non-view procedure "%" called in view-only connection', _procedure;
	END IF;

[...]

PR #771, opened 2024-05-28
fix bug when writing to foreign proc

brennanjl:
This PR fixes several bugs, as well as adds better error messages and adds many tests for foreign procedures.

It fixes the bug found by  @KwilLuke where writing data to a foreign procedure failed, as well as https://github.com/kwilteam/kwil-db/issues/770. It adds tests for both of them. It also adds many negative test cases to check for error messages thrown during runtime.


parse/analyze.go
```diff
@@ -2186,21 +2228,31 @@ func (p *procedureAnalyzer) VisitProcedureStmtCall(p0 *ProcedureStmtCall) any {
...
 		// if it returns null, then we do not need to assign it to a variable.
 		if !returns1.EqualsStrict(types.NullType) {
 			callReturns = append(callReturns, returns1)
 		}
-	} else {
-		// or it might return multiple values
-		returns2, ok := p0.Call.Accept(p).([]*types.DataType)
-		if !ok {
-			p.errs.AddErr(p0.Call, ErrType, "expected function/procedure to return one or more variables")
+	} else if returns2, ok := returnVal.([]*types.DataType); ok {
```
Yaiba, line 2239: probably should consider refactor this using switch?

parse/analyze.go
```diff
@@ -439,7 +438,18 @@ func (s *sqlAnalyzer) expressionTypeErr(e Expression) *types.DataType {
 	if ok {
 		s.errs.AddErr(e, ErrType, "expected procedure to return a single value, returns %d", len(vals))
 		return cast(e, types.UnknownType)
+	}
 
+	_, ok = e.Accept(s).(*returnsTable)
```
Yaiba, line 443: looks like you can refactor this function switch

internal/engine/integration/procedure_test.go
```diff
@@ -206,56 +163,198 @@ func Test_Procedures(t *testing.T) {
...
+		wantErr string
+	}
+
+	tests := []testcase{
+		{
+			name:    "foreign procedure takes nothing, returns nothing",
+			foreign: `foreign procedure do_something()`,
+			otherProc: `procedure call_foreign() public {
+				do_something['%s', 'delete_users']();
+			}`,
```
Yaiba, line 228: what's the expected result?  expect no err and no outputs?

internal/engine/integration/procedure_test.go
```diff
@@ -206,56 +163,198 @@ func Test_Procedures(t *testing.T) {
...
+		{
+			name:    "foreign procedure expects 1 arg, receives 2",
+			foreign: `foreign procedure id_from_name($name text, $name2 text) returns (id uuid)`,
+			otherProc: `procedure call_foreign() public returns (id uuid) {
+				return id_from_name['%s', 'id_from_name']('satoshi', 'zeus');
+			}`,
+			wantErr: "requires 1 arg(s)",
+		},
+		{
+			name:    "foreign procedure returns 1 arg, implementation returns none",
```
Yaiba, line 276: As you pointed out, the test name since here are less confusing. It would be better we change above test case name

Yaiba (changes requested):
LGTM, didn't go through the test cases, just skimmed

internal/engine/integration/procedure_test.go
brennanjl: nothing. No error and no outputs, as the procedure should be successful and it does not return anything.

brennanjl:
Ok I addressed the changes @Yaiba

Yaiba (dismissed):
LGTM

Yaiba (approved):

PR #772, opened 2024-05-28
added crypto functions, encode/decode

brennanjl:
This PR adds three functions from Postgres:

- encode/decode: https://www.postgresql.org/docs/current/functions-binarystring.html
- digest: https://www.postgresql.org/docs/current/pgcrypto.html#PGCRYPTO-GENERAL-HASHING-FUNCS-DIGEST

This was requested by the Truflation team, so they can generate DBIDs within a procedure: https://github.com/truflation/tsn/issues/280#issuecomment-2133458185

I also felt that these were pretty reasonable functions to want to have access to.

### DIFF
--- a/parse/functions.go
+++ b/parse/functions.go
@@ -175,6 +175,90 @@ var (
 				return fmt.Sprintf("uuid_generate_v5(%s)", strings.Join(inputs, ", ")), nil
 			},
 		},
+		"encode": {
+			ValidateArgs: func(args []*types.DataType) (*types.DataType, error) {
+				// first must be blob, second must be text
+				if len(args) != 2 {
+					return nil, wrapErrArgumentNumber(2, len(args))
+				}
+
+				if !args[0].EqualsStrict(types.BlobType) {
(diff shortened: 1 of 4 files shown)


parse/functions.go
```diff
@@ -175,6 +175,90 @@ var (
...
+				}
+				if distinct {
+					return "", errDistinct("encode")
+				}
+
+				return fmt.Sprintf("encode(%s)", strings.Join(inputs, ", ")), nil
+			},
+		},
+		"decode": {
+			ValidateArgs: func(args []*types.DataType) (*types.DataType, error) {
```
Yaiba, line 207: Do we have function return validation? 
When we validate input for example of a foregin procedure call, we will need it right?

Yaiba (approved):
LGTM, just have a question about function return validation.

parse/functions.go
Yaiba: ah I see, it's already defined in the `ValidateArgs`, all good.

Issue #774, opened 2024-05-28
Bug: Maximum uint256 value cannot be stored


KwilLuke:
### Version / Commit

main

### Operating System

Darwin-Arm64

### Configuration

I am using this schema:

database variable_test;

table var_table {
    uuid_col uuid primary,
    text_col text,
    int_col int,
    bool_col bool,
    blob_col blob,
    uint256_col uint256
}

action insert_uint256 ($id, $uint256) public {
    INSERT INTO var_table (uuid_col, uint256_col)
    VALUES ($id, $uint256);
}

### Expected Behavior

I try to execute the `insert_uint256` action with the maximum uint256 value (`115792089237316195423570985008687907853269984665640564039457584007913129639935`), and I expect the variable to be stored in the database.

### Actual Behavior

When checking the transaction result, I receive this error:

```
{"tx_code":4294967295,"hash":"12e8e5725bcd3f29d385746fb4240437197223258e330bb94386142b5a86a356","message":"ERROR: value for domain uint256 violates check constraint \"uint256_check\" (SQLSTATE 23514)"}
```

### Steps to Reproduce

Deploy the schema:

```
kwil-cli database deploy -p variable_test.kf
```

Execute the action:

```
kwil-cli database execute id:8fc2d081-b8c9-4ac6-ab48-c326f4eecb2c uint256:115792089237316195423570985008687907853269984665640564039457584007913129639935 -a insert_uint256 -i your_dbid
```

PR #775, opened 2024-05-28
fix incorrect max calculation in  uint256 domain

brennanjl:
Fixes https://github.com/kwilteam/kwil-db/issues/774

### DIFF
--- a/test/specifications/types.go
+++ b/test/specifications/types.go
@@ -168,6 +168,15 @@ func testUint256(ctx context.Context, t *testing.T, execute ProcedureDSL, dbid s
 		}
 	}
 	assert.Equal(t, 1, count)
+
+	// test that greater than max uint256 is not allowed
+	uint256, err = types.Uint256FromString("115792089237316195423570985008687907853269984665640564039457584007913129639936")
+	require.Error(t, err)
+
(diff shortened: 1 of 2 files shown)


jchappelow (dismissed):
OK as per https://kwilteam.slack.com/archives/C05RZNFKNJD/p1716930957001749?thread_ts=1716924928.033079&cid=C05RZNFKNJD
Magic numbers kinda suck, but I think it's well documented.  The alternative is `power(2::numeric(78), 256)`, which is a little more expressive, but meh

test/specifications/types.go
```diff
@@ -142,7 +142,8 @@ func testDecimal(ctx context.Context, t *testing.T, execute ProcedureDSL, dbid s
 
 func testUint256(ctx context.Context, t *testing.T, execute ProcedureDSL, dbid string, callType string) {
 	// execute store
-	uint256 := types.Uint256FromInt(123)
+	uint256, err := types.Uint256FromString("115792089237316195423570985008687907853269984665640564039457584007913129639935") // max uint256
```
jchappelow, line 145: Do you wanna test that +1 fails?

jchappelow:
Eh, I think the magic number is actually better than the `power` solution in case pg is dumb and computes it every time.

jchappelow (approved):

PR #776, opened 2024-05-29
added debug for cli parse

brennanjl:
This adds debug functionality for `kwil-cli utils parse`. It replaces https://github.com/kwilteam/kwil-db/pull/739

### DIFF
--- a/parse/types.go
+++ b/parse/types.go
@@ -11,10 +11,10 @@ type Position struct {
 	// Set is true if the position of the Position has been set.
 	// This is useful for testing parsers.
 	IsSet     bool `json:"-"`
-	StartLine int  `json:"start_line"`
-	StartCol  int  `json:"start_col"`
-	EndLine   int  `json:"end_line"`
-	EndCol    int  `json:"end_col"`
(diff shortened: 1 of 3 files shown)


parse/types.go
```diff
@@ -11,10 +11,10 @@ type Position struct {
 	// Set is true if the position of the Position has been set.
 	// This is useful for testing parsers.
 	IsSet     bool `json:"-"`
-	StartLine int  `json:"start_line"`
-	StartCol  int  `json:"start_col"`
-	EndLine   int  `json:"end_line"`
-	EndCol    int  `json:"end_col"`
+	StartLine int  `json:"-"`
+	StartCol  int  `json:"-"`
+	EndLine   int  `json:"-"`
```
jchappelow, line 16: No opinion on this, but I'm curious why these needed to be hidden.  Just too noisy in the document?

brennanjl: Yeah, so much noise that it makes the document useless. There's also never really a case where these are necessary in the debug AST, since it is only used for error positions.

Yaiba (approved):
LGTM

PR #777, opened 2024-05-29
added basic string manipulation functions

brennanjl:
Added basic string manipulation functions. These are taken from the top-level SQL functions identified in Postgres: https://www.postgresql.org/docs/16/functions-string.html

This was requested by Truflation here: https://github.com/truflation/tsn/issues/284#issuecomment-2136359976

Documentation for supported functions can be found here: https://github.com/kwilteam/docs/pull/105/files#diff-aacab0e9719a83529316f1b99149bb81e25da9b918b53725d801890376d811a7

@outerlook does this cover what you need?

### DIFF
--- a/internal/engine/integration/procedure_test.go
+++ b/internal/engine/integration/procedure_test.go
@@ -1,4 +1,4 @@
-// go:build pglive
+//go:build pglive
 
 package integration_test
 
--- a/internal/engine/integration/procedure_test.go
+++ b/internal/engine/integration/procedure_test.go
@@ -187,30 +187,56 @@ func Test_Procedures(t *testing.T) {
 				if $val != 'hello world' {
 					error('rtrim failed');
 				}
+				if rtrim($val||' ') != 'hello world' {
+					error('rtrim 2 failed');
+				}
+
 				$val := ltrim($val, 'h');
 				if $val != 'ello world' {
 					error('ltrim failed');
 				}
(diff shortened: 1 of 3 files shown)


jchappelow (approved):
The list of functions looks fine to me.   Scanned the checks quickly and they also look fine.  Only question is if we need/want to support all the synonyms like char_length/character_length.  The tech burden isn't bad.

brennanjl:
Yeah I wasn't sure. I'd be open to removing one of them, but also have no strong preference.

Yaiba (dismissed):
LGTM

internal/engine/integration/procedure_test.go
```diff
@@ -160,6 +160,60 @@ func Test_Procedures(t *testing.T) {
...
+				if $val != 'hello world!' {
+					error('rpad failed');
+				}
+				if overlay($val, 'xx', 2, 5) != 'hxxworld!' {
+					error('overlay failed');
+				}
+				if position('world', $val) != 7 {
+					error('position failed');
+				}
+				if substring($val, 7, 5) != 'world' {
```
Yaiba, line 208: maybe also add test for different args? functions like `ltrim`,`overlay`,`substring`(and more) support different args num

brennanjl: ok done

Yaiba (approved):

PR #779, opened 2024-05-30
Feature: min and max

brennanjl:
This PR adds the min and max aggregate functions. We have had a few people have problems not having this, so I felt it was appropriate to add. It required a non-trivial change to how our lexer and parser work, since `max` and `min` were keywords, and were therefore being lexed and not recognized as function names.


brennanjl:
This PR is based on https://github.com/kwilteam/kwil-db/pull/777

jchappelow (dismissed):
The `[added min and max](https://github.com/kwilteam/kwil-db/pull/779/commits/ad702ea4f9736d254a560f03516485f2fe62f6d7)` commit looks fine to me, but @Yaiba would have more to say.

What's the `primary_key` vs `primary key` change?  Is that something to note in https://github.com/kwilteam/kwil-db/discussions/668?

brennanjl:
> What's the primary_key vs primary key change?

To specify a primary key, we support:
- `primary_key`
- `pk`
- `primary`
- `primary key`

I simply changed the parser unit tests to test all of these.

parse/grammar/KuneiformParser.g4
```diff
@@ -127,14 +127,8 @@ typed_variable_list:
...
-    MIN LPAREN literal RPAREN           # min_constraint
-    | MAX LPAREN literal RPAREN         # max_constraint
-    | MINLEN LPAREN literal RPAREN      # min_len_constraint
-    | MAXLEN LPAREN literal RPAREN      # max_len_constraint
-    | (NOTNULL|NOT NULL)                # not_null_constraint
-    | (LEGACY_PRIMARY_KEY|PRIMARY KEY?) # primary_key_constraint
-    | DEFAULT LPAREN literal RPAREN     # default_constraint
-    | UNIQUE                            # unique_constraint
+    // conditionally allow some tokens, since they are used elsewhere
+    (IDENTIFIER| PRIMARY KEY? | NOT NULL | DEFAULT | UNIQUE) (LPAREN literal RPAREN)?
```
Yaiba, line 131: seems no need to declare `DEFAULT` & `UNIQUE` here, in this rule we only need either a single word `IDENTIFIER` or multi word keywords.

Yaiba (approved):
LGTM

parse/grammar/KuneiformParser.g4
brennanjl: UNIQUE and DEFAULT are both tokens used else where. DEFAULT is used in foreign keys: `on update set default`, and unique is used in indexes: `#idx unique(column1, column2)`

Yaiba: ah I just mean for `constrains` rule, since we check the 'syntax' later, but nbd

Yaiba (approved):

PR #782, opened 2024-05-30
V07 snapshots

brennanjl:
This is a little bit hacky, but I believe it is the best we can do. This _should_ work for network migrations, however there could potentially be edge cases. I think I need to spend some more time thinking about how this can be done in the future, but as far as this release goes, it should work.

It ended up being less code to write that I thought, but it does feel extremely messy and hard to maintain long term.

### DIFF
--- a/internal/engine/execution/queries.go
+++ b/internal/engine/execution/queries.go
@@ -263,11 +263,20 @@ func createSchema(ctx context.Context, tx sql.TxMaker, schema *types.Schema, txi
 	return sp.Commit(ctx)
 }
 
-// getSchemas returns all schemas in the kwil_schemas table
-func getSchemas(ctx context.Context, tx sql.Executor) ([]*types.Schema, error) {
+// getSchemas returns all schemas in the kwil_schemas table.
+// convertFunc converts bytes into a schema. If nil, it will simply unmarshal the bytes.
+func getSchemas(ctx context.Context, tx sql.Executor, convertFunc func([]byte) (*types.Schema, error)) ([]*types.Schema, error) {
 	res, err := tx.Execute(ctx, sqlListSchemaContent)
 	if err != nil {
 		return nil, err
+	}
+
+	if convertFunc == nil {
+		convertFunc = func(b []byte) (*types.Schema, error) {
+			schema := &types.Schema{}
+			err := json.Unmarshal(b, schema)
+			return schema, err
+		}
 	}
 
 	schemas := make([]*types.Schema, len(res.Rows))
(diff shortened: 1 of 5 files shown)


internal/engine/execution/schema_v07.go
```diff
@@ -0,0 +1,224 @@
...
+	"encoding/json"
+
+	"github.com/kwilteam/kwil-db/core/types"
+)
+
+// this file contains the schema from Kwil v0.7, which is used for proper migration to
+// the new schema in v0.8
+
+// convertV07Schema converts a v0.7 schema to a v0.8 schema
+func convertV07Schema(bts []byte) (*types.Schema, error) {
```
jchappelow, line 13: This function and this file look fairly straightforward.  Some suggestions though: https://github.com/jchappelow/kwil-db/commit/15607b09c86bb3b897098d6f4df28cef7e0e8548
The only thing that looks like a bug is the "bool" vs. "BOOLEAN" change in the data type string.  Also case for all the types, but the code uses `strings.ToLower` a lot so case probably wouldn't matter.

internal/engine/execution/queries.go
```diff
@@ -263,13 +263,22 @@ func createSchema(ctx context.Context, tx sql.TxMaker, schema *types.Schema, txi
...
+// getSchemas returns all schemas in the kwil_schemas table.
+// convertFunc converts bytes into a schema. If nil, it will simply unmarshal the bytes.
+func getSchemas(ctx context.Context, tx sql.Executor, convertFunc func([]byte) (*types.Schema, error)) ([]*types.Schema, error) {
 	res, err := tx.Execute(ctx, sqlListSchemaContent)
 	if err != nil {
 		return nil, err
 	}
 
+	if convertFunc == nil {
+		convertFunc = func(b []byte) (*types.Schema, error) {
```
jchappelow, line 275: Related concept, not requesting any changes, just food for thought: https://github.com/jchappelow/kwil-db/commit/423cdc6becfbe6690a5e054a9b4405f206345365
At least within the framework of JSON schema content, there are some simple ways to support reading prior schema versions if the schema doc and struct include a flag.  I suppose the only format we want to support migrating directly from text is KF, but we also support specifying a schema with JSON from kwil-cli so this is an exposed document structure with no real end-user accessible conversion.

Do we want to ditch the JSON option?  `  -t, --type string   file type of the database definition file (kf or json) (default "kf")`

brennanjl: I'm a bit confused, why would we need to ditch the json option in the CLI? It's just used to submit an already parsed schema. I don't see how it is related at all, other than that schemas parsed with v0.7 will not be readable by the CLI

charithabandi:
Screenshot 2024-06-07 at 3 32 31 PM
I see these debugs. need to figure out why. Not sure if something is the issue with the voters/validators.

internal/engine/execution/queries.go
jchappelow: It seems equally valid to provide your schema as an end user to the cli in kf or json.  Is it?

charithabandi:
Ah! I had my node's private key wrong, looks good. Sorry for the alert.

internal/engine/execution/queries.go
brennanjl: It is. They obtain the json by using the parser though. I don't think we have ever directed anyone to construct a schema via json, they simply get it from the IDE (which has an export parse to json) feature.

charithabandi (approved):

PR #783, opened 2024-05-31
Additional testing

brennanjl:
This added new tests for Kuneiform, as well as fixed a few bugs caught by these tests.

The new tests include:
- tests for properly scoped variables in for loops and if blocks
- integration tests for all SQL statements and expressions
- basic tests for decimal math


internal/engine/generate/sql.go
```diff
@@ -10,11 +10,16 @@ import (
 // It can optionally rewrite named parameters to numbered parameters.
 // If so, it returns the order of the parameters in the order they appear in the statement.
 func WriteSQL(node *parse.SQLStatement, orderParams bool, pgSchema string) (stmt string, params []string, err error) {
+	if node == nil {
+		panic("statement ast node is nil")
```
jchappelow, line 14: There's a recover just below that converts to an error return.  Why hard panic here?

parse/analyze.go
```diff
@@ -95,6 +96,11 @@ func (b *blockContext) variableExists(name string) bool {
 	return ok
 }
 
+// copyVariables copies both the user variables and anonymous variables.
+func (b *blockContext) copyVariables() (map[string]*types.DataType, map[string]map[string]*types.DataType) {
+	return maps.Clone(b.variables), maps.Clone(b.anonymousVariables)
```
jchappelow, line 101: Do you need this to be a deep clone?  It's shallow, so in particular the map _inside_ the `anonymousVariables` map is not cloned.  I don't know the requirement

parse/parse.go
```diff
@@ -243,6 +243,11 @@ type SQLParseResult struct {
 // It requires a schema to be passed in, since SQL statements may reference
 // schema objects.
 func ParseSQL(sql string, schema *types.Schema) (res *SQLParseResult, err error) {
+	// add semicolon to the end of the statement, if it is not there
+	if sql[len(sql)-1] != ';' {
```
jchappelow, line 247: I don't see anything that prevents a query from being empty from, say  `(*Service).Query`.  Best to screen out empty queries first. and/or use `strings.HasSuffix`

jchappelow: Hmm, now I'm wondering what happens in other places like deployment if the `Body string` for an action or procedure does not contain valid utft. e.g. `string([]byte{random garbage that is not utf8})`.

I ask because:

```go
// string is the set of all strings of 8-bit bytes, conventionally but not
// necessarily representing UTF-8-encoded text. A string may be empty, but
// not nil. Values of string type are immutable.
type string string
```

We might be trusting that a string is valid but it doesn't have to be.  

`utf8.ValidString()` needs to be used in other contexts?  Do we test for garbage bodies in the Schema struct?

Yaiba: this is a valid concern. The antlr lexer will error since the 'token' won't be recognized I think

Yaiba: Just tested it, and antlr parser does complain

Yaiba: I added `{
name: "non utf-8",
sql:  "\xbd\xb2\x3d\xbc\x20\xe2\x8c\x98;",
err:  nil,
},` to  Test_SQL  

The error is:
`unexpected error: detected multiple parse errors:
0: (sql) syntax error: token recognition error at: '�'
start 1:1 end 1:1
1: (sql) syntax error: token recognition error at: '�'
start 1:2 end 1:2
2: (sql) syntax error: token recognition error at: '�'
start 1:4 end 1:4
3: (sql) syntax error: token recognition error at: '⌘'
start 1:6 end 1:6
4: (sql) syntax error: mismatched input '=' expecting {'delete', 'update', 'with', 'select', 'insert'}
start 1:3 end 1:3`

internal/engine/integration/procedure_test.go
```diff
@@ -256,6 +256,23 @@ func Test_Procedures(t *testing.T) {
...
+				$i := 100.423;
+				$j decimal(16,8) := 46728954.23743892;
+				$k := $i::decimal(16,8) + $j;
+				if $k != 46729054.66043892 {
+					error('decimal failed');
+				}
+				if $k::text != '46729054.66043892' {
+					error('decimal text failed');
+				}
+				if ($k::decimal(16,2))::text != '46729054.66' {
```
Yaiba, line 271: A dumb question, just curious, cut the scale won't round the value right? It's a silly question as I typed it out

Yaiba (changes requested):

internal/engine/integration/procedure_test.go
jchappelow:
```psql://postgres@[local]:5432/postgres 
# select 3.1415926::numeric(10,3);
┌─────────┐
│ numeric │
├─────────┤
│   3.142 │
└─────────┘
(1 row)

Time: 0.506 ms

```

That behavior?

Yaiba: ah so it does round the number?

Yaiba: I'm not sure the behavior, just want to clearfy what's the behavior

jchappelow: > If the scale of a value to be stored is greater than the declared scale of the column, the system will round the value to the specified number of fractional digits. Then, if the number of digits to the left of the decimal point exceeds the declared precision minus the declared scale, an error is raised. For example, a column declared as
> 
> NUMERIC(3, 1)
> 
> will round values to 1 decimal place and can store values between -99.9 and 99.9, inclusive.

https://www.postgresql.org/docs/current/datatype-numeric.html

jchappelow: The error case described looks like this:

```
# select 1234.6::numeric(5,2);
ERROR:  numeric field overflow
DETAIL:  A field with precision 5, scale 2 must round to an absolute value less than 10^3.
Time: 10.250 ms
```


5 - 2 = 3, but that requires 4 digits for 1234, so it errors.

Not sure if that's an issue for us, or just a predictable runtime error.

internal/engine/generate/sql.go
brennanjl: I don't remember why I did this. Will change

parse/analyze.go
brennanjl: We do not need to deep copy the nested map, because the nested map is representative of an object. Variable names have to be unique, so once an object is declared, its structure cannot change (e.g. it cannot get a new field).  I will add a note.

parse/parse.go
brennanjl: I've added Gavin's test, but I don't believe this will get parsed in the first place, so it should be ok.

internal/engine/integration/procedure_test.go
brennanjl: I think that is the runtime error we want

brennanjl:
Have addressed all feedback

jchappelow (approved):

Yaiba (approved):

PR #784, opened 2024-05-31
rpc: openrpc spec

jchappelow:
This adds OpenRPC spec generation.

The spec is generated automatically when services are registered. When a server is started, the spec is provided on both `/spec/v1` and the `rpc.discover` JSON-RPC method.

This also refactors service handler method registry creation using generics.

Getting the generated spec to properly use `$ref` objects and to pass validation at https://playground.open-rpc.org/ was a lot harder than expected.  Here it is: https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/jchappelow/kwil-db/openrpc/internal/services/jsonrpc/usersvc/user.openrpc.json&uiSchema[appBar][ui:title]=Kwil%20DB%20JSON-RPC%20API


internal/services/jsonrpc/methods.go
```diff
@@ -15,6 +17,44 @@ import (
 // that instance to perform some operations.
 type MethodHandler func(ctx context.Context, s *Server) (argsPtr any, handler func() (any, *jsonrpc.Error))
 
+type Handler[I, O any] func(context.Context, *I) (*O, *jsonrpc.Error)
+
+func ioTypes[I, O any](Handler[I, O]) (reflect.Type, reflect.Type) {
+	return reflect.TypeFor[I](), reflect.TypeFor[O]()
```
brennanjl, line 23: I'm getting a build error for `reflect.TypeFor`. Since `reflect.TypeFor` is only in go 1.22, we should update our go mod (which is currently 1.21) or use something else.

jchappelow: Oh, good catch.  It's just the following, so easily avoided:

```go
func TypeFor[T any]() Type {
	return TypeOf((*T)(nil)).Elem()
}
```

core/rpc/json/openrpc/reflect.go
```diff
@@ -0,0 +1,238 @@
...
+func typeToSchemaType(t reflect.Type) string {
+	// Some special cases first.
+	switch t {
+	case reflect.TypeOf((*big.Int)(nil)), reflect.TypeFor[big.Int]():
+		return "string"
+	case reflect.TypeFor[types.HexBytes]():
+		return "string"
+	case reflect.TypeFor[[]byte]():
+		return "string"
+	}
```
brennanjl, line 116: Should we also include our UUID, decimal, and uint256 types? Alternatively, we could just check for a fmt.Stringer, which applies to all of them.

brennanjl: I guess there aren't any definitions where we return these data types, however I would imagine there might be for UUID (e.g. for vote store related things)

jchappelow: Yeah, very well could be.  The `String` idea's not bad.  Will make some changes here.

core/rpc/json/user/commands.go
```diff
@@ -22,13 +22,13 @@ type VersionRequest struct{}
...
-	DBID string `json:"dbid,omitempty"`
+	DBID string `json:"dbid"`
 }
 
 // AccountRequest contains the request parameters for MethodAccount.
 type AccountRequest struct {
-	Identifier types.HexBytes `json:"identifier,omitempty"`
-	Status     *AccountStatus `json:"status,omitempty"` // Mapped to URL query parameter `status`.
+	Identifier types.HexBytes `json:"identifier" desc:"account identifier"`
+	Status     *AccountStatus `json:"status,omitempty" desc:"blockchain status (confirmed or unconfirmed)"` // Mapped to URL query parameter `status`.
```
jchappelow, line 31: Oh, I was contemplating pulling parameter descriptions from the field tags like this.  I think I will remove for now.

core/rpc/json/openrpc/reflect.go
```diff
@@ -0,0 +1,238 @@
...
+	RequestType  reflect.Type
+	ResponseType reflect.Type
+	RespTypeDesc string
+	// Errors ...
+}
+
+// InventoryAPI iterates through all handler types, making a map of all unique
+// OpenRPC schemas and building a []Method with every Param populated and
+// referencing a Schema if an "object" instead of a plain type.
+func InventoryAPI(handlerDefs map[string]*MethodDefinition, knownSchemas map[reflect.Type]Schema) []Method {
```
jchappelow, line 26: mmm, now I'm thinking to move this `openrpc` package to internal, at least for now.  Initially I had the service specs generated from types and such in `core`, but now I have the RPC server's generating them based on the services that are registered with them... and that's the only things using the types and functions defined here.

core/rpc/json/openrpc/reflect.go
jchappelow: Now covered recent data types that have `MarshalJSON` methods that return a `[]byte` that represents a JSON string, such as uint256, uuid, and decimal.
But AFAICT there's no special significance of a `fmt.Stringer` in any circumstance with json marshalling, and we cannot identify how a field will actually marshal without knowing the implementation of the `MarshalJSON` method defined for that type (or the parent type that contains this field type).
Updated this section of the code with docs explaining why we have to list exceptions here.

brennanjl (approved):

internal/services/jsonrpc/openrpc/reflect.go
```diff
@@ -0,0 +1,267 @@
...
+	RequestType  reflect.Type
+	ResponseType reflect.Type
+	RespTypeDesc string
+	// Errors ...
+}
+
+// InventoryAPI iterates through all handler types, making a map of all unique
+// OpenRPC schemas and building a []Method with every Param populated and
+// referencing a Schema if an "object" instead of a plain type.
+func InventoryAPI(handlerDefs map[string]*MethodDefinition, knownSchemas map[reflect.Type]Schema) []Method {
```
Yaiba, line 27: Curious if using https://pkg.go.dev/go/ast could make this easier haha.

I'm not suggest change implementation, just feel like it's a perfect use case for go/ast'

Yaiba (approved):
LGTM

internal/services/jsonrpc/openrpc/reflect.go
jchappelow: Not a bad point.  Establishing the hierarchy is what we're doing after all.  `go/ast` briefly occurred to me for the different job of getting method *descriptions* from the docs on types and struct fields too.
However, I wasn't sure about ast for this since afaict from the few uses I've seen of go/ast is that it has to build it from a source file.  Is there a way around that?  Looks like the strong suits of go/ast are for static code analysis and code generation, where as reflect is for dynamic (runtime) use such as when we register a bunch of concrete instantiations of services with a server.  Would love to learn more at some point, but I think it's working alright for now.

PR #785, opened 2024-06-03
added backwards compatibility for kwil-provider

brennanjl:
This adds backwards compatibility for the `--kwil-provider` flag, which was changed to `--provider` in this release. Users will get a warning when they use this flag, however it will still work.

I think it is important to have a deprecation period for this, since both Fractal and Truflation have critical infra built around our CLI.

### DIFF
--- a/cmd/kwil-cli/cmds/root.go
+++ b/cmd/kwil-cli/cmds/root.go
@@ -1,6 +1,8 @@
 package cmds
 
 import (
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


jchappelow (dismissed):
OK.  Why is `CliCfg` exported though?

brennanjl:
> OK. Why is CliCfg exported though?

Whoops. I had made a change where we needed to expose it, but then undid that change in favor of something else. Will fix

jchappelow (approved):

PR #787, opened 2024-06-03
fixed client docs, added parse

brennanjl:
This PR fixes the `core/client` docs to correctly reference the parse module, instead of the old kfparser repo. This was noticed by the Truflation team: https://github.com/truflation/tsn/issues/292

This also makes one small change to the parse submodule, which adds a top-level `Parse` function. This function is suitable for most usecases, as opposed to the other exposed parsing functions which expose much more information (and don't handle errors intuitively, since they are meant to be used in parsing tools). I have updated most places in our repo where we parse to use the `Parse` function.


jchappelow:
@brennanjl Pardon my push.  Just update the example app and it's go.mod.

You may still wish to update it to use the function suggested in the revised docs however.  It's not quite how I wanted the example app to flow, but it works for docs I think.

Remember that to work on the example app you probably either need to `go work use ./core/client/example` or do stuff with `GOWORK=off`.

brennanjl:
We can get rid of the test.json and the go:embed, right? It seems like you just use a hard-coded schema

brennanjl:
the `unmarshalKf` function in the example app now seems a bit useless, right? We create an io.Reader just to read it and parse.

jchappelow:
> We can get rid of the test.json and the go:embed, right? It seems like you just use a hard-coded schema

Yup, that's funky.  Pick whichever seems nicest, embed or literal string.

> the unmarshalKf function in the example app now seems a bit useless, right? We create an io.Reader just to read it and parse.

Yes again. That's pretty funny.  It definitely started with more purpose, but it's useless now.

jchappelow (approved):

PR #789, opened 2024-06-03
additional parse debug info

brennanjl:
This PR adds additional info for parse debugging. I originally made these changes to help me debug https://github.com/truflation/tsn/pull/295

It adds the generated plpgsql for debugging.

### DIFF
--- a/cmd/kwil-cli/cmds/utils/parse.go
+++ b/cmd/kwil-cli/cmds/utils/parse.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/kwilteam/kwil-db/cmd/common/display"
 	"github.com/kwilteam/kwil-db/core/types"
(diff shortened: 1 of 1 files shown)


jchappelow (approved):
LGTM.  I imagine this would be quite useful to see what's really going on (assuming you're cli is build with the same version of the parser as the node, which is only solvable with an RPC to do this).
Only comment is that the `Errors` might be difficult to understand unless they are prefixed/decorrated with what table/action/procedure/etc caused it.

brennanjl:
> Only comment is that the Errors might be difficult to understand unless they are prefixed/decorrated with what table/action/procedure/etc caused it.

Ah yeah that's a good point. I'm not too concerned about it for now, so since I have already merged it I'll wait. Errors in generation would almost always be due to the procedure/sql itself being invalid.

PR #790, opened 2024-06-04
added tests for decimal

brennanjl:
This adds more tests for decimal and properly handles return types from `sum`, fixing a bug noted by Truflation https://github.com/truflation/tsn/pull/295.

It also fixes a variety of other small bugs, such as failure to catch runtime errors for calculations done on numeric array types, and not enforcing precision and scale on final results returned to the client.

### DIFF
--- a/parse/functions.go
+++ b/parse/functions.go
@@ -573,6 +573,12 @@ var (
 					return nil, fmt.Errorf("expected argument to be numeric, got %s", args[0].String())
 				}
 
+				// we check if it is an unknown type before the switch,
+				// as unknown will be true for all EqualsStrict checks
+				if args[0] == types.UnknownType {
+					return types.UnknownType, nil
+				}
+
 				var retType *types.DataType
 				switch {
 				case args[0].EqualsStrict(types.IntType):
@@ -582,8 +588,6 @@ var (
 					retType.Metadata[0] = 1000 // max precision
 				case args[0].EqualsStrict(types.Uint256Type):
 					retType = decimal1000.Copy()
-				case args[0].EqualsStrict(types.UnknownType):
-					retType = types.UnknownType.Copy()
 				default:
 					panic(fmt.Sprintf("unexpected numeric type: %s", retType.String()))
 				}
--- a/parse/functions.go
+++ b/parse/functions.go
@@ -563,15 +563,32 @@ var (
 		},
 		"sum": {
 			ValidateArgs: func(args []*types.DataType) (*types.DataType, error) {
+				// per https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-TABLE
+				// the result of sum will be made a decimal(1000, 0)
 				if len(args) != 1 {
 					return nil, wrapErrArgumentNumber(1, len(args))
 				}
 
-				if !args[0].EqualsStrict(types.IntType) {
-					return nil, wrapErrArgumentType(types.IntType, args[0])
+				if !args[0].IsNumeric() {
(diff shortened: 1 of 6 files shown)


parse/functions.go
```diff
@@ -563,15 +563,32 @@ var (
...
+				var retType *types.DataType
+				switch {
+				case args[0].EqualsStrict(types.IntType):
+					retType = decimal1000.Copy()
+				case args[0].Name == types.DecimalStr:
+					retType = args[0].Copy()
+					retType.Metadata[0] = 1000 // max precision
+				case args[0].EqualsStrict(types.Uint256Type):
+					retType = decimal1000.Copy()
+				case args[0].EqualsStrict(types.UnknownType):
```
Yaiba: Can you give an example what's a unknown numeric type?

parse/functions.go
```diff
@@ -653,6 +670,17 @@ func defaultFormat(name string) FormatFunc {
 	}
 }
 
+// decimal1000 is a decimal type with a precision of 1000.
+var decimal1000 *types.DataType
+
+func init() {
```
Yaiba, line 680: I feel we should always put `init` at the very top, but I couldn't find a guide for this. nbd

parse/functions.go
```diff
@@ -563,15 +563,32 @@ var (
...
 				}
 
-				return types.IntType, nil
+				var retType *types.DataType
+				switch {
+				case args[0].EqualsStrict(types.IntType):
+					retType = decimal1000.Copy()
+				case args[0].Name == types.DecimalStr:
+					retType = args[0].Copy()
+					retType.Metadata[0] = 1000 // max precision
```
jchappelow, line 582: Is there any need to modify scale? Error if not already 0?  I'm just guessing about the assumptions here.

parse/functions.go
```diff
@@ -563,15 +563,32 @@ var (
...
+				case args[0].EqualsStrict(types.Uint256Type):
+					retType = decimal1000.Copy()
+				case args[0].EqualsStrict(types.UnknownType):
+					retType = types.UnknownType.Copy()
```
jchappelow, line 586: `EqualsStrict` has this:

```go
	// if unknown, return true. unknown is a special case used
	// internally when type checking is disabled.
	if c.Name == unknownStr || other.Name == unknownStr {
		return true
	}
```

So this can be true with any type?

parse/functions.go
```diff
@@ -563,15 +563,32 @@ var (
...
+				if !args[0].IsNumeric() {
+					return nil, fmt.Errorf("expected argument to be numeric, got %s", args[0].String())
 				}
 
-				return types.IntType, nil
+				var retType *types.DataType
+				switch {
+				case args[0].EqualsStrict(types.IntType):
+					retType = decimal1000.Copy()
+				case args[0].Name == types.DecimalStr:
```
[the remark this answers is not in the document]
brennanjl: Exactly. Regardless of what the precision and scale is, if it is any sort of decimal we want to set it to precision 1000

parse/functions.go
brennanjl: Oh jeez, that's a good catch. Will fix this

parse/functions.go
brennanjl: Yeah I thought about it, but honestly I think it's more clear next to the decimal since it is what we are initializing.

parse/functions.go
brennanjl: unknown is a special case that can be used for any type. I refactored this part though, as noted by Jon's bug above.

Yaiba (approved):
LGTM

PR #791, opened 2024-06-04
added empty returns

brennanjl:
This PR fixes a bug where procedures that do not return anything cannot return early:

```
procedure do_something() public {
  if true {
     return;
   }

  // something else
}
```


Yaiba (approved):
LGTM

Issue #752, opened 2024-05-16
Problem: Kuneiform Wasm includes `database` declaration in `schema.name`


KwilLuke:
When i begin typing a database definition in kuneiform, the parser begins returning the `database` keyword as the schema name. This stops after the full `database` keyword is typed.

For example, if I start typing:

```javascript
databa
```

`schema` shows:

```
{name: 'datab', owner: '', extensions: null, tables: null, actions: null, …}
```

However, if i type the full `database` declaration:

```javascript
database
```

`schema` shows `null`.

## Ideal Behavior

Ideally, `schema.name` would not show the database keyword in the `name` field. It would only return the database name once it has been defined after the `database` definition.

Issue #751, opened 2024-05-16
Problem: Kuneiform WASM returns `null` for `schema` while column is being defined


KwilLuke:
When a column is being defined on a table, there is a strange behavior where the WASM `parseKuneiform` function returns `null` for `schema` when a column is being defined (or no column exists on the table). Example:

Step 1: If I have this schema, WASM parse fine:

```javascript
database mydb;

table posts {
    id int primary notnull,
    name text notnull,
    post_title text notnull,
    post_h text
}

action add_post ($id, $user, $title, $body) public {
    INSERT INTO posts VALUES ( $id , $user , $title , $body );
}
```

Wasm output:
```
{schema: {…}, schema_info: {…}}
```

Step 2: But if i begin adding another table, the entire `schema` field goes to `null`:

```javascript
database mydb;

table posts {
    id int primary notnull,
    name text notnull,
    post_title text notnull,
    post_h text
}

table users {}

action add_post ($id, $user, $title, $body) public {
    INSERT INTO posts VALUES ( $id , $user , $title , $body );
}
```

Wam output:

```
{schema: null, errors: Array(1)}
```

Step 3: If I fully define one column, wasm returns `schema` without problem:

```javascript
database mydb;

table posts {
    id int primary notnull,
    name text notnull,
    post_title text notnull,
    post_h text
}

table users {
    id int primary
}

action add_post ($id, $user, $title, $body) public {
    INSERT INTO posts VALUES ( $id , $user , $title , $body );
}

```

Wam output:

```
{schema: {…}, schema_info: {…}}
```

Step 4: Then, if i begin defining a second column, the `schema` goes back to `null`:

```javascript
database mydb;

table posts {
    id int primary notnull,
    name text notnull,
    post_title text notnull,
    post_h text
}

table users {
    id int primary,
    username
}

action add_post ($id, $user, $title, $body) public {
    INSERT INTO posts VALUES ( $id , $user , $title , $body );
}
```

Wasm output:

```
{schema: null, errors: Array(1)}
```

## Desired Behavior

In my opinion, when a table does not have any column defined (as in step 2) or has an incomplete column defined (as in step 4), `schema` should still contain:

1. The rest of the schema that can be parsed (other tables, actions, procedures)
2. The problematic table's name in `schema.table`, and any parseable columns
3. The error on the table in `schema.errors`.

[...]

Issue #738, opened 2024-05-14
Problem: Inconsistencies between procedures & foreign calls


KwilLuke:
One thing I just noticed:

The Kuneiform parser has a slight inconsistency in how it name returns from procedures versus foreign_calls. `procedures.return_types` versus `foreign_calls.returns`. Because these are both the same thing (`foreign_calls.returns` is just referencing the `procedures.return_types` on another schema), I think these should be consistent.

IMO, `procedures.return_types` and `foreign_calls.return_types` is good.

![Screen Shot 2024-05-14 at 6 38 31 PM](https://github.com/kwilteam/kwil-db/assets/105600466/05bd2e04-b191-40fe-9ea6-f9830e101639)

brennanjl:
Was this addressed

jchappelow:
I think it was good going to be with fffe109b3ada77f72e4a4ca48c8b3b042ea5e6d8b12e41d0bbdc3cadfe2d3239 but no more

brennanjl:
Will add it to https://github.com/kwilteam/kwil-db/pull/794

PR #794, opened 2024-06-05
Parse fixes

brennanjl:
This PR fixes several bugs:
- We were not returning positions when json encoding. We originally did this to prevent spam while debugging, but this broke the IDE. We now have a function that allows us to do it.
- Some positions were not being set while parsing. This was fixed, and I also added tests to ensure that positions were set while parsing.
- Extraneous outputs occurring after EOF were not causing parse errors. This created very unintuitive, non-erroring behavior if someone, for example, doubly declared `database a; ... database b;`.
- Resolves https://github.com/kwilteam/kwil-db/issues/751
- Resolves https://github.com/kwilteam/kwil-db/issues/752
- Resolves https://github.com/kwilteam/kwil-db/issues/738

It also adds better error messages to parse debug, as well as gives the option to output position info.

This PR is based on https://github.com/kwilteam/kwil-db/pull/783, so 783 should go in first.


brennanjl:
@KwilLuke notice that this makes a minor change to the schema returned via JSON RPC

chappjc:
Will make some revisions and run tests

Yaiba (approved):
LGTM

PR #797, opened 2024-06-05
pg,kwild: smoother handling of repl conn hangups, determinism guarantees 

jchappelow:
This addresses some UX issues around the DB connections.  When a connection dies unexpectedly (e.g. postgresql shutdown or connection broken), `kwild` would previously not find out until something on the consensus thread tried to use the connection.  It now breaks cleanly and immediately.  This makes it much more clear what failed and when.

Related to the cryptic errors seen with DB connection issues is a key timeout setting, `wal_sender_timeout`, that should be disabled.  Our `kwildb/postgres:16.2-1` image disable it, but the 16.1 image did not, and a system install does not by default.  `kwild` now checks to ensure that it is disabled.

Several other `postgres` settings that could affect determinism are also verified.  These are already the default settings for the most part.  We are now just verifying that they are set as expected to ensure determinism of execution.

Also, `kwild` was double printing errors, and needlessly showing CLI usage info for application errors that were unrelated to CLI flag/config/etc.  This is resolved.

Finally, some very silly error messages from cometbft that we typically see on shutdown are now demoted to warnings so we don't have a distracting stack dump printed.

### DIFF
--- a/internal/sql/pg/system.go
+++ b/internal/sql/pg/system.go
@@ -67,7 +67,20 @@ func pgVersion(ctx context.Context, conn *pgx.Conn) (ver string, verNum uint32, 
 
 type settingValidFn func(val string) error
 
-func wantMinIntFn(wantMin int64) settingValidFn {
+func wantIntFn(want int64) settingValidFn { //nolint:unused
+	return func(val string) error {
+		num, err := strconv.ParseInt(val, 10, 64)
+		if err != nil {
+			return err
+		}
+		if num != want {
+			return fmt.Errorf("require %d, but setting is %d", want, num)
+		}
+		return nil
+	}
+}
+
+func wantMinIntFn(wantMin int64) settingValidFn { //nolint:unused
 	return func(val string) error {
 		num, err := strconv.ParseInt(val, 10, 64)
 		if err != nil {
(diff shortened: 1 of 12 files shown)


internal/sql/pg/system.go
```diff
@@ -110,10 +166,29 @@ var settingValidations = map[string]settingValidFn{
...
+	// the effect of allowing this kind of clean up.
+	"idle_in_transaction_timeout": wantIntFn(0), // disable disable idle transaction timeout for now
+
+	// Behavior related settings that must be set properly for determinism
+	// https://www.postgresql.org/docs/16/runtime-config-compatible.html
+	// The following are the documented defaults, which we enforce.
+	"array_nulls":                 wantOnFn(true),                // recognize NULL in array parser, false is for pre-8.2 compat
+	"standard_conforming_strings": wantOnFn(true),                // backslashes (\) in string literals are treated literally -- only escape syntax (E'...') is still usable if needed
+	"transform_null_equals":       wantOnFn(false),               // do not treat "expr=NULL" as "expr IS NULL"
+	"backslash_quote":             wantStringFn("safe_encoding"), // reject escaped single quotes like \', require standard ''
```
jchappelow, line 181: Just to clarify, these are the default postgresql behaviors that we are designing around.  These checks are here to ensure postgres is running with the expected settings.

charithabandi (dismissed):
lgtm!

jchappelow:
> lgtm!

adding docs to some of the new setting validation functions

brennanjl (dismissed):
lgtm

charithabandi (approved):

brennanjl:
@jchappelow will let you merge in case there is anything else you want to add

PR #798, opened 2024-06-06
re-enable cli tests

brennanjl:
Noticed that CLI integration tests were not being run, and that the flag was not set. Not sure when this happened.


Yaiba (approved):
lgtm

PR #799, opened 2024-06-07
fix: datatype pointer bug, access bugs

brennanjl:
This PR fixes several bugs, and adds tests for them.

Bugs:
- Data type metadata was not being copied properly https://github.com/truflation/tsn/issues/302
- Owner was not properly being enforced on external procedure calls
- Owner was not properly being enforced on foreign procedure calls. This was broken by a previous change that was made.
- Calling a non-view procedure outside of a blockchain tx that does not mutate state would not fail.

### DIFF
--- a/core/types/schema.go
+++ b/core/types/schema.go
@@ -1099,7 +1099,7 @@ type DataType struct {
 	// IsArray is true if the type is an array.
 	IsArray bool `json:"is_array"`
 	// Metadata is the metadata of the type.
-	Metadata *[2]uint16 `json:"metadata"`
+	Metadata [2]uint16 `json:"metadata"`
 }
 
 // String returns the string representation of the type.
@@ -1121,6 +1121,8 @@ func (c *DataType) String() string {
 	return str.String()
 }
 
+var ZeroMetadata = [2]uint16{}
+
 // PGString returns the string representation of the type in Postgres.
 func (c *DataType) PGString() (string, error) {
 	var scalar string
@@ -1138,7 +1140,7 @@ func (c *DataType) PGString() (string, error) {
 	case uint256Str:
 		scalar = "UINT256"
 	case DecimalStr:
-		if c.Metadata == nil {
+		if c.Metadata == ZeroMetadata {
 			return "", fmt.Errorf("decimal type must have metadata")
 		}
 
@@ -1162,13 +1164,13 @@ func (c *DataType) Clean() error {
 	c.Name = strings.ToLower(c.Name)
 	switch c.Name {
 	case intStr, textStr, boolStr, blobStr, uuidStr, uint256Str: // ok
-		if c.Metadata != nil {
+		if c.Metadata != ZeroMetadata {
 			return fmt.Errorf("type %s cannot have metadata", c.Name)
 		}
 
 		return nil
 	case DecimalStr:
-		if c.Metadata == nil {
+		if c.Metadata == ZeroMetadata {
 			return fmt.Errorf("decimal type must have metadata")
 		}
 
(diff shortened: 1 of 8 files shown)


core/types/schema.go
```diff
@@ -1195,11 +1195,16 @@ func (c *DataType) Clean() error {
...
-		Name:     c.Name,
-		IsArray:  c.IsArray,
-		Metadata: c.Metadata,
+	d := &DataType{
+		Name:    c.Name,
+		IsArray: c.IsArray,
 	}
+
+	if c.Metadata != nil {
+		d.Metadata = &[2]uint16{c.Metadata[0], c.Metadata[1]}
```
jchappelow, line 1204: It's a pointer because it's optional, but is there a sensible way to keep it a regular `[2]byte` with tests like `c.Metadata == zeroMetaData` or something?

internal/engine/execution/dataset.go
```diff
@@ -66,11 +78,11 @@ func (d *baseDataset) Call(caller *precompiles.ProcedureContext, app *common.App
 	// otherwise, it is an action
 	act, ok := d.actions[method]
```
Yaiba, line 79: action should also have `ownerOnly` right?

core/types/schema.go
brennanjl: We could do that. There are quite a few places where we do `datatype.Metadata == nil`, I guess that would just be `datatype.Metadata == [2]uint26{}`?

jchappelow: I can't tell which would be more cumbersome or error prone at a glance.  Whatever feels best to you.

But we'd probably do:

```go
var zeroMetaData [2]uint16 // global

...

if c.Metadata == zeroMetaData {

```

jchappelow: The benefit of an actual array field is that it deep copies automatically, and arrays are `comparable`, so you can do simple stuff like `==` there.

If it's optional, then it can be awkward.  Again, totally your call.  Just something to consider.

internal/engine/execution/procedure.go
Yaiba, line 1: L177 is duplicated with L102

Yaiba: or not?

core/types/schema.go
brennanjl: Yeah I can make the change

internal/engine/execution/dataset.go
Yaiba: ok, found it

internal/engine/execution/procedure.go
brennanjl: good point

internal/engine/execution/dataset.go
Yaiba: can procedure use instruction as well?

Yaiba: gotcha

internal/engine/execution/procedure.go
```diff
@@ -172,18 +172,6 @@ func (p *preparedAction) call(scope *precompiles.ProcedureContext, global *Globa
...
 	}
 
-	// if procedure does not have view tag, then it can mutate state
-	// this means that we must have a readwrite connection
-	if !p.view {
-		tx, ok := db.(sql.AccessModer)
-		if !ok {
-			return errors.New("DB does not provide access mode needed for mutative action")
-		}
-		if tx.AccessMode() != sql.ReadWrite {
```
jchappelow, line 182: Just to confirm, this was redundant with the special instruction appended when preparing the action?

jchappelow (approved):

PR #802, opened 2024-06-07
prep parse/v0.2.0

jchappelow:
This prepares for `parse/v0.2.0`:
- bump go.mod require's to use the just-tagged `core/v0.2.0`
- remove `parse` module's replace for `../core`

Doing this to get the `parse/v0.2.0` tag on `main`.  Then we'll branch `release-v0.8` where we'll remove the main module's replaces and update its `require`s.

Back on main (after the branch) we'll add the replace back into `parse/go.mod`


brennanjl (approved):

Yaiba (approved):
LGTM, we also need to bump the self report version

PR #808, opened 2024-06-10
fix parse bugs

brennanjl:
This fixes 3 parse errors that @KwilLuke found. Probably good for this to get backported, however it's not critical, as this simply fixes issues with errors output to the CLI and IDE.

The bugs are:
1. Parse error positions were being omitted if they occurred at the 0 line/column position. This was due to an `omitempty` tag that is no longer used anyways
2. Parse error types were not being included in the marshalled result.
3. Tables lacking primary keys were correctly identified as having errors, but the errors were not thrown in the parser, creating different error handling than other errors.

EDIT:

Also includes fixes for:
- `error` function not working in `SELECT CASE`
- The engine returning int literals as int32 instead of int64 (which all ints in kwil are assumed to be int64)

### DIFF
--- a/internal/sql/pg/query.go
+++ b/internal/sql/pg/query.go
@@ -337,6 +337,26 @@ func decodeFromPGType(vals ...any) ([]any, error) {
 		switch v := v.(type) {
 		default:
 			return v, nil
+
+		// we need to handle all ints as int64 since Kwil treats all
+		// ints as int64, but an integer literal in postgres can get
+		// returned as an int32
+		case int:
+			return int64(v), nil
+		case int8:
+			return int64(v), nil
+		case int16:
+			return int64(v), nil
+		case int32:
+			return int64(v), nil
+		case int64:
+			return v, nil
+		case uint:
+			return int64(v), nil
+		case uint16:
+			return int64(v), nil
+		case uint32:
+			return int64(v), nil
 		case pgtype.UUID:
 			u := types.UUID(v.Bytes)
 			return &u, nil
--- a/parse/analyze.go
+++ b/parse/analyze.go
@@ -136,8 +136,9 @@ type sqlContext struct {
 	ctes []*Relation
 	// outerScope is the scope of the outer query.
 	outerScope *sqlContext
-	// isAction is true if the visitor is analyzing SQL within an action.
-	isAction bool
+	// isInlineAction is true if the visitor is analyzing a SQL expression within an in-line
+	// statement in an action
+	isInlineAction bool
 	// inConflict is true if we are in an ON CONFLICT clause
 	inConflict bool
 	// targetTable is the name (or alias) of the table being inserted, updated, or deleted to/from.
@@ -335,7 +336,7 @@ func (c *sqlContext) scope() {
 		joinedTables:    make(map[string]*types.Table),
 		// we do not need to copy ctes since they are not ever modified.
 		targetTable:       c.targetTable,
-		isAction:          c.isAction,
+		isInlineAction:    c.isInlineAction,
 		inConflict:        c.inConflict,
 		inSelect:          c.inSelect,
 		hasAnonymousTable: c.hasAnonymousTable,
@@ -443,6 +444,18 @@ func (s *sqlAnalyzer) expectedNumeric(node Node, t *types.DataType) {
 // but it returns something else. It will attempt to read the actual type and create an error
 // message that is helpful for the end user.
 func (s *sqlAnalyzer) expressionTypeErr(e Expression) *types.DataType {
+
+	// prefixMsg is a function used to attempt to infer more information about
+	// the error. expressionTypeErr is typically triggered when someone uses a function/procedure
+	// with an incompatible return type. prefixMsg will attempt to get the name of the function/procedure
+	prefixMsg := func() string {
+		msg := "expression"
+		if call, ok := e.(ExpressionCall); ok {
+			msg = fmt.Sprintf(`function/procedure "%s"`, call.FunctionName())
+		}
+		return msg
+	}
+
 	switch v := e.Accept(s).(type) {
 	case *types.DataType:
 		// if it is a basic expression returning a scalar (e.g. "'hello'" or "abs(-1)"),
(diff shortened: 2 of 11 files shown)


jchappelow (approved):
Lgtm, but Luke can confirm.

Yaiba (dismissed):
LGTM

KwilLuke:
This resolves the issue I found, but two additional things should be added (I have already mentioned these to Brennan, but I am adding them here for record):
1. Returning the error type (e.g. `syntax`) is `parser_errors`. This happened before, but for some reason is no longer returned.
2. There is an unhandled error when opening a new procedure/action block. Unhandled means it is thrown in the `error` field of the `parseKuneiform` function rather than the `json.parser_errors` field where known errors are thrown. E.g., if you parse:

```
database mydb;

table my_table {
  id int primary notnull,
  col1 text notnull,
  col2 uuid notnull,
  col3 blob notnull,
  col4 bool notnull,
  col5 uint256 notnull,
  col6 decimal(4,1) notnull
}

action add_recor ($id) public {
 INSERT INTO my_table (id)
 VALUES ($id);
}

procedure add_proc ($id int) public {
  
}
```

The parser throws this error:

```
error: "parsing failed: runtime error: slice bounds out of range [316:312]\n\ngoroutine 6 [running]:\ngithub.com/kwilteam/kwil-db/parse.setupParser.func1({0x6b5e0, 0x1918228})\n\tgithub.com/kwilteam/kwil-db/parse/parse.go:433 +0x10\ngithub.com/kwilteam/kwil-db/parse.ParseSchemaWithoutValidation.func1()\n\tgithub.com/kwilteam/kwil-db/parse/parse.go:117 +0x5\npanic({0x6b5e0, 0x1918228})\n\truntime/panic.go:770 +0x1a\ngithub.com/antlr4-go/antlr/v4.(*InputStream).GetText(...)\n\tgithub.com/antlr4-go/antlr/v4@v4.13.0/input_stream.go:133\ngithub.com/kwilteam/kwil-db/parse.(*schemaVisitor).VisitProcedure_declaration(0x1b12b10, 0x1915a80)\n\tgithub.com/kwilteam/kwil-db/parse/antlr.go:776 +0xec\ngithub.com/kwilteam/kwil-db/parse/gen.(*Procedure_declarationContext).Accept(0x1915a80, {0xf2cd8, 0x1b12b10})\n\tgithub.com/kwilteam/kwil-db/parse/gen/kuneiform_parser.go:6774 +0x9\ngithub.com/kwilteam/kwil-db/parse.(*schemaVisitor).VisitSchema(0x1b12b10, 0x1914780)\n\tgithub.com/kwilteam/kwil-db/parse/antlr.go:342 +0xad\ngithub.com/kwilteam/kwil-db/parse/gen.(*SchemaContext).Accept(0x1914780, {0xf2cd8, 0x1b12b10})\n\tgithub.com/kwilteam/kwil-db/parse/gen/kuneiform_parser.go:2955 +0x9\ngithub.com/kwilteam/kwil-db/parse.(*schemaVisitor).VisitSchema_entry(0x1b12b10, 0x1914700)\n\tgithub.com/kwilteam/kwil-db/parse/antlr.go:57 +0x1f\ngithub.com/kwilteam/kwil-db/parse/gen.(*Schema_entryContext).Accept(0x1914700, {0xf2cd8, 0x1b12b10})\n\tgithub.com/kwilteam/kwil-db/parse/gen/kuneiform_parser.go:949 +0x9\ngithub.com/kwilteam/kwil-db/parse.ParseSchemaWithoutValidation({0x18f4780, 0x13d, 0x13d})\n\tgithub.com/kwilteam/kwil-db/parse/parse.go:123 +0x1c\ngithub.com/kwilteam/kwil-db/parse.ParseAndValidate({0x18f4780, 0x13d, 0x13d})\n\tgithub.com/kwilteam/kwil-db/parse/parse.go:56 +0x2\nmain.parseAndMarshal({0x18f4780, 0x13d})\n\t./wasm.go:15 +0x2\nmain.parseWrapper.func1({{}, 0x7ff8000100000005, 0x1b99870}, {0x18e90c0, 0x1, 0x1})\n\t./wasm.go:43 +0x5\nsyscall/js.handleEvent()\n\tsyscall/js/func.go:100 +0x23\n\n"
```

parse/analyze.go
```diff
@@ -136,8 +136,9 @@ type sqlContext struct {
 	ctes []*Relation
 	// outerScope is the scope of the outer query.
 	outerScope *sqlContext
-	// isAction is true if the visitor is analyzing SQL within an action.
-	isAction bool
+	// isInlineAction is true if the visitor is analyzing a SQL expression within an in-line
+	// statement in an action
+	isInlineAction bool
```
jchappelow, line 141: Ah, so this was correctly false.  But where is it actually set in our Go code for in-line SQL? `scope()` just clones, but I don't see where it would actually be set to `true` anywhere.

brennanjl: Whoops. Just fixed this, and added a test to confirm we receive the expected error when this occurs.

internal/sql/pg/query.go
```diff
@@ -337,6 +337,26 @@ func decodeFromPGType(vals ...any) ([]any, error) {
...
+			return int64(v), nil
+		case int16:
+			return int64(v), nil
+		case int32:
+			return int64(v), nil
+		case int64:
+			return v, nil
+		case uint:
+			return int64(v), nil
+		case uint16:
```
jchappelow, line 356: Did you want `uint8` here as well?  (although I don't think postgres+pgx is ever going to use an unsigned type, but I could be wrong).

Also, I'm wondering if this forced int64 behavior is to opinionated for this level of the system.  e.g. is this an engine requirement?  Should it be forced on all consumers of pg including internal stores, etc?  In engine, we have an `cleanseIntValue` function that does very similar.  Perhaps we should be applying this type issue up there somewhere.

brennanjl: I intentionally did not include uint8, in case it is reading bytes

brennanjl: > Also, I'm wondering if this forced int64 behavior is to opinionated for this level of the system

Very much could be, however the SQL package is already familiar with the other Kwil types, and does not support other types like float, which is already pretty Kwil specific. But if you think it is too opinionated, I am definitely ok to remove

jchappelow: OK.  Regardless of "the right place" for this.  It's hard to see why it would be a problem for any lesser integer types to return to `query` consumers as `int64`, so OK. :heavy_check_mark:

jchappelow: > and does not support other types like float, which is already pretty Kwil specific

Well, it returns float unconverted if you're using `query`.  But NBD.  Nothing is losing precision.

jchappelow (approved):

PR #813, opened 2024-06-11
added genesis-state and extension flags

brennanjl:
PR fixes two issues I found regarding kwild flags.

1. `genesis-state` was not configurable via flags
2. extensions were not configurable via flags.

Adding genesis state was very straight forward. Extensions were tricky, because users should be able to specify any arbitrary extension name and key. To achieve this, I used the `--` delimiter from the [POSIX guidelines](<https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02>) (Guideline 10) to allow users to specify additional flag arguments for `kwild`. It verifies that these are correct flags for extensions:

```
kwild --autogen -- --my_ext.key=value --my_ext.key2 value2
```

### DIFF
--- a/cmd/kwild/root/root.go
+++ b/cmd/kwild/root/root.go
@@ -48,7 +48,7 @@ func RootCmd() *cobra.Command {
 			fmt.Printf("kwild version %v (Go version %s)\n", version.KwilVersion, runtime.Version())
 
 			// args can be passed to kwild in the form of extension flags.
-			// These are delimited using a double dash, e.g. `kwild -- --extension1.flag1=value1 --extension2.flag2=value2`
+			// These are delimited using a double dash, e.g. `kwild -- --extension.extension1.flag1=value1 --extension.extension2.flag2=value2`
 			// This is in-line with guideline 10 of the POSIX guidelines: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02
 			extensionConfig, err := parseExtensionFlags(args)
 			if err != nil {
@@ -131,13 +131,13 @@ func parseExtensionFlags(args []string) (map[string]map[string]string, error) {
 		}
 		// split the flag into the extension name and the flag name
 		// we intentionally do not use SplitN because we want to verify
-		// there are exactly 2 parts.
+		// there are exactly 3 parts.
 		parts := strings.Split(args[i], ".")
-		if len(parts) != 2 {
+		if len(parts) != 3 {
 			return nil, fmt.Errorf("invalid extension flag %q", args[i])
 		}
 
-		extName := strings.TrimPrefix(parts[0], "--")
+		extName := strings.TrimPrefix(parts[1], "--")
 
 		// get the extension map for the extension name.
 		// if it doesn't exist, create it.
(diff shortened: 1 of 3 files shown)


cmd/kwild/root/root.go
```diff
@@ -112,6 +121,57 @@ func RootCmd() *cobra.Command {
...
+			ext = make(map[string]string)
+			exts[extName] = ext
+		}
+
+		// we now need to get the flag value. Flags can be passed
+		// as either "--extension.extname.flagname value" or
+		// "--extension.extname.flagname=value"
+		if strings.Contains(parts[1], "=") {
+			// flag value is in the same argument
+			val := strings.SplitN(parts[1], "=", 2)
```
Yaiba, line 155: what about `--ext.name=`?  Since `--ext.name --ext.xx=2` will error, to make it consistent, I think here should check `val[1]` has value.

cmd/kwild/root/root.go
```diff
@@ -41,12 +42,20 @@ func RootCmd() *cobra.Command {
 		Short:             "kwild node and rpc server",
 		Long:              "kwild: the Kwil blockchain node and RPC server",
```
Yaiba, line 43: I think we'll need help message on optional flags

Yaiba (dismissed):
LGTM. 
I'm not sure whether the flags follows the `--` should be using `-` or `--`, I have seen single dash flags before, but not double dashes, an example is `taskfile`.

cmd/kwild/root/root.go
```diff
@@ -112,6 +121,57 @@ func RootCmd() *cobra.Command {
...
+		// if it doesn't exist, create it.
+		ext, ok := exts[extName]
+		if !ok {
+			ext = make(map[string]string)
+			exts[extName] = ext
+		}
+
+		// we now need to get the flag value. Flags can be passed
+		// as either "--extension.extname.flagname value" or
+		// "--extension.extname.flagname=value"
```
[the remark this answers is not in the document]
brennanjl: Oh right, old comment

brennanjl:
> I'm not sure whether the flags follows the -- should be using - or --, I have seen single dash flags before, but not double dashes, an example is taskfile.

Some quick google searches shows both. e.g.: https://medium.com/fhinkel/the-curious-case-of-double-dashes-b5e7711698f#:~:text=POSIX.&text=The%20first%20%2D%2D%20argument%20that,with%20the%20'%2D'%20character.

cmd/kwild/root/root.go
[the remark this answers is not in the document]
Yaiba: hmm actually I feel 3 parts seems more clear and extensible, the first part is to determine the config section.
Unless we're sure only extension configure can be configure after `--`

cmd/kwild/root/root.go
brennanjl: This matches Cobra's parsing

Yaiba: yeah you're right

cmd/kwild/root/root.go
```diff
@@ -112,6 +121,57 @@ func RootCmd() *cobra.Command {
...
+		}
+		// split the flag into the extension name and the flag name
+		// we intentionally do not use SplitN because we want to verify
+		// there are exactly 3 parts.
+		parts := strings.Split(args[i], ".")
+		if len(parts) != 3 {
+			return nil, fmt.Errorf("invalid extension flag %q", args[i])
+		}
+
+		extName := strings.TrimPrefix(parts[1], "--extension.")
```
Yaiba, line 140: `parts[1]` doesn't need formatting

Yaiba: btw, I think `parseExtensionFlags` should check `parts[0]` equals `--extension`

brennanjl: ok

brennanjl: I now check that the arg starts with `--extension.`

cmd/kwild/root/root.go
```diff
@@ -31,6 +32,26 @@ import (
 	"github.com/spf13/viper"
 )
 
+var long = `kwild is the Kwil blockchain node and RPC server.
```
Yaiba, line 35: 👍

Yaiba (dismissed):

Yaiba (approved):

jchappelow (approved):

PR #814, opened 2024-06-11
State sync fixes

jchappelow:
This gets state sync working and fixes some tangential issues.

- add logging around state sync code, particularly with trusted rpc provider and snapshot retrieval and apply
- Fix the default and example config files using `snapshot_heights` instead of `recurring_height`
- Reduce the chunk size limit to < 16,000,000 bytes (previously was set at 16 MiB, not MB, plus lacked overhead space)
- With statesync restore error handling, return `ResponseApplySnapshotChunk_REJECT_SNAPSHOT` to cometbft as a fallback when the error is not typed.  `UNKNOWN` caused cometbft to fail rather than continue with another snapshot candidate.

With this and https://github.com/kwilteam/kwil-db/pull/812 we were able to get longhorn running and state sync on the sync node working.

### DIFF
--- a/cmd/kwild/config/test_data/config.toml
+++ b/cmd/kwild/config/test_data/config.toml
@@ -96,14 +96,15 @@ hostname = ""
 ###                Snapshot store Config Options                    ###
 #######################################################################
 [app.snapshots]
+
 # Toggle to enable snapshot store
-# This would snapshot the application state at every snapshot_heights blocks
+# This would snapshot the application state at every recurring_height blocks
 # and keep max_snapshots number of snapshots in the snapshot_dir
 # Application state includes the databases deployed, accounts and the Validators db
 enabled = false
 
 # The height at which the snapshot is taken
-snapshot_heights = 100000
+recurring_height = 100000
 
 # Maximum number of snapshots to be kept in the snapshot_dir.
 # If max limit is reached, the oldest would be deleted and replaced by the latest snapshot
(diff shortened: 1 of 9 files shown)


internal/statesync/store.go
```diff
@@ -97,6 +97,7 @@ func NewSnapshotStore(cfg *SnapshotConfig, dbCfg *DBConfig, logger log.Logger) (
 
 // IsSnapshotDue checks if a snapshot is due at the given height.
 func (s *SnapshotStore) IsSnapshotDue(height uint64) bool {
+	s.log.Infof("due? height %d / recurring %d", height, s.cfg.RecurringHeight)
```
charithabandi, line 100: remove this log. too noisy

cmd/kwild/config/test_data/config.toml
```diff
@@ -96,14 +96,15 @@ hostname = ""
...
 # Toggle to enable snapshot store
-# This would snapshot the application state at every snapshot_heights blocks
+# This would snapshot the application state at every recurring_height blocks
 # and keep max_snapshots number of snapshots in the snapshot_dir
 # Application state includes the databases deployed, accounts and the Validators db
 enabled = false
 
 # The height at which the snapshot is taken
-snapshot_heights = 100000
+recurring_height = 100000
```
brennanjl, line 107: Do we need to change the CLI flags too?

Also, could either you or @charithabandi add a pr for docs?

jchappelow: It was already changed everywhere else in this repo.  Not sure about docs though

internal/statesync/store.go
jchappelow: done.

cmd/kwild/config/test_data/config.toml
jchappelow: Just checked and docs repo has correct setting.  These two toml files where the only place `snapshot_heights` was left.

brennanjl (approved):
lgtm but I'll wait for @charithabandi to look too since she knows more

PR #815, opened 2024-06-11
fix error in resolve funcs

brennanjl:
While working through the Powerpod integration, I spotted a bug in our resolution code. If a resolution fails, it will halt consensus. This should not be the case, since resolutions are made to target databases, which can have arbitrary user-defined logic.

### DIFF
--- a/internal/txapp/txapp.go
+++ b/internal/txapp/txapp.go
@@ -631,7 +631,11 @@ func (r *TxApp) processVotes(ctx context.Context, blockHeight int64) error {
 			if err2 != nil {
 				return fmt.Errorf("error rolling back transaction: %s, error: %s", err.Error(), err2.Error())
 			}
-			return err
+
+			// if the resolveFunc fails, we should still continue on, since it simply means
+			// some business logic failed in a deployed schema.
+			r.log.Error("error resolving resolution", log.String("type", resolveFunc.Resolution.Type), log.String("id", resolveFunc.Resolution.ID.String()), log.Error(err))
+			continue
 		}
 
 		err = tx.Commit(ctx)


jchappelow (approved):
Seems fine.  If it's not actually an issue with "arbitrary user-defined logic" but some other fatal error, instead something local to the node, would there least be an app hash mismatch?  Or would that be go unnoticed for some arbitrary time?

brennanjl:
> Seems fine. If it's not actually an issue with "arbitrary user-defined logic" but some other fatal error, instead something local to the node, would there least be an app hash mismatch? Or would that be go unnoticed for some arbitrary time?

There are only two cases I could see where the node does something node-specific, resulting in a different result:
1. The node loses connection to Postgres, in which case we will find out immediately after the failed resolution when we try to roll the tx back.
2. The author creates something with a non-deterministic, in which case there likely would be an apphash mismatch.

Since we do not include user balances and validator power in app-hashes, there is a risk that a non-deterministic change there might not result in an app-hash mismatch until later, but this seems more like a Kwil issue to fix.

brennanjl:
I see what you're saying. IMO, it still seems like this could be solved with improved apphash tracking. Right now, only user schemas affect apphash. If the validator schema affected apphash too, then the node that gets an unexpected result / error from changing validator power would recognize their state diverged in the next block, right?

jchappelow:
Hypothetically.  I think it's handled more easily by hard failing where the extension author does not anticipate an error.  You cannot roll back an already-committed an app state change.  If you commit something erroneous (and we can know it is), then you diverge permanently and have to resync if you want to get back on the network.

I think the API is simply lacking for the resolution resolution function, or the ResolveFunc should log (not the app) when it's not fatal, and return an error when it's fatal.

jchappelow:
In any case, if I were writing an extension, I would look at the docs for the field:

```
	// ResolveFunc is a function that is called once a resolution has
	// received a required number of votes, as defined by the
	// ConfirmationThreshold. It is given a readwrite database
	// connection and the information for the resolution that has
	// been confirmed. All nodes will call this function as a part of
	// block execution. It is therefore expected that the function is
	// deterministic, regardless of a node's local configuration.
	ResolveFunc func(ctx context.Context, app *common.App, resolution *Resolution) error
```

When should I return an error.  How does the app handle an error?  How should I deal with internal errors?

The contract for that API could use some clarification.

brennanjl:
> Agreed

brennanjl:
> The contract for that API could use some clarification.

Agreed

brennanjl:
> I think the API is simply lacking for the resolution function, or the ResolveFunc should log (not the app) when it's not fatal, and return an error when it's fatal.

Unfortunately, this is hard to delineate right now. For example, if procedure execution failed due to an `error()` function, and if procedure execution failed due to a lost connection to Postgres, the error would be returned in the same place. Maybe something we can solve in the future.

PR #816, opened 2024-06-11
Backport fixes

brennanjl:
Backports:
- https://github.com/kwilteam/kwil-db/pull/815

Will also wait for https://github.com/kwilteam/kwil-db/pull/813 and https://github.com/kwilteam/kwil-db/pull/814

We should make sure to rebase and merge


jchappelow (approved):
This looks like everything.
Since the PR has multiple back ports, let's be sure to not squash so that we have a correspondence on `main` (plus it's easier to bisect and ultimately revert a subset of changes if need be)

PR #823, opened 2024-06-11
[WIP] docs: add release notes for v0.8

jchappelow:
Will get done with this tonight.  Still working on it.

Will render [here](https://github.com/jchappelow/kwil-db/blob/relnotes-0.8/docs/release-notes/v0.8.1.md), but again, not done at all.

I'm drafting in `docs/release-notes`, but I don't think we *have to* check them in.  It probably a good idea, but it can be awkward when release notes on GitHub diverge from what's checked in, or when the notes end up only on `main` because they frequently come after a release tag (they aren't code after all).


jchappelow:
Might consider this later.

PR #826, opened 2024-06-13
fix resolution body index

brennanjl:
@outerlook ran into an issue where Kwil indexes a column that we should not be indexing. This places an unnecessary constraint on the amount of data that can fit in an index.

### DIFF
--- a/internal/voting/sql.go
+++ b/internal/voting/sql.go
@@ -43,7 +43,7 @@ const (
 		vote_body_proposer BYTEA, -- vote_body_proposer is the identifier of the node that supplied the vote body
 		expiration INT8 NOT NULL, -- expiration is the blockheight at which the resolution expires
 		extra_vote_id BOOLEAN NOT NULL DEFAULT FALSE, -- If vote_body_proposer had sent VoteID before VoteBody, this is set to true
-		UNIQUE (id, body, type),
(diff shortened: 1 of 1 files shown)


brennanjl:
Since there aren't any long-run networks that rely on this mechanism anyways, I feel it is probably not too critical that we build an upgrade process for this, especially with the migrations I am adding for this next release.

jchappelow (approved):

jchappelow:
LG @charithabandi ?  Any considerations come to mind?

charithabandi (commented):
Looks good.  I am not sure of the best practices in updating the internal schema here. Shouldn't we introduce new version and have corresponding upgrade procedures ? bytea to UUID changes were done without introducing new version which made it harder to backtrack. But as we have a different plan to consolidate upgrade changes, this fix is okay I guess.

charithabandi (approved):

brennanjl:
> Looks good. I am not sure of the best practices in updating the internal schema here. Shouldn't we introduce new version and have corresponding upgrade procedures ? bytea to UUID changes were done without introducing new version which made it harder to backtrack and migrate. But as we have a different plan to consolidate upgrade changes, this fix is okay I guess.

If we were going to backport, I'd say yes.

PR #828, opened 2024-06-13
Introduce genesis consensus params to configure max limits for votebo…

charithabandi:


cmd/kwild/config/genesis.go
```diff
@@ -69,76 +69,3 @@ func loadGenesisAndPrivateKey(autoGen bool, privKeyPath, rootDir string) (privKe
 	fmt.Printf("Generated new genesis file: %v\n", genFile)
 	return privKey, genesisCfg, nil
 }
-
```
charithabandi, line 72: Removing this, as we decided to use genesis app hash as a way to enforce genesis state through snapshots.

common/chain/chaincfg.go
```diff
@@ -116,6 +116,12 @@ type VoteParams struct {
 	// VoteExpiry is the number of blocks after which the resolution expires if
 	// consensus is not reached.
 	VoteExpiry int64 `json:"vote_expiry"`
+
+	// MaxVoteIDsPerTx is the maximum number of votes to include in a ValidatorVoteIDs transaction.
+	MaxVoteIDsPerTx int64 `json:"vote_id_limit"`
+
+	// MaxVoteBodiesPerTx is the maximum number of vote bodies to include in a ValidatorVoteBodies transaction.
+	MaxVoteBodiesPerTx int64 `json:"vote_body_limit"`
```
jchappelow, line 124: These don't look like genesis parameters to me, rather config.toml node settings.

One affects transaction authoring and broadcast.  The other affects block proposal creation.  The other nodes need not be running with the same settings (another validator has no logic to reject a block if either of these values appears to be different when they inspect the proposal)

jchappelow: If we want to add logic to `ProcessProposal` or elsewhere to make these limits part of consensus, that's another thing to consider.  Presently these don't appear to be enforced by consensus.

brennanjl (dismissed):

common/chain/chaincfg.go
jchappelow: OK so both *can* be enforced by consensus with changes to `ProcessProposal`.

Just one question: Is one limit more important to enforce than the other?  For instance, which of these is easier for a rogue validator to abuse if the limit is not enforced?

jchappelow:
This now makes sense as a Genesis parameter with it part of consensus (not just the Genesis app hash).

If we want this on 0.8, we'll need to remove the affect on Genesis app hash before backporting.  It's really not need anyway.

common/chain/chaincfg.go
charithabandi: I can see both being the problem if not enforced correctly. But we have tx costs proportional to the vote sizes for both, which can limit the damage the validators can do on a gas costs enabled network. If gas costs are not enabled, these hard limits are kinda important

brennanjl:
> If we want this on 0.8, we'll need to remove the affect on genesis app hash before backporting. It's really not need anyway.

@charithabandi can you confirm you removed this? We do want this on v0.8

charithabandi:
> > If we want this on 0.8, we'll need to remove the affect on genesis app hash before backporting. It's really not need anyway.
> 
> @charithabandi can you confirm you removed this? We do want this on v0.8

@brennanjl  Yup, removed this. Gonna backport these changes to v0.8

jchappelow (dismissed):
Just two minor nits, but LGTM

jchappelow (approved):
Reapproving. LGTM.

PR #830, opened 2024-06-13
added persistence of consensus params

brennanjl:
This adds a separate consensus params struct that is used to persist Kwil specific consensus params. I am going to use this struct in the migrations PR, which will allow us to change these values.

As an aside, this PR was really really messy to make, somewhat unnecessarily. I think it would be way better if we managed the db and transaction creation in ABCI, rather than TxApp. It seems like we do a lot of hacking around this, and it has made the entire thing pretty confusing.


brennanjl:
closed in favor of https://github.com/kwilteam/kwil-db/pull/831

PR #831, opened 2024-06-14
Abci db

brennanjl:
This PR makes a substantial change to the repo by having ABCI control the creation and passing of database transactions, rather than TxApp. We had discussed making this change prior to the release, but decided against it since it is substantial. Overall, I think it removes a lot of the complexity of the relationship between abci and txapp.

This PR also establishes `NetworkParameters`, which are the consensus parameters that Kwil applications care about. It handles persistence of them after every block, so that they can be changed.

Finally, it uses both of these changes to establish a `BlockContext`, which is passed to functions that are being executed within a block. We ran into a small issue of this being missing with the [Streamr integration](<https://github.com/kwilteam/kwil-streamr/blob/2f3d833d80df0963b66a215acf89210a61c6928a/extensions/resolution/resolution.go#L51>), and then I ran into it again while working on migrations (which is how I got set down this rabbit hole in the first place).

This PR also adds a new method for creating DB transactions, `BeginDelayedReadTx`. One of the tradeoffs of the above mentioned changes is that database transactions are passed to some methods that may be able to simply read data that is cached and return it. To avoid unnecessary transaction creation, `BeginDelayedReadTx` returns a `sql.Tx`, but the underlying transaction is not actually created in Postgres until it is used for the first time, saving a round-trip in some cases. I considered simply adding this functionality to our existing `BeginReadTx` method, since I can't think of many cases where we wouldn't want this, but I held off for now.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -124,3 +124,28 @@ func (e *ExecutionData) Clean() error {
 	e.Procedure = strings.ToLower(e.Procedure)
 	return nil
 }
+
+// NetworkParameters are network level configurations that can be
+// evolved over the lifetime of a network.
+type NetworkParameters struct {
+	// MaxBlockSize is the maximum size of a block in bytes.
+	MaxBlockSize int64
+	// JoinExpiry is the number of blocks after which the validators
+	// join request expires if not approved.
+	JoinExpiry int64
+	// VoteExpiry is the default number of blocks after which the validators
+	// vote expires if not approved.
+	VoteExpiry int64
+	// DisabledGasCosts dictates whether gas costs are disabled.
+	DisabledGasCosts bool
+}
+
+// Copy returns a deep copy of the network parameters.
+func (n *NetworkParameters) Copy() *NetworkParameters {
+	return &NetworkParameters{
+		MaxBlockSize:     n.MaxBlockSize,
+		JoinExpiry:       n.JoinExpiry,
+		VoteExpiry:       n.VoteExpiry,
+		DisabledGasCosts: n.DisabledGasCosts,
+	}
+}
--- a/common/common.go
+++ b/common/common.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"strings"
 
-	"github.com/kwilteam/kwil-db/common/chain"
 	"github.com/kwilteam/kwil-db/common/sql"
 	"github.com/kwilteam/kwil-db/core/log"
 	"github.com/kwilteam/kwil-db/core/types"
@@ -41,9 +40,6 @@ type App struct {
 // router and extension implementations in extensions/consensus.
 type TxContext struct {
 	Ctx context.Context
-	// ConsensusParams holds network level parameters that can be evolved
-	// over the lifetime of a network.
-	ConsensusParams *chain.ConsensusParams
 	// BlockContext is the context of the current block.
 	BlockContext *BlockContext
 	// TxID is the ID of the current transaction.
--- a/common/common.go
+++ b/common/common.go
@@ -41,13 +41,11 @@ type App struct {
 // router and extension implementations in extensions/consensus.
 type TxContext struct {
 	Ctx context.Context
-	// BlockHeight gets the height of the current block.
-	BlockHeight int64
-	// Proposer gets the proposer public key of the current block.
-	Proposer []byte
 	// ConsensusParams holds network level parameters that can be evolved
 	// over the lifetime of a network.
 	ConsensusParams *chain.ConsensusParams
+	// BlockContext is the context of the current block.
+	BlockContext *BlockContext
 	// TxID is the ID of the current transaction.
 	TxID []byte
 }
--- a/common/context.go
+++ b/common/context.go
@@ -1,0 +1,20 @@
+package common
+
+// ChainContext provides context for all chain operations.
+type ChainContext struct {
+	// ChainID is the unique identifier for the chain.
+	ChainID string
+	// NetworkParams holds network level parameters that can be evolved
+	// over the lifetime of a network.
+	NetworkParameters *NetworkParameters
+}
+
+// BlockContext provides context for all block operations.
+type BlockContext struct {
+	// ChainContext contains information about the chain.
+	ChainContext *ChainContext
+	// Height gets the height of the current block.
+	Height int64
+	// Proposer gets the proposer public key of the current block.
+	Proposer []byte
+}
--- a/common/context.go
+++ b/common/context.go
@@ -1,6 +1,9 @@
 package common
 
 // ChainContext provides context for all chain operations.
+// Fields in ChainContext should never be mutated, except
+// NetworkParameters can be deterministically mutated as part
+// of block execution.
 type ChainContext struct {
 	// ChainID is the unique identifier for the chain.
 	ChainID string
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -454,10 +454,9 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		}
 
 		txRes := a.txApp.Execute(txapp.TxContext{
-			ConsensusParams: a.consensusParams,
-			Ctx:             ctx,
-			TxID:            tmhash.Sum(tx), // use cometbft TmHash to get the same hash as is indexed
-			BlockContext:    &blockCtx,
+			Ctx:          ctx,
+			TxID:         tmhash.Sum(tx), // use cometbft TmHash to get the same hash as is indexed
+			BlockContext: &blockCtx,
 		}, a.consensusTx, decoded)
 
 		abciRes := &abciTypes.ExecTxResult{}
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -1307,12 +1307,21 @@ func (a *AbciApp) SetReplayStatusChecker(fn func() bool) {
 	a.replayingBlocks = fn
 }
 
-// Close cleans up the application.
+// Close is used to end any active database transaction that may exist if the
+// application tries to shut down before closing the transaction with a call to
+// Commit. Neglecting to rollback such a transaction may prevent the DB
+// connection from being closed and released to the connection pool.
 func (a *AbciApp) Close() error {
 	if a.genesisTx != nil {
 		err := a.genesisTx.Rollback(context.Background())
 		if err != nil {
 			return fmt.Errorf("failed to rollback genesis transaction: %w", err)
+		}
+	}
+	if a.consensusTx != nil {
+		err := a.consensusTx.Rollback(context.Background())
+		if err != nil {
+			return fmt.Errorf("failed to rollback consensus transaction: %w", err)
 		}
 	}
 	return nil
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -123,7 +123,10 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
 		// height, these updates would already have been applied.
 	}
 
-	// we need to persist the genesis consensus params if they are not already
+	// if the network params have never been stored (which is the case for a fresh network),
+	// we need to persist them for the first time. If they are found, we need to update our consensus params
+	// with whatever the value is, since the consensus params here are read from the genesis file, and
+	// may have been altered if this is not a new network.
 	networkParams, err := meta.LoadParams(ctx, tx)
 	if err == meta.ErrParamsNotFound {
 		// we need to store the genesis network params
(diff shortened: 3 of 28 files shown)


jchappelow:
All looks sensible from a 5 minute turbo review.  More changes, or should I go ahead an look closer, @brennanjl ?

brennanjl:
> All looks sensible from a 5 minute turbo review. More changes, or should I go ahead an look closer, @brennanjl ?

No more changes if you're good to go

internal/abci/abci.go
```diff
@@ -304,12 +377,36 @@ func (a *AbciApp) CheckTx(ctx context.Context, incoming *abciTypes.RequestCheckT
 func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinalizeBlock) (*abciTypes.ResponseFinalizeBlock, error) {
 	logger := a.log.With(zap.String("stage", "ABCI FinalizeBlock"), zap.Int64("height", req.Height))
 
+	if a.genesisTx != nil {
+		// if we are in the first block, we use the genesisTx.
+		// This is to prevent a bug where a node crashing between InitChain and
+		// genesis is unable to recover.
```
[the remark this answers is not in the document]
brennanjl: This can happen on any block though, right?

brennanjl:
I believe this may have introduced a bug with shutdown. Let me dig in, hold off on merging

brennanjl:
Ok, all fixed. Shutting down during the genesis tx caused the issue, since it was never being committed or rolled back.

jchappelow:
Scanning through quick.

internal/txapp/txapp.go
```diff
@@ -29,8 +28,8 @@ import (
 )
 
 // NewTxApp creates a new router.
-func NewTxApp(ctx context.Context, db DB, engine common.Engine, signer *auth.Ed25519Signer,
-	events Rebroadcaster, snapshotter Snapshotter, chainParams *chain.GenesisConfig,
+func NewTxApp(ctx context.Context, db sql.Executor, engine common.Engine, signer *auth.Ed25519Signer,
```
jchappelow, line 31: Does not need `db` at all now, it seems.

internal/abci/abci.go
```diff
@@ -1067,3 +1297,22 @@ func (a *AbciApp) TxInMempool(txHash []byte) bool {
...
+
+// Close cleans up the application.
+func (a *AbciApp) Close() error {
+	if a.genesisTx != nil {
+		err := a.genesisTx.Rollback(context.Background())
+		if err != nil {
+			return fmt.Errorf("failed to rollback genesis transaction: %w", err)
+		}
+	}
+	return nil
```
jchappelow, line 1317: Here's the removed `Close` from txapp: https://github.com/kwilteam/kwil-db/pull/831/files#diff-29b2bc6069630247b4e88988ee7f8c5a0c4fa090867fc9d1e5d050c64da892d1L129-L138

We'll need to check for both tx fields to avoid a hang in all cases, not just genesis, I believe.  Can probably just grab that old method, docs and all.

internal/abci/interfaces.go
```diff
@@ -82,3 +67,15 @@ type ConsensusParams interface {
...
+// from within a transaction. A DB can create read transactions or the special
+// two-phase outer write transaction.
+type DB interface {
+	sql.OuterTxMaker
+	sql.ReadTxMaker
+	sql.SnapshotTxMaker
+	sql.Executor
+	sql.TxMaker
+	AutoCommit(auto bool)
+}
```
jchappelow, line 81:
```go
type DB interface {
	sql.OuterTxMaker
	sql.ReadTxMaker
	sql.SnapshotTxMaker
}
```

is all that's required by `AbciApp`.

Having the other methods available are likely to lead to misuse.

internal/abci/meta/meta.go
```diff
@@ -29,17 +35,30 @@ const (
...
+
+	getParams = `SELECT param_name, param_value FROM ` + chainSchemaName + `.consensus_params;`
 )
 
 func initTables(ctx context.Context, tx sql.DB) error {
 	_, err := tx.Execute(ctx, initChainTable)
+	if err != nil {
+		return err
+	}
 	return err
```
jchappelow, line 51: This function's error handing looks pretty odd now.  No conditional is required here, but if you want to keep an consistent approach, maybe change the second one to `return nil`, but personally it's too verbose.

internal/abci/meta/meta.go
```diff
@@ -80,15 +99,176 @@ func GetChainState(ctx context.Context, db sql.Executor) (int64, []byte, error)
 }
 
 // SetChainState will update the current height and app hash.
-func SetChainState(ctx context.Context, db sql.Executor, height int64, appHash []byte) error {
+func SetChainState(ctx context.Context, db sql.DB, height int64, appHash []byte) error {
```
jchappelow, line 102: Some of these `db` params could be tightened to `sql.TxMaker`, but nbd.

common/common.go
```diff
@@ -124,3 +118,28 @@ func (e *ExecutionData) Clean() error {
...
+	MaxBlockSize int64
+	// JoinExpiry is the number of blocks after which the validators
+	// join request expires if not approved.
+	JoinExpiry int64
+	// VoteExpiry is the default number of blocks after which the validators
+	// vote expires if not approved.
+	VoteExpiry int64
+	// DisabledGasCosts dictates whether gas costs are disabled.
+	DisabledGasCosts bool
+}
```
jchappelow, line 135: @charithabandi does this seem complete for our currently-used parameters?  Anything we don't use presently can be part of a scheme upgrade.

common/context.go
```diff
@@ -0,0 +1,23 @@
...
+
+// BlockContext provides context for all block operations.
+type BlockContext struct {
+	// ChainContext contains information about the chain.
+	ChainContext *ChainContext
+	// Height gets the height of the current block.
+	Height int64
+	// Proposer gets the proposer public key of the current block.
+	Proposer []byte
+}
```
jchappelow, line 23: Yeah, agree this is the way, for both of these context structs.  We can reuse these pretty much anywhere.  :+1:

internal/sql/pg/tx.go
```diff
@@ -114,3 +114,65 @@ func (tx *readTx) Rollback(ctx context.Context) error {
...
+	return nil
+}
+
+func (d *delayedReadTx) Execute(ctx context.Context, stmt string, args ...any) (*common.ResultSet, error) {
+	if err := d.ensureTx(); err != nil {
+		return nil, err
+	}
+
+	return d.tx.Execute(ctx, stmt, args...)
+}
```
jchappelow, line 148: I believe this transaction becomes abandoned on this path if `ensureTx` had to set `d.tx`.  For example, from `(*TxApp) AccountInfo` where it does `Execute` directly.  It will become an idle transaction in the postgres backend if not committed or rolled back.
Related, that `(*TxApp) AccountInfo` method should be adjusted to take either a `db.Executor` or a `db.TxMaker`.

jchappelow (changes requested):

internal/abci/meta/meta.go
brennanjl: Oh whoops. I had made a change to this function and then undid it

internal/sql/pg/tx.go
brennanjl: > I believe this transaction becomes abandoned on this path if ensureTx had to set d.tx

I don't think so. The understanding of `delayedReadTx` is that it works like any other tx, and the caller always needs to call Rollback or Commit. Calling Rollback or Commit will always clean this up. The `ensureTx` function is used to effectively "delay" the creation of the original tx, but from the perspective of the consumer, that tx is already open.

brennanjl: > Related, that (*TxApp) AccountInfo method should be adjusted to take either a db.Executor or a db.TxMaker. Probably worth checking all the interfaces on the params that take some dbish type.

I'm actually not too sure about this. There aren't many times (if ever) where a `sql.Executor` is not a `sql.TxMaker`, and vice versa. With something as high-level as TxApp, I could very possibly imagine a point in the future where we want transactionality and thus would have to use `sql.DB` (either via type assertion or changing the `AccountInfo` method). 

Overall its pretty minor, and I'm happy to change it if you feel strongly.

brennanjl: Have addressed all other changes, let me know if you still want me to make this one.

jchappelow: Is the following correct though?

```go
// sendTx makes a transaction and sends it to the local node.
func (svc *Service) sendTx(ctx context.Context, payload transactions.Payload) (*userjson.BroadcastResponse, *jsonrpc.Error) {
	readTx := svc.db.BeginDelayedReadTx()
	defer readTx.Rollback(ctx)

	// Get the latest nonce for the account, if it exists.
	_, nonce, err := svc.TxApp.AccountInfo(ctx, svc.db.BeginDelayedReadTx(), svc.signer.Identity(), true)
```

Did you mean to pass `readTx` to `svc.TxApp.AccountInfo`?

jchappelow: > With something as high-level as TxApp, I could very possibly imagine a point in the future where we want transactionality and thus would have to use `sql.DB` (either via type assertion or changing the `AccountInfo` method).

Fair point for the exported methods of `TxApp`, and other high level functions.

brennanjl: > Did you mean to pass readTx to svc.TxApp.AccountInfo?

Ah this is a bug. I meant to pass readTx

brennanjl: fixed

internal/sql/pg/tx.go
```diff
@@ -114,3 +114,65 @@ func (tx *readTx) Rollback(ctx context.Context) error {
...
+// the implementation might not need to use the transaction.
+type delayedReadTx struct {
+	db *DB
+
+	tx *readTx
+}
+
+func (d *delayedReadTx) ensureTx() error {
+	if d.tx == nil {
+		tx, err := d.db.BeginReadTx(context.Background())
```
jchappelow, line 131: Sorry, last thing.  Should we pass the context into `ensureTx`?  I was noting that `BeginDelayedReadTx` does not take a context for clear reasons (it doesn't do anything to cancel), but it seems that the `begin transaction ....` bits should be cancelable as well.

brennanjl: Ah yeah. Not sure why I did that, we definitely want to pass the context there.

brennanjl:
I'm going to make another change to this, to include the fork struct in the Chain Context

jchappelow (approved):
LGTM

internal/abci/abci.go
```diff
@@ -476,11 +612,55 @@ func validatorUpdates(initial, final []*types.Validator) []*types.Validator {
...
+		// giving pg_dump the snapshot ID to guarantee it has an isolated view of the database.
+		snapshotTx, snapshotId, err := a.db.BeginSnapshotTx(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to begin snapshot tx: %w", err)
+		}
+		defer snapshotTx.Rollback(ctx) // always rollback, since this is just for view isolation
+
+		err = a.snapshotter.CreateSnapshot(ctx, uint64(a.height), snapshotId)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create snapshot: %w", err)
```
charithabandi, line 658: Snapshot creation errors should be ignored, else the block production would stop. Just log the error and return nil to the cometbft

brennanjl: In what cases would this be an error though?

common/common.go
charithabandi: Yeah, this looks good.

internal/abci/abci.go
jchappelow: We actually just hit one when playing with the snapshots folder config setting:

on `main` it doesn't crash, only log:

```go
2024-06-17T22:08:22.186Z        error   kwild.tx-router txapp/txapp.go:816      failed to dump logical snapshot {"error": "failed to create snapshot at height 530: mkdir /data/kwil-snaps2/block-530: permission denied", "goversion": "go1.22.4", "mod_version": "(devel)", "revision": "6a09f17e03f532df2b3fa58ffb2c113b5d313779", "build_time": "2024-06-13T16:24:07Z", "dirty": tru
e}                                                                                                                                                                                          
github.com/kwilteam/kwil-db/internal/txapp.(*TxApp).snapshotDatabase                                                                                                                        
        /home/ubuntu/kwil/git/kwil-db/internal/txapp/txapp.go:816                                                                                                                           github.com/kwilteam/kwil-db/internal/txapp.(*TxApp).Commit                                                                                                                                  
        /home/ubuntu/kwil/git/kwil-db/internal/txapp/txapp.go:799                                                                                                                           github.com/kwilteam/kwil-db/internal/abci.(*AbciApp).Commit                                   
        /home/ubuntu/kwil/git/kwil-db/internal/abci/abci.go:479                                                                                                                             
github.com/cometbft/cometbft/abci/client.(*localClient).Commit                                                                                                                              
        /data/gopath/pkg/mod/github.com/cometbft/cometbft@v0.38.7/abci/client/local_client.go:113               
```

charithabandi:
```
snapshotTx, snapshotId, err := a.db.BeginSnapshotTx(ctx)
		if err != nil {
			return nil, fmt.Errorf("failed to begin snapshot tx: %w", err)
		}
```

Lets remove this too. Essentially intent is the node should produce blocks even if snapshot creation fails.

brennanjl: I see. Ok.

internal/abci/abci.go
jchappelow: I think this still needs resolving.  Everything else looks good now.

jchappelow (approved):

PR #832, opened 2024-06-17
add foreign caller

brennanjl:
This PR adds `@foreign_caller`, which gives information about what schema called another. If it is an outermost call, it will be an empty string (I tried to make it null, but due to Postgres not allowing unsettling of config parameters, and not allowing null config parameters, it does not appear possible).

This was requested by Truflation, as it is needed for their desired governance mechanisms for their schemas (https://github.com/truflation/tsn/issues/221). I can see how this would be useful information, so I felt it was appropriate to add.

This cannot be backported to v0.8.

### DIFF
--- a/internal/engine/generate/foreign_procedure.go
+++ b/internal/engine/generate/foreign_procedure.go
@@ -241,9 +241,7 @@ func GenerateForeignProcedure(proc *types.ForeignProcedure, pgSchema string, dbi
 
 	// set the foreign caller back to what it was
 	str.WriteString(`;`)
-	// this is a hack, since simply running SET LOCAL ctx.foreign_caller = __old_foreign_caller
-	// will set it to the string "__old_foreign_caller", not the value of the variable.
-	str.WriteString(`EXECUTE 'SET LOCAL ctx.foreign_caller = ' || quote_literal(__old_foreign_caller);`)
+	str.WriteString(`PERFORM set_config('ctx.foreign_caller', __old_foreign_caller, true);`)
 
 	// end block
 	str.WriteString(` END; $$ LANGUAGE plpgsql;`)
--- a/internal/engine/generate/foreign_procedure.go
+++ b/internal/engine/generate/foreign_procedure.go
@@ -15,7 +15,7 @@ import (
 // GenerateForeignProcedure generates a plpgsql function that allows the schema to dynamically
 // call procedures in other schemas, expecting certain inputs and return values. It will prefix
 // the generated function with _fp_ (for "foreign procedure").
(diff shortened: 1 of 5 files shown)


internal/engine/generate/foreign_procedure.go
```diff
@@ -231,8 +239,14 @@ func GenerateForeignProcedure(proc *types.ForeignProcedure, pgSchema string) (st
 		str.WriteString(formatStringList(argList))
 	}
 
+	// set the foreign caller back to what it was
+	str.WriteString(`;`)
+	// this is a hack, since simply running SET LOCAL ctx.foreign_caller = __old_foreign_caller
+	// will set it to the string "__old_foreign_caller", not the value of the variable.
+	str.WriteString(`EXECUTE 'SET LOCAL ctx.foreign_caller = ' || quote_literal(__old_foreign_caller);`)
```
jchappelow, line 246: It seems the non-hack is using the `set_config` function:

```go
str.WriteString(`perform set_config('ctx.foreign_caller', __old_foreign_caller, true);`)

This appears to be the counterpart of `current_setting`: https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-SET
```

jchappelow:
*Adding* something is manageable with a planned upgrade unlike removing something (albeit with potentially some transient liveness issues as 2/3+ validators upgrade of a relevant schema is deployed during the upgrade).  Only way this would be a major problem is if someone had already deployed a schema with `@foreign_caller` and it was rightly a failed transaction, because then it would succeed on resync.  Assuming that's unlikely, this change is so minor that it wouldn't be terrible on 0.8.x.

My other reservation about it on a 0.8.2 is that it's more than implied for a patch, but these things can happen.

Last thought is what a hard fork definition would look like if we wanted to go to that trouble.  First there's the obvious definition of the activation height in genesis.json after installing the new binaries.  Deciding how to make foreign caller available after some height can be done in different ways.  One is update the `parse.SessionVars` map at activation, which I don't love since it's not in engine and we'd probably have to guard that map.  Another is to pass height context down from the engine deploy method to all the generators i.e. make `sqlGenerator.allowForeignCaller bool` and put  logic like `if !s.allowForeignCaller && p0.Prefix == parse.VariablePrefixAt && p0.Name == "foreign_caller" { panic("foreign_caller is disallowed")	}` in `(*sqlGenerator) VisitExpressionVariable`.  That requires passing new info through large call stacks.  Not sure what would be simplest, or if it is worth it at all.

brennanjl:
For this case, I'd probably prefer to keep a preview branch. I am confident this won't be the last time we have this dilemma, so it might be better to just provide a preview for partners that need features asap.

jchappelow (approved):
LGTM.

PR #833, opened 2024-06-17
fixed default snapshot and statesync dirs

brennanjl:
Fixes the default directors for snapshot and statesync not being in the root directory.

### DIFF
--- a/build/package/docker/kwild.dockerfile
+++ b/build/package/docker/kwild.dockerfile
@@ -19,10 +19,10 @@ RUN GOWORK=off GIT_VERSION=$version GIT_COMMIT=$git_commit BUILD_TIME=$build_tim
 RUN GOWORK=off GIT_VERSION=$version GIT_COMMIT=$git_commit BUILD_TIME=$build_time CGO_ENABLED=0 TARGET="/app/dist" GO_RACEFLAG=$go_race ./scripts/build/binary kwil-cli
 RUN chmod +x /app/dist/kwild /app/dist/kwil-admin /app/dist/kwil-cli
 
-FROM alpine:3.19
+FROM ubuntu:24.04
 WORKDIR /app
 RUN mkdir -p /var/run/kwil && chmod 777 /var/run/kwil 
-RUN apk --no-cache add postgresql-client
+RUN apt update &&  apt install -y postgresql-client
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=build /app/dist/kwild ./kwild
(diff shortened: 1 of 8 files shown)


jchappelow:
In some default toml files we have:

```toml
# The directory where the snapshots are stored. Can be absolute or relative to the kwild root directory
snapshot_dir = "snapshots"
```

which gives me the following error when running the binary from somewhere the process does not have permission to write to:

```
Error: failed to sanitize config paths: failed to expand snapshot directory "snapshots": failed to get absolute path of file: snapshots due to error: stat .: permission denied
```

With the change in this PR, there's no sensible default for the toml, unless it's a full absolute path?

brennanjl:
The default could just be `~/.kwild/snapshots` right? Or does that not work on linux

jchappelow:
Let's just update the default config files that still specify "snapshots".  I'd prefer to have this be relative to root dir though.

> The default could just be `~/.kwild/snapshots` right? Or does that not work on linux

Same on linux, it's just problematic because what evaluates the "~" depends on where it is set.  If it's in config.toml, then it's the process' owner.  If it' on the cli flags, then it's the shell of the person running the command. These are often different.  When defining say a systemd service, it's going to launch as a user specified by the systemd .service file and a working directory that's generally irrelevant to the process.

brennanjl:
So we just need to remove the value from the default config?

charithabandi:
Oh shoot. I messed this up. Created same branch name as yours and the fix got pushed here. I get why jon recommends doing it from a fork

brennanjl:
> Oh shoot. I messed this up. Created same branch name as yours and the fix got pushed here. I get why jon recommends doing it from a fork

Lol no worries, I was going to close this in favor of what you put up.

build/package/docker/kwild.dockerfile
```diff
@@ -19,10 +19,10 @@ RUN GOWORK=off GIT_VERSION=$version GIT_COMMIT=$git_commit BUILD_TIME=$build_tim
 RUN GOWORK=off GIT_VERSION=$version GIT_COMMIT=$git_commit BUILD_TIME=$build_time CGO_ENABLED=0 TARGET="/app/dist" GO_RACEFLAG=$go_race ./scripts/build/binary kwil-cli
 RUN chmod +x /app/dist/kwild /app/dist/kwil-admin /app/dist/kwil-cli
 
-FROM alpine:3.19
+FROM ubuntu:22.04
```
jchappelow, line 22: 24.04?  No reason to start with the old release, right?

jchappelow (approved):

PR #838, opened 2024-06-18
added look back for token errors

brennanjl:
This PR improves error positions for syntax errors. Currently, syntax errors are only present on a single line and column, which leads to them often being incorrect. This PR makes them appear for ranges.

If desirable, we could make the errors always appear at only the beginning (this is what Go does), however I am not sure of all of the edge cases of doing this in Antlr, and it could very well result in errors appearing too early.

### DIFF
--- a/parse/errors.go
+++ b/parse/errors.go
@@ -71,6 +71,7 @@ type ParseErrs interface {
 type errorListener struct {
 	errs []*ParseError
 	name string
+	toks *antlr.CommonTokenStream
 }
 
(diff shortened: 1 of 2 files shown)


brennanjl:
If we choose to go with this, we should backport. admin.kwil.com should use this if we merge it

charithabandi:
lgtm. Not sure of the effects of it though

Yaiba (approved):
LGTM, a test case on the expected result would be nice

PR #839, opened 2024-06-19
[v0.8.2] Introduce genesis consensus params to configure max limits for votebody

charithabandi:


jchappelow (approved):
LGTM.  Just noting that it's a new consensus rule and if there happen to be old versions on the network and this vote limit is hit, the old validators will reject the block.  Since the genesis apphash isn't affected in the backport, this isn't cause for a network migration.  Still, meatspace upgrade coordination should be recommended in release notes if there are active resolutions on the network.

PR #840, opened 2024-06-19
abci: skip repeated tx sig verification in proposal processing

jchappelow:
Profiling (`http/pprof`) under heavy transaction load revealed that there was major CPU utilization in `ProcessProposal` -> `validateProposalTransactions` at `ident.VerifyTransaction`.  This PR allows to skip that if the transaction (identified by hash) has already gone through this check in `CheckTx`, which would happen either via p2p gossip or rpc broadcast of the transaction.

WIP since there's a `txCache` map with a slightly different purpose that I may try to remove.

With this change, in a network with multiple validators that have to process another validator's block proposal, transaction throughput is much higher on account of much faster block proposal processing (can vanish almost 2 seconds).

### DIFF
--- a/cmd/kwild/server/build.go
+++ b/cmd/kwild/server/build.go
@@ -622,6 +622,11 @@ func buildStatesyncer(d *coreDependencies) *statesync.StateSyncer {
 			continue
 		}
 
+		if res.Header.Height == 0 {
+			d.log.Warnf("zero height from provider %v", p)
+			continue
+		}
+
 		// Get the trust height and trust hash from the remote server
 		d.cfg.ChainCfg.StateSync.TrustHeight = res.Header.Height
 		d.cfg.ChainCfg.StateSync.TrustHash = res.Header.Hash().String()
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -35,7 +35,6 @@ import (
 
 	abciTypes "github.com/cometbft/cometbft/abci/types"
 	"github.com/cometbft/cometbft/crypto/ed25519"
-	cmtTypes "github.com/cometbft/cometbft/types"
 	"go.uber.org/zap"
 )
 
@@ -65,9 +64,9 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
 		log: log,
 
 		validatorAddressToPubKey: make(map[string][]byte),
-		txCache:                  make(map[string]bool),
-		verifiedSigs:             make(map[[32]byte]struct{}),
+		verifiedTxns:             make(map[chainHash]struct{}),
 	}
+	app.forks.FromMap(cfg.ForkHeights)
 
 	tx, err := db.BeginOuterTx(ctx)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


jchappelow:
Here's the hotspot that this PR is addressing (for cgo build):

image

for non-cgo build, same hotspot in our app, but different:

image

cmd/kwild/server/build.go
```diff
@@ -622,6 +622,11 @@ func buildStatesyncer(d *coreDependencies) *statesync.StateSyncer {
 			continue
 		}
 
+		if res.Header.Height == 0 {
+			d.log.Warnf("zero height from provider %v", p)
+			continue
+		}
```
jchappelow, line 628: Somewhat unrelated change here.  @charithabandi and I could not understand why sometimes a bad RPC provider would get past this build step without triggering the `failBuild` below.  My guess was that perhaps it was getting a nonsense response, but no error. :man_shrugging: wish I had checked the logs when it happened.

internal/abci/abci.go
```diff
@@ -66,21 +64,30 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get validators: %w", err)
+	}
+	for _, val := range validators {
+		addr, err := pubkeyToAddr(val.PubKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert pubkey to address: %w", err)
+		}
+
+		app.validatorAddressToPubKey[addr] = val.PubKey
```
jchappelow, line 88: Also somewhat unrelated other than being in abci, but this was moved from the `Info` method, which is not part of cometbft's consensus connection methods, making it not thread safe.  The map isn't guarded, so it should only be touched in constructor or the consensus connection methods that are used sequentially.

Honestly I don't know why this was done in `Info`, but my guess is that we were using it like a constructor where we only had DB there.  But now we've got DB from the outset, so this is an easy fix now.

internal/abci/abci.go
```diff
@@ -478,10 +501,9 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		res.TxResults = append(res.TxResults, abciRes)
 
 		// Remove the transaction from the cache as it has been included in a block
-		hash := cmtTypes.Tx(tx).Hash()
```
jchappelow, line 481: Here's another optimization in `FinalizeBlock`: it hashed each transaction twice.

internal/abci/abci.go
```diff
@@ -1210,8 +1219,12 @@ func (a *AbciApp) validateProposalTransactions(ctx context.Context, txns [][]byt
 
 			// This block proposal may include transactions that did not pass
 			// through our mempool, so we have to verify all signatures.
-			if err = ident.VerifyTransaction(tx); err != nil {
-				return fmt.Errorf("transaction signature verification failed: %w", err)
+			if !a.TxSigVerified(txO.hash) {
+				if err = ident.VerifyTransaction(tx); err != nil {
```
jchappelow, line 1223: This is the headline change in the PR.  Validators processing block proposals would do repeated and expensive signature verification.

jchappelow:
@charithabandi We've been running the tests with this one.  Look OK to merge?

charithabandi (approved):
lgtm!

PR #841, opened 2024-06-19
test: procedures in stress tool and performance improvements

jchappelow:
Based on top of https://github.com/kwilteam/kwil-db/pull/840

This started with the goal of simply adding procedures to the stress tool. and it now contains a good deal of related refactoring to increase transaction throughput.

With this PR and https://github.com/kwilteam/kwil-db/pull/840, we are finally able to fill blocks with small transaction <=1kB.

Needs a little cleanup since I was focusing on metrics and https://github.com/kwilteam/kwil-db/pull/840 while hacking on this, but for the most part it is what I wanted it to do.  Future work should wire up the foreign procedures schema in place of the simple one.

In addition to the stress tool updates, this also:
- performs more strict type assertions in acceptance tests (no need for conv package that permits too much)
- silence (via debug log instead of warn) a number of notices in the pg package that pertain to user dataset query errors


test/stress/foreign_social.kf
```diff
@@ -0,0 +1,139 @@
+database foreign_social_media;
```
jchappelow, line 1: The foreign proc schemas aren't wired up yet, but I want to soon.

test/stress/hammer.go
```diff
@@ -227,90 +259,62 @@ func hammer(ctx context.Context) error {
...
+	if maxPosters > 0 {
+		maxProcPost := maxPosters / 2
+		operation := "create post (proc)"
+		posters := make(chan struct{}, maxProcPost)
+		wg.Add(1)
+		go runLooped(ctx,
+			asyncFn(ctx, posters, h.printf, operation,
+				func() (<-chan asyncResp, error) {
+					content := bigData[:contentLen]
+					// content := bigData[:rand.Intn(maxContentLen)+1] // random.String(rand.Intn(maxContentLen) + 1) // randomBytes(maxContentLen)
```
charithabandi, line 309: Maybe add a flag to configure we want fixed size payload or variable?

cmd/kwild/config/default_config.toml
```diff
@@ -251,7 +251,7 @@ max_txs_bytes = 536870912
 max_tx_bytes = 4194304
 
 # Size of the cache (used to filter transactions we saw earlier) in transactions
-cache_size = 10000
+cache_size = 20000
```
charithabandi, line 254: I think we should have higher cache_size here, atleast to the size of mempool. This is what keeps track of previously seen transactions and avoids rechecks and duplicate txs in the mempool.

jchappelow: I would hope that a cache **miss** would just lead to a tx index check.  I'll increase this number, but do you have a link for how their mempool cache is works if there's a miss?  It's a badly cache-supported system that would result in duplicate txns in mempool in the event of a cache, so the purpose of this value doesn't seem clear.

test/stress/hammer.go
jchappelow: Yeah, good call.

cmd/kwild/config/default_config.toml
jchappelow: Bumped to 60,000 in https://github.com/kwilteam/kwil-db/pull/841/commits/a9ec473094c75cb8629e3e6f1a1830f51ce7aec9.
Will look into the functionality of their internal cache, but I think it's safer higher as you suggested.

test/stress/hammer.go
jchappelow: Added `-vl` flag for variable length: https://github.com/kwilteam/kwil-db/pull/841/commits/ebf62833893e8d6b23aed85e4e41b43c6f628447

jchappelow:
Needs merge **without squash**.  A few independent changes might be cherry-picked for 0.8

cmd/kwild/config/default_config.toml
charithabandi: Yeah, else we might get into the hash recalculations, which we really wanna avoid as much as possible.

core/utils/json/json.go
```diff
@@ -59,6 +59,10 @@ func convertJsonNumbers(val any) any {
 			return val.String()
 		}
 		return i
+	case string:
+		return val
+	case int64:
+		return val
```
jchappelow, line 65: Worth mentioning that these are just optimizations.  These types would end up in default and go through reflection to check sliceiness  only to return as-is.

Issue #843, opened 2024-06-19
apphash mismatch on every restart of node (in network)


jchappelow:
Given a network with at least two nodes (one can be sentry), you always get an apphash mismatch after restarting the node.

This is because we fail to load the apphash from the chain state db in the `AbciApp` constructor.  Previously this was done in `TxApp`'s constructor, but that was moved to ABCI in https://github.com/kwilteam/kwil-db/pull/831.  A `git bisect` confirms that's the cause.

Assigning myself.  Fix in preparation.

brennanjl:
Oh man, good catch. Sorry about that

jchappelow:
Was a really minor thing.  Hard to automatically test stuff like this where restarts are involved, hence why I messed up the consensus parameters storage issue in the first place.  :(

PR #844, opened 2024-06-19
abci: construct app with current appHash from chain meta store

jchappelow:
Resolves https://github.com/kwilteam/kwil-db/issues/843


jchappelow:
The reproduction was:
- start one validator node
- start sentry node, let it sync
- get to block 3 or later
- stop sentry node
- wait for validator to make a block or two (may not be needed)
- start sentry node
- first new block added on sentry block uses the wrong previous apphash (will look like block 1's app hash)

The issue seems to be resolved with this change.

charithabandi (approved):

PR #846, opened 2024-06-21
added genesis and commit hooks

brennanjl:
As per a discussion with the idOS team, we found an issue that could have been pretty elegantly solved by hooks that run arbitrary logic at genesis and at the end of a block. This is not the first time we have wanted these: in February, we wanted these hooks to build an Arweave integration. For migrations, I was also planning on a genesis hook. Given how stupid simple they are to implement, I decided to throw them together.

This also fixes a bug where the node identity is not being passed as part of the `*common.App` to other extensions.

### DIFF
--- a/extensions/hooks/hooks.go
+++ b/extensions/hooks/hooks.go
@@ -73,3 +73,8 @@ func ListEndBlockHooks() []EndBlockHook {
 
 	return hooks
 }
+
+func init() {
+	genesisHooks = make(map[string]GenesisHook)
+	endBlockHooks = make(map[string]EndBlockHook)
+}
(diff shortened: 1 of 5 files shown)


charithabandi (approved):
looks good!

jchappelow (approved):
LGTM too.
What kind of thing might be the end block hook?

brennanjl:
> What kind of thing might be the end block hook?

Deferred resolution application: https://gist.github.com/brennanjl/6937ffa74331a1e2db254ef69ff1172f

Any sort of job that should automatically happen periodically.

PR #848, opened 2024-06-21
backport hooks to v0.7

brennanjl:
backports https://github.com/kwilteam/kwil-db/pull/846

### DIFF
--- a/internal/txapp/txapp.go
+++ b/internal/txapp/txapp.go
@@ -163,7 +163,7 @@ func (r *TxApp) GenesisInit(ctx context.Context, validators []*types.Validator, 
 
 	// genesis hooks
 	for _, hook := range hooks.ListGenesisHooks() {
-		err := hook(&ctx, &common.App{
+		err := hook(ctx, &common.App{
 			Service: &common.Service{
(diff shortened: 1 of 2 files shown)


internal/txapp/txapp.go
```diff
@@ -318,6 +339,21 @@ func (r *TxApp) Finalize(ctx context.Context, blockHeight int64) (appHash []byte
 		return nil, nil, err
 	}
 
+	// end block hooks
+	for _, hook := range hooks.ListEndBlockHooks() {
+		err := hook(ctx, &common.App{
```
jchappelow, line 344: One of the things you mentioned earlier today as an example was passing apphash to the end block hook.   I suppose it's acceptable to only have access to the previous app hash?  Would a pre- and/or post-commit hook be something we'd want is this end-block hook called prior to Precommit adequate for current use cases?

brennanjl: That's a good point.

Maybe we can explore some other areas where hooks might be valuable, but I think we can keep this PR as-is for now (to mitigate backported features to only what idOS needs).

jchappelow (approved):

PR #851, opened 2024-06-21
fixed panic on parser returns

brennanjl:
Fixes a panic found when parsing a procedure that expects a return but does not return anything.

We tested for this with schema parsing, but the panic actually occurred when analyzing a procedure relative to a schema, so it slipped through our test cases. I've made a new test case to check for this.


jchappelow (approved):
lgtm

PR #856, opened 2024-06-22
added kuneiform testing framework

brennanjl:
This PR adds a testing framework for Kuneiform. The intent is for it to be useable by Kwil users who want to set up tests for their schemas that can run synchronously and avoid waiting for consensus.

There are two ways to use it:

1. `testing`: Users can import the `testing` package to write their own tests in Go. This gives a lot of flexibility, as they can code any sort of function they want against the engine.
2. `kwil-cli utils test`: Users can use `kwil-cli` to run tests. Tests can be defined in JSON, where users can specify the schemas to deploy, seed data, and execute actions/procedures and check the results.

The tests can be configured to talk to any Postgres connection, or users can tell the test to setup and teardown a test container.

I still have to write docs for this, but an example if it in usage can be found at `testing/proxy`, where I test 3 contracts that make a proxy pattern. I started on this because I wanted to test my proxy schema contract that I was sending to Fractal, but I got a bit carried away. I knew this was something we wanted for this release anyways, so I decided to just knock it out.

Edit: docs here https://github.com/kwilteam/docs/pull/123


brennanjl:
Docs here if anyone wants to try it out: https://github.com/kwilteam/docs/pull/123

testing/testing.go
```diff
@@ -0,0 +1,519 @@
...
+
+	req := testcontainers.ContainerRequest{
+		Image:        "kwildb/postgres:latest",
+		ExposedPorts: []string{"5432/tcp"},
+		Env: map[string]string{
+			"POSTGRES_HOST_AUTH_METHOD": "trust",
+		},
+		WaitingFor: wait.ForLog("database system is ready to accept connections"),
+	}
+	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
```
jchappelow, line 427: @Yaiba Do you think this is the best option?  This is for kwil-cli, so in our main module.  It would be best not to have testcontainers as a dependency in our main module, but if we have to that's alright I guess.  I'm just wondering what's gained by using `testcontainers` vs. say, os.Exec to docker run docker stop etc.

brennanjl: Yeah I thought about this. Another option is having a separate `kwil-test` tool that exists in the `test` package.

Yaiba: yeah I agree, I would like just calling a shell command.

testing/proxy/proxy_test.go
```diff
@@ -0,0 +1,213 @@
...
+				 VALUES ('42f856df-b212-4bdc-a396-f8fb6eae6901'::uuid, 'satoshi', '0xAddress'),
+				 ('d68e737d-708f-45f8-9311-317afcaccc63'::uuid, 'zeus', 'zeus.eth')`,
+			},
+		},
+		TestCases: []kwilTesting.TestCase{
+			{
+				// should create a user - happy case
+				Name:     "create user - success",
+				Database: "impl_1",
+				Target:   "create_user",
```
Yaiba, line 36: wow, if we can execute this in REPL would be very handy.

testing/testing.go
jchappelow: If there were a complex service definition with dependencies between containers, I'd probably get the use of testcontainers for a docker compose approach, but we're just firing up the pg container.

They need docker installed regardless, so running docker commands would be preferable unless there's a notable challenge that testcontainers addresses.

If `github.com/docker/docker/client` has a reasonable dependency tree, that could work too.  https://medium.com/@Frikkylikeme/controlling-docker-with-golang-code-b213d9699998

testing/proxy/proxy_test.go
brennanjl: what do you mean in REPL?

Yaiba: directly interact with engine

brennanjl: Ah I see. Yeah that would be pretty easy to do.

testing/testing.go
jchappelow: With just `github.com/docker/docker/client`, the go.mod deps aren't terrible:  https://gist.github.com/jchappelow/bebc12d3ed4744a93b7f25f92131ded1

brennanjl: Ok I removed the test containers dependency and just made it use `exec.Command`

testing/proxy/proxy_test.go
```diff
@@ -0,0 +1,215 @@
+//go:build rely_docker
+
+// package proxy is an example of how the testing package can be used. It tests
+// three contracts that are used to form a proxy pattern. An explanation of
+// proxy contracts in Solidity can be found here:
+// https://www.cyfrin.io/blog/upgradeable-proxy-smart-contract-pattern
+package proxy
+
+import (
+	"context"
```
Yaiba, line 10: This is nbd.  
I notice that the tab width in you IDE is configured to 8 spaces, that's why your newly created files like this one `       "context"`.

Can you try add `tab_width = 4` to `[{*.go,*.go2}]` sestion in `.editorconfig` file, see if it chages?

testing/testing.go
```diff
@@ -0,0 +1,518 @@
...
+				err = err2
+			} else {
+				err = errors.Join(err, err2)
+			}
+		}
+	}()
+
+	opts.Logger.Logf("running test container: %s", string(bts))
+
+	time.Sleep(1 * time.Second) // stupid hack needed for the container to be ready
```
Yaiba, line 448: This is kind too hacky(not stable), maybe apply retry logic on  `pg.NewDB` if got connection error?

Yaiba (changes requested):
I skimmed through the P

testing/proxy/proxy_test.go
Yaiba: Just to make this clear. The purpose of `.editorconfig` is to make sure all developers have same view of the source code, to avoid unnecessary pr conflicts.

Yaiba (approved):
LGTM

PR #864, opened 2024-06-24
removed an unneeded uuid case

brennanjl:
While reviewing something with @Yaiba , I realized that we had an unnecessary case in a switch statement. We checked for a UUID literal, but Kuneiform doesn't support UUID literals due to a deficiency in the ANTLR Go target making it impossible to parse.


Yaiba (approved):
LGTM

PR #866, opened 2024-06-24
fix: default time says 0s

brennanjl:
In `kwild -h`, anything that was a Duration value returned the default as 0s. This is misleading, because the default is actually set elsewhere. This simply omits the default from being printed.


jchappelow (approved):
I'm just recalling that `config.DefaultConfig` is not used for the values in `flagCfg` because how how merging configs work.  Every default set in `AddConfigFlags` is discarded.  Oh well.

jchappelow (approved):

PR #868, opened 2024-06-25
make dollarsigns optional in engine

brennanjl:
As discussed here https://kwilteam.slack.com/archives/C0578NPEU21/p1719337560326199?thread_ts=1718905393.728919&cid=C0578NPEU21, the idOS team got tripped up expecting the engine to accept values that do not have dollarsigns. This is understandable, because virtually all other Kwil tooling makes the $ optional. This PR makes the dollarsign optional for the engine.

### DIFF
--- a/internal/engine/execution/procedure.go
+++ b/internal/engine/execution/procedure.go
@@ -467,6 +467,8 @@ func makeExecutables(params []*generate.InlineExpression) []evaluatable {
 // orderAndCleanValueMap takes a map of values and a slice of keys, and returns
 // a slice of values in the order of the keys. If a value can be converted to an
 // int, it will be. If a value does not exist, it will be set to nil.
+// They keys are expected to match Kwil's bind syntax, i.e. $key.
+// The values in the value map can be either $key or key.
 func orderAndCleanValueMap(values map[string]any, keys []string) []any {
 	// we need to iterate over all values, and see if it has a $. If not,
 	// we need to add one so that the keys (which have $) match the values
--- a/internal/engine/execution/procedure.go
+++ b/internal/engine/execution/procedure.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"reflect"
 	"strings"
 
@@ -467,6 +468,24 @@ func makeExecutables(params []*generate.InlineExpression) []evaluatable {
 // a slice of values in the order of the keys. If a value can be converted to an
 // int, it will be. If a value does not exist, it will be set to nil.
 func orderAndCleanValueMap(values map[string]any, keys []string) []any {
+	// we need to iterate over all values, and see if it has a $. If not,
+	// we need to add one so that the keys (which have $) match the values
+	// (which do not have $)
+	cloned := false
+	for k, v := range values {
+		if k[0] != '$' {
+			// we need to copy the values map to ensure
+			// we do not modify the original map
+			if !cloned {
+				values = maps.Clone(values)
+				cloned = true
+			}
+
+			delete(values, k)
+			values["$"+k] = v
+		}
+	}
+
 	ordered := make([]any, 0, len(keys))
 	for _, key := range keys {
 		val, ok := values[key]
--- a/internal/engine/execution/procedure_test.go
+++ b/internal/engine/execution/procedure_test.go
@@ -1,0 +1,59 @@
+package execution
+
+import (
+	"maps"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_OrderAndClean(t *testing.T) {
+	type testcase struct {
+		name   string
+		values map[string]any
+		keys   []string
+		res    []any
+	}
+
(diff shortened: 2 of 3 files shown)


jchappelow (approved):
LGTM, although I went to test something, and I don't see any instance in product or test code where we pass anything but nil values to `(*GlobalContext) Execute`.  Just toying with `Test_SQL` in internal/engine/integration/sql_test.go, I cannot get past `ParseSQL`.  I'm feeling kinda dumb, what am I missing?

brennanjl:
> LGTM, although I went to test something, and I don't see any instance in product or test code where we pass anything but nil values to (*GlobalContext) Execute. Just toying with Test_SQL in internal/engine/integration/sql_test.go, I cannot get past ParseSQL. I'm feeling kinda dumb, what am I missing?

If you're getting an error at `ParseSQL`, your SQL is probably invalid. We should probably add tests that use the values, currently the only place we use it is in `extensions/precompiles/adhoc.go`

jchappelow:
For example, in sql_test.go / `Test_SQL`:

```go
		{
			name: "delete",
			pre:  `delete from users where name = 'satoshi'`,
			sql:  `select name from users where name = $name`,
			values: map[string]any{
				"$name": "zeus",
			},
			want: [][]any{
				{"zeus"},
			},
		},
```

```
        	Error:      	Received unexpected error:
        	            	(sql) undeclared variable: $name
        	            	 start 1:37 end 1:37
```

jchappelow:
I think v0.7 `(*GlobalContext) Execute` lets this by, but it does not work in v0.8

brennanjl:
Oh man thats a good point. The SQL parser should allow any variable there. Will get a fix for this

internal/engine/execution/procedure.go
```diff
@@ -467,6 +468,24 @@ func makeExecutables(params []*generate.InlineExpression) []evaluatable {
 // a slice of values in the order of the keys. If a value can be converted to an
 // int, it will be. If a value does not exist, it will be set to nil.
 func orderAndCleanValueMap(values map[string]any, keys []string) []any {
+	// we need to iterate over all values, and see if it has a $. If not,
+	// we need to add one so that the keys (which have $) match the values
+	// (which do not have $)
+	cloned := false
+	for k, v := range values {
+		if k[0] != '$' {
```
jchappelow, line 476: Any need to worry about contextual variables (`@` prefixed) here?
Would it be safer to deal with this in the `range keys` loop below, with a `!ok` condition handled by `values[strings.TrimPrefix(key, "$")]` or something similar?

brennanjl: Good question. We actually don't pass `@` prefixed variables via values anymore. We did in SQLite because you can bind `@` variables in a statement, but internally we translate them to `current_setting`, so it doesn't really matter.

If `@caller` comes in, it will get translated to `$@caller`, but it would get ignored anyways.

jchappelow: > but internally we translate them to `current_setting`, so it doesn't really matter.

Ah right.  I had my head in the past from the fractal thread, but they are on postresql with 0.7 too.

> If `@caller` comes in, it will get translated to `$@caller`, but it would get ignored anyways.

Even in the context of `(*dmlStmt).execute` where it's not an ad hoc query?  With an action, the keys passed would be from:

https://github.com/kwilteam/kwil-db/blob/4ea9379752af786bb6c310ad2c06e7be89326bfb/internal/engine/generate/actions.go#L51-L55

Is the above comment still accurate?  I think it is.

`orderAndCleanValueMap` is a simple utility function, so context is not obvious.

jchappelow: If `delete(values, k)` were removed below, we wouldn't lose the ability to lookup the contextual variable, if I'm reading this right.

brennanjl: Correct, however I'm not planning on backporting this, as I don't really think it is a bug, as much as a UX improvement

jchappelow: Yah, I agree.  Want to merge this now or are you planning for the ParseSQL fix here?

brennanjl: I'll make the ParseSQL fix in a separate PR since we will want to backport it

brennanjl: Let me update the docs for `orderAndCleanValueMap` though

jchappelow: Oh, so should `(*ProcedureContext) Values() map[string]any` stop setting the contextual variables?  That part of the code is a bit misleading now.  The SET LOCAL / current_setting approach is used for both actions and procs, right?  I guess it could still be helpful for debugging.

jchappelow (approved):

internal/engine/execution/procedure.go
brennanjl: It probably should. I considered this, but avoided it b/c it is a breaking change, and then forgot to get rid of it for v0.8. But yeah probably.

PR #870, opened 2024-06-26
added notice support

brennanjl:
Pushing this up in case you're interested @jchappelow, but no need for a review right now. This adds the ability to listen to notifications raised by specific txs. I've made the required additions to the pg package, but now need to make the higher level adjustments for abci + parse.

Switching gears to handle the `engine.Execute` bug, but will circle back to this for the other additions later.

EDIT: This PR makes two fairly important changes:

1. It makes a breaking change to the `core/types/client` package, where `Call` now returns logs.
2. It changes how we read data in the CLI driver of the integration test


jchappelow:
Cool, thanks.  I like it.
Could you merge the two `pCfg.ConnConfig.OnNotice` definitions in `NewPool`?  The one you added overwrites the other one for logging other postgresql notices emitted from `postgres` itself.

internal/sql/pg/conn.go
```diff
@@ -264,3 +294,42 @@ func (p *Pool) BeginReadTx(ctx context.Context) (sql.Tx, error) {
...
 	}, nil
 }
+
+// Listen listens for notifications thrown by the custom `notify()` function.
+// `notify` is similar to kwil's `error` function, but is used for non-error
+// logging in the database. It returns a `done` function that cleans up
+// the listener. Only notifications raised in the passed transaction will be
+// used. The passed Tx must either be either a dbTx or a readTx; if it is a
+// anything else, it will return an error.
+func (p *Pool) Listen(ctx context.Context, tx sql.Tx, ch chan string) (func(), error) {
```
jchappelow, line 304: This is really the only thing that raises an eyebrow.
When Listen gets used, we might find that it might as well be a method of the transactions, which contain a reference to the DB/Pool and thus the subscribers map.

internal/sql/pg/conn.go
```diff
@@ -145,6 +153,27 @@ func NewPool(ctx context.Context, cfg *PoolConfig) (*Pool, error) {
...
+
+		sub, ok := subscribers.Get(txid)
+		if !ok {
+			return
+		}
+
+		select {
+		case sub <- log:
+		default:
+			logger.Warnf("txid %d subscriber is full, dropping notice", txid)
```
jchappelow, line 173: Hopefully the presence or absence or even the content of the notice does not affect consensus in anyway. Can we engineer it to not affect consensus?

brennanjl:
> Could you merge the two pCfg.ConnConfig.OnNotice definitions in NewPool? The one you added overwrites the other one for logging other postgresql notices emitted from postgres itself.

Oops! Total accident, good catch.

internal/sql/pg/conn.go
brennanjl: Ah good point. I was debating this. Maybe we should just block here then?

internal/sql/pg/conn.go
brennanjl: Yeah I considered that. Happy to make that change

internal/sql/pg/conn.go
jchappelow: Possibly, if missing a log is a problem.  I guess it would be on account of cost?  What if the log contents didn't exactly  match on two different nodes?  Ideally we'd be robust to that too, unless this is typically something that affects consensus.

brennanjl: The log contents should exactly match, as it will be part of consensus. Enforcing skipping logs on account of cost should happen at a higher level IMO (ideally with a gas pricing mechanism when the `notify` function is called, but for now it will simply stop reading logs if too many have been emitted).

jchappelow: > The log contents should exactly match, as it will be part of consensus

Via cometbft's block hash, I assume?  (I don't know how this kind of emitted data is considered by the consensus engine.)  I don't expect _we'll_ be using these logs to affect our computed component of the app hash, just putting it in the TxResult to be retrieved by a tx query?

brennanjl: > I don't expect we'll be using these logs to affect our computed component of the app hash, just putting it in the TxResult to be retrieved by a tx query? i.e. it's not going to change application state, only blockchain data.

Exactly

brennanjl:
Ok, this is ready for review. Will add an integration test tonight, but via manually testing all seems to work well.

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -494,9 +499,36 @@ func (svc *Service) Call(ctx context.Context, req *userjson.CallRequest) (*userj
...
-	readTx := svc.db.BeginDelayedReadTx()
+	// we use a basic read tx since we are subscribing to notices,
+	// and it is therefore pointless to use a delayed tx
+	readTx, err := svc.db.BeginReadTx(ctx)
+	if err != nil {
+		return nil, jsonrpc.NewError(jsonrpc.ErrorNodeInternal, "failed to start read tx", nil)
+	}
 	defer readTx.Rollback(ctx)
 
+	logCh := make(chan string)
```
jchappelow, line 510: You probably want a buffer of 1.

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -494,9 +499,36 @@ func (svc *Service) Call(ctx context.Context, req *userjson.CallRequest) (*userj
...
+	var logs []string
+
+	done, err := readTx.Subscribe(ctx, logCh)
+	if err != nil {
+		close(logCh)
+		return nil, jsonrpc.NewError(jsonrpc.ErrorNodeInternal, "failed to subscribe to notices", nil)
+	}
+	defer done()
+	defer close(logCh) // close logCh after done to avoid potential infinite blocking.
```
jchappelow, line 519: As a general rule, the sender closes, but it's ok if the pattern is clear and there's no possible send on the channel past this point.

I'm a little uncertain about the infinite blocking comment.  Will dig in briefly.

internal/sql/pg/fns.go
```diff
@@ -0,0 +1,85 @@
...
+
+// containsTxid checks if a string is the proper format for a txid.
+// if it is, it will return the txid, the rest of the string, and true.
+// if it is not, it will return 0, "", false.
+func containsTxid(notice string) (int64, string, bool) {
+	// trim off the txid prefix
+	if !strings.HasPrefix(notice, "pgtx:") {
+		return 0, "", false
+	}
+	n := strings.TrimPrefix(notice, "pgtx:")
```
jchappelow, line 70: First line of `TrimPrefix` is `HasPrefix` all over.  Go solved this with `CutPrefix`:

```go
	// trim off the txid prefix
	notice, cut := strings.CutPrefix(notice, "pgtx:")
	if !cut {
		return 0, "", false
	}
```

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -519,6 +551,7 @@ func (svc *Service) Call(ctx context.Context, req *userjson.CallRequest) (*userj
 
 	return &userjson.CallResponse{
 		Result: btsResult,
+		Logs:   logs,
```
jchappelow, line 554: I'm not seeing any concurrency controls with the closure goroutine above that appends to this.

brennanjl: Ah yeah I guess we should close before this.

brennanjl:
Ok, I have changed it so that notices can be run synchronously. This involves an extra round-trip per Postgres transaction, but IMO that felt ok, since it is only one extra roundtrip during consensus

internal/sql/pg/conn.go
```diff
@@ -113,9 +121,39 @@ func NewPool(ctx context.Context, cfg *PoolConfig) (*Pool, error) {
...
+	subscribers := syncmap.New[int64, chan<- string]()
+
 	// NOTE: we can consider changing the default exec mode at construction e.g.:
 	// pCfg.ConnConfig.DefaultQueryExecMode = pgx.QueryExecModeSimpleProtocol
 	pCfg.ConnConfig.OnNotice = func(_ *pgconn.PgConn, n *pgconn.Notice) {
+		// Handling the Listen system.
+
+		// if this is the final log for a transaction, send a notice to the subscribers
+		if txid, isFin := isFinishTxid(n.Message); isFin {
+			subscribers.ExclusiveRead(func(m map[int64]chan<- string) {
```
jchappelow, line 134: I'm doing a more in depth review, but this being `ExclusiveRead` instead of `Exclusive` is probably the reason for the concurrent map write crash.

internal/abci/abci.go
```diff
@@ -594,6 +672,16 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		res.ValidatorUpdates[i] = abciTypes.Ed25519ValidatorUpdate(up.PubKey, up.Power)
 	}
 
+	// wait for all logs to be received
+	wg.Wait()
+	for _, result := range resultArr {
+		logs, ok := logMap[hex.EncodeToString(result.TxHash)]
```
jchappelow, line 678: The map key is a postgresql transaction id from `txid_current`, but this is trying to access by the blockchain transaction hash.

internal/sql/pg/fns.go
```diff
@@ -0,0 +1,114 @@
...
+		BEGIN
+			RAISE EXCEPTION '%', msg;
+		END;
+		$$ LANGUAGE plpgsql;`
+
+	sqlCreateFuncNotice = `CREATE OR REPLACE FUNCTION notice(payload text)
+		RETURNS void AS $$
+		DECLARE txid bigint;
+		BEGIN
+			SELECT txid_current() INTO txid;
```
jchappelow, line 26: I think an assumption made here and in ABCI's FinalizeBlick is that `txid_current` will be an ID for the nested transactions, but it only pertains to the outermost transaction.  So all blockchain transactions in a block will have the same `pgtx:<txid>` prefix.

internal/abci/abci.go
jchappelow: I might be mixing up something, or the PR is mixing concepts from the first iteration.  I see that parse defines `PGFormat` with `return fmt.Sprintf("notice('txid:' || current_setting('ctx.txid') || ' ' || %s)", inputs[0]), nil`.
So does this mean that we actually have to get the _blockchain_ txid from the log text, which would be _after_ the pg txid in the text?
So, for each tx result, check for a some log text that includes that chain txid in the logged notice?

jchappelow: Arg, nvm, I misread `ParseNotice`.  It's getting the blockchain txid as I would expect it should.

internal/sql/pg/conn.go
```diff
@@ -113,9 +121,39 @@ func NewPool(ctx context.Context, cfg *PoolConfig) (*Pool, error) {
...
+			})
+			return
+		}
+
+		// parse off any leading txid indicator. If successful, send the notice to the subscribers
+		if txid, log, ok := containsTxid(n.Message); ok {
+			subscribers.ExclusiveRead(func(m map[int64]chan<- string) {
+				sub, ok := m[txid]
+				if !ok {
+					return
```
jchappelow, line 151: Log this as an error too?

internal/sql/pg/conn.go
```diff
@@ -251,16 +290,59 @@ func (p *Pool) BeginTx(ctx context.Context) (sql.Tx, error) {
...
+
+	subscribers.Exclusive(func(m map[int64]chan<- string) {
+		_, ok := m[txid]
+		if ok {
+			err = fmt.Errorf("pgtx %d already has a subscriber", txid)
+			return
+		}
+
+		m[txid] = ch
+	})
```
jchappelow, line 333: `subscribers.Exclusive` may set `err`, so need to check after?

```
if err != nil {
	return nil, nil, err
}
```

internal/sql/pg/conn.go
jchappelow: Circling back on this, @charithabandi discovered that cometbft does not consider logs or events in the deterministic part of tx results, so block hash should be unaffected regardless.

brennanjl:
Ok, feedback is addressed. I am going to do some testing now to see if I can recreate some of the errors.

jchappelow (dismissed):
I dropped the ball on this last week, but LGTM now.
Despite being a tricky implementation, it appears to work as intended now.  Let's put it to use on main.

jchappelow (approved):

PR #871, opened 2024-06-26
fix engine.Execute

brennanjl:
Resolves the bug found in https://github.com/kwilteam/kwil-db/pull/868 where `engine.Execute` can't use parameterized SQL.

### DIFF
--- a/internal/engine/integration/sql_test.go
+++ b/internal/engine/integration/sql_test.go
@@ -190,6 +190,18 @@ func Test_SQL(t *testing.T) {
 				{int64(1)},
 			},
 		},
+		{
+			// this is a regression test for a bug introduced
+			// in v0.8
+			name: "values",
+			sql:  "select $id",
+			values: map[string]any{
+				"id": "4a67d6ea-7ac8-453c-964e-5a144f9e3004",
+			},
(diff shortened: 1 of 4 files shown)


jchappelow (approved):

jchappelow:
I feel like we should consider a sql parsing mode that works with a nil schema, allowing unknown table and column references.   e.g. `SELECT $id FROM tblX WHERE name = $name`.

brennanjl:
> I feel like we should consider a sql parsing mode that works with a nil schema, allowing unknown table and column references. e.g. SELECT $id FROM tblX WHERE name = $name.
(unknown table ref and unknown column ref)

Ah man you are totally right. I rushed this and absolutely neglected that. Will address

brennanjl:
@jchappelow actually I don't think we want to support this. When we parse the raw SQL, we still parse it in the context of a schema. The reason the parameters are unknown is because we aren't in a procedure/action. However, the parse function still takes the schema during the parse stage, and checks whether the tables / columns / procedures used within the SQL exist.

jchappelow:
OIC, that makes sense.  We don't have a need to implement a generic SQL parser that doesn't have context on the schema.

PR #874, opened 2024-06-26
fix crash on invalid snapshot

brennanjl:
I ran into an issue where I stopped by node at just the right time where it was able to get bricked and become unrecoverable without manual intervention. While creating a snapshot, my node stopped, and was then unable to restart. It would crash on restart due to a nil pointer dereference. This PR fixes this.

### DIFF
--- a/internal/statesync/store.go
+++ b/internal/statesync/store.go
@@ -2,6 +2,7 @@ package statesync
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


jchappelow (approved):
OK but we should Close those files that we Open IMO.

jchappelow:
LoadSnapshotChunk has the Open problem too.  That method I believe is called on demand by peers, so DOS

internal/statesync/store.go
```diff
@@ -245,20 +245,23 @@ func (s *SnapshotStore) loadSnapshots() error {
...
 		header, err := loadSnapshot(headerFile)
 		if err != nil {
 			s.log.Debug("Invalid snapshot header file, ignoring the snapshot", log.String("height", height), log.String("Error", err.Error()))
+			continue
 		}
 
 		// Ensure that the chunk files exist
 		for i := uint32(0); i < header.ChunkCount; i++ {
 			chunkFile := snapshotChunkFile(s.cfg.SnapshotDir, heightInt, DefaultSnapshotFormat, i)
 			if _, err := os.Open(chunkFile); err != nil { // chunk file doesn't exist
```
jchappelow, line 262: This one should use os.Stat or one of the "file exists" helpers laying around

charithabandi (approved):

charithabandi:
Looks good!

PR #880, opened 2024-07-05
fix in-line action error

brennanjl:
This PR fixes two bugs:

1. Actions using in-line statements with type casts for ambiguous types would not generate SQL properly. This would only occur when performing an expression (such as a comparison). This is the main purpose of this PR.
2. Another very minor bug that caused a stack overflow when using the parse debugger tool was fixed. It was caused by having circular pointer references while removing error positions.

### DIFF
--- a/internal/engine/generate/plpgsql.go
+++ b/internal/engine/generate/plpgsql.go
@@ -134,7 +134,12 @@ func (s *sqlGenerator) VisitExpressionVariable(p0 *parse.ExpressionVariable) any
 		// Postgres uses $1, $2, etc. for numbered parameters.
 
 		s.orderedParams = append(s.orderedParams, str)
-		return "$" + fmt.Sprint(len(s.orderedParams))
+
+		res := strings.Builder{}
+		res.WriteString("$")
+		res.WriteString(fmt.Sprint(len(s.orderedParams)))
+		typeCast(p0, &res)
+		return res.String()
 	}
 
 	str := strings.Builder{}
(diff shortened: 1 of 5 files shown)


internal/engine/generate/plpgsql.go
```diff
@@ -134,7 +134,12 @@ func (s *sqlGenerator) VisitExpressionVariable(p0 *parse.ExpressionVariable) any
 		// Postgres uses $1, $2, etc. for numbered parameters.
 
 		s.orderedParams = append(s.orderedParams, str)
-		return "$" + fmt.Sprint(len(s.orderedParams))
+
+		res := strings.Builder{}
+		res.WriteString("$")
+		res.WriteString(fmt.Sprint(len(s.orderedParams)))
```
jchappelow, line 140: `fmt.Sprint` is slow and kinda overkill here.  I'd use `res.WriteString(strconv.Itoa(len(s.orderedParams)))` to skip all the printer's reflection that would end up doing that eventually.

brennanjl:
Ok addressed the feedback. I will wait until this gets merged before updating the backport pr.

jchappelow (approved):

jchappelow:
Could the parse bug be exploited on a node or is it really just the cli debug command?

brennanjl:
> Could the parse bug be exploited on a node or is it really just the cli debug command?

Nope, it is just the CLI. The bug was caused via infinite recursion in `parse.RecursivelyVisitPositions` function. This function is used exactly twice: in the CLI, and in testing.

PR #881, opened 2024-07-05
[v0.8] fix: in-line type assertions not comparable

brennanjl:
Backports https://github.com/kwilteam/kwil-db/pull/880

This also includes a fix for parse. It is not time sensitive that we bump our parse module version on the release, as the parse fix simply fixes a non-critical bug that is not used in consensus, and isn't a super important feature.


brennanjl:
closed in favor of https://github.com/kwilteam/kwil-db/pull/885

PR #882, opened 2024-07-12
Fix Scanner error: token limit exceeded

charithabandi:


jchappelow (changes requested):
You have a debug.test file checked in.  We should add that pattern to .gitignore.

`max_row_size` for the default and example config tomls?

internal/statesync/snapshotter.go
```diff
@@ -43,13 +43,15 @@ const (
 type Snapshotter struct {
 	dbConfig    *DBConfig
 	snapshotDir string
+	MaxRowSize  int
```
jchappelow, line 46: Really no need to export the field, unless we want to set it from the caller directly rather than via the constructor, which would be fine too.

charithabandi: ah, autocorrected. changing back

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -36,6 +36,8 @@ ls /path/to/snapshot/dir
 genesis.json    kwildb-snapshot.sql.gz`
 
 	createDatabase = `CREATE DATABASE `
+
+	maxRowSize = 4 * 1024 * 1024 // 4MB
```
jchappelow, line 40: Should this not be defined in `createCmd` with the other bound flag vars?

jchappelow (approved):

jchappelow:
@brennanjl This was a relatively trivial fix.  It covers both places where the Scanner lacked a large enough buffer: (1) kwil admin create snapshot, and (2) `internal/statesync` for kwild.  I went ahead and merged.

PR #883, opened 2024-07-12
feature(core): option to skip checking chain info

Yaiba:
This pr adds an option to skip checking chain info when create Client, this is useful to KGW to reduce the API calls to kwild.

### DIFF
--- a/core/types/client/opts.go
+++ b/core/types/client/opts.go
@@ -24,6 +24,10 @@ type Options struct {
 	// communication with a trusted node (using TLS or Unix sockets).
 	ChainID string
 
+	// TrustLocalChainID will trust the chain ID passed from the user, i.e. no check
+	// against remote node's chain ID. This option is only be effective when ChainID is set.
+	TrustLocalChainID bool
+
 	// Silence silences warnings logged from the client.
 	Silence bool
 
@@ -52,6 +56,8 @@ func (c *Options) Apply(opts *Options) {
 	if opts.Conn != nil {
 		c.Conn = opts.Conn
 	}
+
+	c.TrustLocalChainID = opts.TrustLocalChainID
 
 	c.Silence = opts.Silence
 }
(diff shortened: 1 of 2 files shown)


core/types/client/opts.go
```diff
@@ -24,6 +24,10 @@ type Options struct {
 	// communication with a trusted node (using TLS or Unix sockets).
 	ChainID string
 
+	// TrustLocalChainID will trust the chain ID passed from the user, i.e. no check
+	// against remote node's chain ID. This option is only be effective when ChainID is set.
+	TrustLocalChainID bool
```
Yaiba, line 29: I'm not really like this name, any suggestion?

core/types/client/opts.go
```diff
@@ -92,6 +98,14 @@ func WithChainID(chainID string) Option {
 	}
 }
 
+// WithTrustLocalChainID will trust the chain ID passed from the user, i.e. no check
+// against remote node's chain ID. This option is only be effective when ChainID is set.
+func WithTrustLocalChainID() Option {
```
Yaiba, line 103: I feel I can actually just delete all `func Withxxx() Option` functions, since we only use `Options`

core/types/client/opts.go
jchappelow: `SkipVerifyChainID` maybe

core/types/client/opts.go
jchappelow: This is a weird one.  I don't know why we have all the functional options funcs but the Options struct has settable exportable fields...  We should pick an approach.  I find functional options cumbersome, but I can live with it.  Whatever y'all want.

Yaiba: This is easy pick, those are just relic code and we no longer used those `Option` type in sdk.
I'll delete those functions

jchappelow (approved):
LGTM, but not personally tested.

PR #885, opened 2024-07-15
[v0.8] fix: in-line type assertions not comparable 

brennanjl:
backports https://github.com/kwilteam/kwil-db/pull/880

replaces https://github.com/kwilteam/kwil-db/pull/881


brennanjl:
An additional change is necessary for release v0.8 to make the tests pass. v0.9 does not require `$` prepended in `engine.Execute`'s value map, however v0.8 does.

jchappelow (approved):

PR #887, opened 2024-07-15
Network migrations continued...

charithabandi:
Testing network migrations:

### Submit a migration proposal:
```sh
kwil-admin migrate propose <start-height> <migration-duration> <chain-id>
```

### List pending migration proposals:
```sh
kwil-admin migrate list
```
This will list the migration proposal ID and the proposal contents.  Proposal ID will be used to vote or check the status. If a proposal is already resolved, it wont show up here. 

### Vote on the proposal
```sh
kwil-admin migrate approve <proposalID>
```

### Check proposal status
```sh
kwil-admin migrate status <proposalID>
```

Once the migration starts, the old nodes will take the database snapshot at the activation height and generates a genesis file and start storing changesets(user schema changes and accounts changes) to disk. 

- deploys, drops, adding new resolutions, voting, credits, transfers etc are all disabled during migration

### Download genesis state
```sh 
kwil-admin migrate genesis-state --root-dir <path/to/root-dir> 

Migration State:
    Start Height: 1000
    End Height: 5000
    Genesis File: /path/to/root-dir/genesis.json
    Snapshot File: /path/to/root-dir/snapshot.sql.gz
```

You may wanna provide `-rpcserver` flag if the server is not using default socket for admin service. 

This command will download the genesis and snapshot files to the directory specified in the flags. 

### Configuring new nodes
- Install latest kwild binaries
- Download the `genesis.json` and `snapshot.sql.gz` as shown in the previous step into the new node's root directory
- Make the below changes to the `config.toml` file:

> 
> ```
> # Configure the Kwild to boot using a snapshot file
> [app]
>     genesis_state = "/path/to/snapshot/file"
> ```

> ```
> # Configure the rpc provider to migrate the state from the old network
[app]
migrate_from = "http://localhost:8484"

- Make required changes on all the listening addresses and the persistent peers. 
- Configure correct postgres settings. 
- Run kwild node.


jchappelow:
You'll want to drop the first two commits that are already merged to main.

internal/abci/abci.go
```diff
@@ -52,18 +57,72 @@ type AbciConfig struct {
 	ForkHeights        map[string]*uint64
 }
 
+type AbciApp struct {
```
jchappelow, line 60: I agree with moving this struct up here near the constructor.  It was hard to find before.  Looks like the only change is the added `migrator` field.  Is that correct?

charithabandi: Yup

internal/sql/pg/repl_changeset.go
```diff
@@ -239,59 +245,236 @@ func (c *ChangesetGroup) UnmarshalBinary(data []byte) error {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// DeserializeChangeset deserializes a changeset stream.
+func DeserializeChangeset(data io.Reader) (*ChangesetGroup, error) {
```
jchappelow, line 419: For this iteration in this PR, I think it's alright as a PoC, but we're defeating the purpose of streaming the changeset data.  We don't have a production ready solution if it involves loading an entire changeset into memory.

On the changeset **writer** (old node collecting and storing changesets to be served on RPC), the `storeChangesets` functionality needs to open a file and streaming write the changeset data into it, one tuple update at a time, not using WriteFile for the whole thing.

On the changeset **receiver** the sequence of `getChangesets` -> `unmarshal` -> `ApplyChangesets`  is also working with the entire change set group in memory at once.  This will need to change, and all of the `(*Changeset).applyXXX` methods will probably go away and be replaced with functions that apply individual insert/update/deletes as they are read and decoded from the full on-disk changeset.

charithabandi: :/ Yeah, This is what I was referring to that day at lunch. I am gonna take one more stab at it, got some ideas.

internal/abci/abci.go
```diff
@@ -672,6 +681,12 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		a.halted.Store(true)
 	}
 
+	// Notify the migrator of the changeset
+	err = a.migrator.NotifyHeight(ctx, &blockCtx, a.db, writer)
```
jchappelow, line 685: Pertaining to https://github.com/kwilteam/kwil-db/pull/887/files#r1702316468, here we've got this `io.ReadWriter` that's supposed to function as a pipe between the changeset collection from the pg logical replication stream and the storage routines.  However these two things two things cannot happen concurrently yet, with the current design.  That is, `NotifyHeight` does needs to be called after all the above is done, but before we call `Precommit` (and it starts writing to the `io.Writer`), another goroutine needs to be consuming the `io.Reader` and writing to disk.  i.e. the part that does `storeChangeset`

internal/sql/pg/repl_changeset.go
jchappelow: OK.  I edited my comment BTW.  I'm not entirely convinced we want to be streaming opaque data (bytes through io.ReadWriter) rather than structured updates as one of the tuple update types over a channel.  Something to consider.  Part of the reason I'm looking at it like that is because I'm considering table statistics updates, and I want a stream of the updates *outside of the `pg` package*, not in the form of opaque bytes.  I can approach it differently by making keeping statistics updates internal to pg... I don't know yet.

jchappelow: I'm curious if anything panned out on Friday.  The solution here affects how I perform stats updates.  Either I have the `pg` package do everything as a black box:

```go
Precommit(ctx, context.Context, w io.Writer, statsToUpdate *sql.Statistics) error
```

or we stream the structured change set updates instead of opaque changeset content.  i.e.

```go
Precommit(ctx, context.Context, csChan chan<- any) error
```

with the elements being sent on the channel being stuff like this:

```go
// Relation is a table in a schema.
type Relation struct {
	Schema string
	Name   string
	Cols   []*Column
}

type ChangesetEntry struct {
	// Relation // schema, table name, and column info (names and types)

	// RelationIdx is the index in the full relation list for the changeset that
	// precedes the tuple change entries.
	RelationIdx uint32

	OldTuple []*TupleColumn // empty for insert
	NewTuple []*TupleColumn // empty for delete
	// both old and new are set for update
}

func (ce *ChangesetEntry) Kind() string {
	if len(ce.NewTuple) == 0 {
		return "delete"
	}
	if len(ce.OldTuple) == 0 {
		return "insert"
	}
	return "update"
}

func (ce ChangesetEntry) String() string {
	return fmt.Sprintf("Change type %s, rel ID %d, %d old tuples, %d new tuples",
		ce.Kind(), ce.RelationIdx, len(ce.OldTuple), len(ce.NewTuple))
}
```

And then the consumer makes the decisions about, every thing, inc:

- serialization for storage and transmission
- table statistics update
- whatever

Experimented with this on Friday (but didn't actually remove the io.Writer yet): https://github.com/jchappelow/kwil-db/commit/f424b0e8a1ef7b0e035025edd4dc93029f957bec

Either way can work for me.  It's more versatile streaming out structured data, but I can see wanting to keep certain details internal to `pg` package.  On the other hand, we're not talking about exporting outside of kwil-db repo and these are all internal packages, so I think the stakes are low.

Depending on where you got, maybe we have a real life huddle to discuss?

brennanjl: Seems ok to me, except my only concern is on the useability. Consumers will need to keep a map for all of their known relations, and then using that map, decode column values. If there's no other way around it then nbd, but it does just feel a bit cumbersome.

jchappelow: We can re-send the relation definitions.  I had considered that too as a usability issue, and agree.

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,272 @@
...
+			for err := range errChan {
+				errs = errors.Join(errs, err)
+			}
+
+			if errs != nil {
+				return errs
+			}
+
+			if btsReceived != size {
+				return fmt.Errorf("incorrect changeset size received: %d, expected: %d", btsReceived, size)
```
brennanjl, line 184: Not sure about the error return right here, since it will halt the node entirely, and it can be caused by data received external to the listener being manipulated (if the migration metadata says it should expect some size and send another). Might be better to just log and return nil as to not crash the node.

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,272 @@
...
+			}
+
+			if err = setLastStoredHeight(ctx, eventStore, currentHeight); err != nil {
+				return err
+			}
+			currentHeight++
+
```
brennanjl, line 191: Overall the listener looks good, however could you implement exponential backoff? It is probably nbd, but it is something that Randal used to obsess over, and I think it's better to spend the 20 minutes to do it right.

brennanjl: If it's helpful, eth deposit oracle has very basic exponential backoff for reference

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return err
+	}
+
+	c.AdminListenAddress, ok = cfg["admin_listen_address"]
+	if !ok {
+		return errors.New("migration admin_listen_address not provided")
+	}
+
+	c.AdminPass = cfg["admin_pass"]
```
brennanjl, line 225: One of the reasons I want the migration JSON RPC getter methods bound to txsvc is because it allows users to migrate without exposing their admin endpoint (which by default only runs on UNIX sockets). This makes it easier to leverage this feature. Assuming we go ahead with this change, we should update that in the oracle configuration.

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,272 @@
...
+			}
+			currentHeight++
+
+			// Delay before polling the admin server again for the next changeset
+			time.Sleep(retryDelay)
+		}
+	}
+}
+
+func (c *MigrationConfig) ExtractConfig(cfg map[string]string) error {
```
brennanjl, line 198: I'm not 100% sure how we would achieve this, but it would be awesome to allow oracle configuration to be done as part of Kwild's "official" configuration, rather than relying on the double-nested map for user-defined extensions. 

I guess for now nbd since this is a different feature altogether, but it would be great if users could register extension configs as an "official flag" that would show up with their locally built binary's `-h` flag.

internal/migrations/changesets.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/kwilteam/kwil-db/common/sql"
+	"github.com/kwilteam/kwil-db/core/log"
+	"github.com/kwilteam/kwil-db/core/types/serialize"
+	"github.com/kwilteam/kwil-db/extensions/resolutions"
+	"github.com/kwilteam/kwil-db/internal/sql/pg"
+)
+
+const (
+	// "changeset_migration" is the event type used for replication of changesets from old chain to new chain during migration.
+	ChangesetMigrationEventType = "changeset_migration"
```
brennanjl, line 19: unexport?

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// AdminListenAddress is the address of the admin server to receive changesets from.
+	AdminListenAddress string
+
+	// AdminPass is the admin server password.
+	AdminPass string
+
+	// KwildTLSCertFile is the path to the TLS certificate file for the Kwil node.
+	// The path should be an absolute path or relative to the node's root directory.
+	KwildTLSCertFile string
```
brennanjl, line 50: Can we further automate this? Essentially, just give the new node an endpoint to connect to, and it will contact that node, read the migration metadata, and apply it all itself?

internal/migrations/changesets.go
```diff
@@ -0,0 +1,319 @@
...
+// The changesets are applied in the order of block heights.
+var ChangesetMigrationResolution = resolutions.ResolutionConfig{
+	ConfirmationThreshold: big.NewRat(2, 3),
+	ExpirationPeriod:      24 * 60 * 10, // 1 day in blocks
+	ResolveFunc: func(ctx context.Context, app *common.App, resolution *resolutions.Resolution, block *common.BlockContext) error {
+
+		// Extract the migration config from the app, executed only once
+		if migrationCfg == nil {
+			migrationCfg = &MigrationConfig{}
+			cfg, ok := app.Service.ExtensionConfigs[ListenerName]
```
brennanjl, line 80: Feels like a big footgun, since a simple misconfiguration could lead to a validator forking themselves.

IMO, I think it would make more sense to include the previous migration state as part of the network's genesis data (maybe. as part of the snapshot). This reduces the risk of screwing up, and decreases the manual coordination needed to upgrade a network.

internal/migrations/sql.go
```diff
@@ -28,24 +57,37 @@ var (
 
 	// getMigrationSQL is the sql query used to get the current migration.
 	getMigrationSQL = `SELECT start_height, end_height, chain_id FROM ` + migrationsSchemaName + `.migration;`
-	// migrationIsActiveSQL is the sql query used to check if a migration is active.
+	// migrationIsActiveSQL is the sql query used  to check if a migration is active.
```
brennanjl, line 60: I believe the answer is yes, but just to confirm, this schema is part of statesync / snapshotting right?

internal/migrations/changesets.go
```diff
@@ -0,0 +1,319 @@
...
+
+			// Increment the last changeset
+			if err = setLastChangeset(ctx, tx, currentHeight); err != nil {
+				return err
+			}
+
+			currentHeight += 1 // move to the next height
+		}
+
+		return tx.Commit(ctx)
```
brennanjl, line 177: I know this is a somewhat substantial change, but after looking at this, I think there is potentially a less error prone way to do it:

Within each changeset chunk's resolution, you should store the data locally as a part of some database that is tied to the networks state (you already do this). Then, you have sn `EndBlockHook` that, while the migration is running, queries the local database for any blocks where it has all of the chunks, but has not applied them, and then applies them and deletes them from the local store.

internal/migrations/changesets.go
```diff
@@ -0,0 +1,319 @@
...
+			// If the current height is greater than the migration end height, break
+			if currentHeight >= migrationCfg.EndHeight {
+				app.Service.Logger.Info("changeset migration completed", log.Int("height", currentHeight))
+				break
+			}
+
+			// Check if all chunks have been received for the current height, if not, break
+			totalChunks, chunksReceived, err := getChangesetMetadata(ctx, tx, currentHeight)
+			if err != nil {
+				break
```
brennanjl, line 149: Shouldn't this return an error?

internal/migrations/migrations.go
```diff
@@ -23,19 +23,36 @@ import (
...
 	"github.com/kwilteam/kwil-db/core/types/serialize"
 	"github.com/kwilteam/kwil-db/extensions/resolutions"
 )
 
+const (
+	StartMigrationEventType = "migration"
+)
+
+// migrator instance responsible for managing zero downtime migrations.
+var migrator *Migrator
```
brennanjl, line 36: Seems unnecessary to make this a global singleton rather than an object passed around? AFAICT it is used in abci and the rpc service, both which take it as a dependency in their constructors. Furthermore, your `SetupMigrator` returns it.

brennanjl: Hmm, I see that you have a `migration` resolution which starts the migration process. This seems related to another comment I made, where it seems like the more appropriate place to define a migration is with the genesis data?

internal/migrations/sql.go
charithabandi: Yup, it is.

cmd/kwil-admin/cmds/setup/testnet.go
```diff
@@ -94,8 +94,8 @@ func testnetCmd() *cobra.Command {
 	cmd.Flags().IntVarP(&nonValidatorAmount, "non-validators", "n", 0, "number of non-validators to generate")
 	cmd.Flags().BoolVar(&withGas, "gas", false, "enable gas")
 	cmd.Flags().Var(&allocs, "alloc", "account=amount pairs of genesis account allocations")
-	cmd.Flags().BoolVar(&snapshotsEnabled, "--snaps", false, "enables db snapshots")
-	cmd.Flags().Uint64Var(&maxSnapshots, "--max-snaps", 3, "maximum number of snapshots to store in the device")
-	cmd.Flags().Uint64Var(&snapshotHeights, "--snap-heights", 10000, "recurring heights(multipes of --snap-heights) to take snapshots at")
+	cmd.Flags().BoolVar(&snapshotsEnabled, "snaps", false, "enables db snapshots")
+	cmd.Flags().Uint64Var(&maxSnapshots, "max-snaps", 3, "maximum number of snapshots to store in the device")
+	cmd.Flags().Uint64Var(&snapshotHeights, "snap-heights", 10000, "recurring heights(multipes of --snap-heights) to take snapshots at")
```
jchappelow, line 99: Sorry to keep nit picking the parenthesis, but this looks very odd.

jchappelow (commented):
Review pt 1/n.  It's taking me time, so dropping my initial comments here now.

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,272 @@
...
+				// Synced up with the old chain, no more changesets to pull. Close the listener
+				service.Logger.Info("changesets have been synchronized with the old chain", log.Int("height", currentHeight))
+				return nil
+			}
+
+			// Get the changeset metadata from the admin server
+			numChunks, size, err := adminClient.ChangesetMetadata(ctx, currentHeight)
+			if err != nil {
+				fmt.Println("Couldnt fetch changesets", err)
+				time.Sleep(retryDelay) // If no changeset is found, wait and try again
```
jchappelow, line 117: Should try to avoid `time.Sleep` in almost all cases, here and in the goroutines below.  The main reason is shutdown can't bypass, as it's stuck waiting even if context canceled.   This will be especially irritating if you move to an exponential backoff as the delays will just get longer.  Better to refactor with a receive from a timer/ticker/After with a competing case on the context in case of shutdown.

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,272 @@
...
+		currentHeight = lastHeight + 1
+	}
+
+	// Start polling the admin server for changesets
+	service.Logger.Info("start syncing changesets from old chain", log.Int("startHeight", cfg.StartHeight))
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
```
jchappelow, line 106: Unless there's something else in this select, this select doesn't do anything useful.  Any blocking function call would get the context and be cancelled that way.  Even if we had good reason to do a context cancellation check prior to each iteration, we'd outdent the body of `default`.

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,272 @@
...
+			errChan := make(chan error, numChunks)
+
+			for i := int64(0); i < numChunks; i++ {
+				wg.Add(1)
+				go func(chunkIdx int64) {
+					for {
+						// Load the changeset from the admin server
+						cs, err := adminClient.LoadChangeset(ctx, currentHeight, chunkIdx)
+						if err != nil {
+							// If no changeset is found, wait and try again
```
jchappelow, line 133: What if `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)` here?

Also, a timeout on each RPC that downloads a lot should get a timeout as connections have been known to hang.

internal/sql/pg/repl_changeset.go
```diff
@@ -1,29 +1,253 @@
...
-	metadata  *changesetMetadata
-	oidToType map[uint32]*datatype
+	metadata  *changesetMetadata   // reset at end of each commit, builds new list of relations for each db tx
+	oidToType map[uint32]*datatype // immutable map of OIDs to Kwil data types
 
-	writer io.Writer
+	// the io.Writer is multipurpose:
+	//  - for capturing and transmitting the changeset in long running network migrations
+	//  - updating table statistics required for query cost
+	// writer io.Writer
```
jchappelow, line 23: Can remove.  These comments were just postulation.

internal/sql/pg/repl_changeset.go
```diff
@@ -1,29 +1,253 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = tx.Execute(ctx, insertSql, newVals...)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
jchappelow, line 165:
```suggestion
	_, err = tx.Execute(ctx, insertSql, newVals...)
	return err
```

internal/sql/pg/repl_changeset.go
```diff
@@ -1,29 +1,253 @@
...
+func decodeTuple(cols []*TupleColumn, relation *Relation) ([]any, error) {
+	if cols == nil {
+		return nil, nil
+	}
+
+	values := make([]any, len(cols))
+	for i, col := range cols {
+		switch col.ValueType {
+		case NullValue, ToastValue:
+			values[i] = nil
```
jchappelow, line 123: This is OK, but I omitted it because it's superfluous.  `values := make([]any, len(cols))` puts the zero value in each element on construction. (nil)

internal/sql/pg/repl_changeset.go
```diff
@@ -1,29 +1,253 @@
...
+}
+
+func (c *ChangesetEntry) applyInserts(ctx context.Context, tx sql.DB, rel *Relation) error {
+	var columnStr, placeholderStr string
+	if len(rel.Columns) > 0 {
+		columnStr = rel.Columns[0].Name
+		placeholderStr = "$1"
+		for i := 1; i < len(rel.Columns); i++ {
+			columnStr += ", " + rel.Columns[i].Name
+			placeholderStr += ", $" + fmt.Sprint(i+1)
```
jchappelow, line 147: Instead of `fmt.Sprint`, use `strconv.FormatUint` or `strconv.Itoa`.

internal/sql/pg/repl_changeset.go
```diff
@@ -1,29 +1,253 @@
...
+
+			values[i] = val
+		}
+	}
+	return values, nil
+}
+
+func (c *ChangesetEntry) applyInserts(ctx context.Context, tx sql.DB, rel *Relation) error {
+	var columnStr, placeholderStr string
+	if len(rel.Columns) > 0 {
```
jchappelow, line 142: Is it not an error if zero length?  I'm pretty sure insertSql will fail if so.

internal/sql/pg/repl_changeset.go
```diff
@@ -1,29 +1,253 @@
...
+
+	whereClause := ""
+	var args []any
+	cnt := 1
+	for i, v := range record {
+		if i > 0 {
+			whereClause += " AND "
+		}
+		if v == nil {
+			whereClause += fmt.Sprintf("%s IS NULL", rel.Columns[i].Name)
```
jchappelow, line 236: Normally this suggestion would be a micro optimization, but these methods are likely to be processing a significant amount of data: Prefer `var whereClause strings.Builder` and do `whereClause.WriteString(" AND ")` and `fmt.Fprintf(&whereClause, "%s IS NULL", rel.Columns[i].Name)` etc.

jchappelow (commented):
2/n Lots of comments, but this is looking quite nice!

internal/sql/pg/repl_changeset.go
jchappelow: BTW, have we tested with TOASTed values.  By default that's records over 4096, but we should also try with really large values, particularly for the "old" tups in update or delete.

jchappelow: I hope that with replica identity full we get the full values, but I don't know.

internal/migrations/changesets.go
```diff
@@ -0,0 +1,319 @@
...
+		if err != nil {
+			return err
+		}
+		defer tx.Rollback(ctx)
+
+		var currentHeight int64
+		// get the last changeset height applied
+		lastChangeset, err := getLastChangeset(ctx, tx)
+		if err != nil {
+			return tx.Commit(ctx)
```
jchappelow, line 127: Why `return tx.Commit(ctx)` on error?

internal/migrations/changesets.go
```diff
@@ -0,0 +1,319 @@
...
+
+			app.Service.Logger.Info("Applied changesets", log.Int("height", currentHeight))
+
+			// Delete the changeset after it has been applied
+			if err = deleteChangesets(ctx, tx, currentHeight); err != nil {
+				return err
+			}
+
+			// Increment the last changeset
+			if err = setLastChangeset(ctx, tx, currentHeight); err != nil {
```
jchappelow, line 170: I don't understand why "last applied changeset" is a thing to track at all.  This loop and the transaction it uses is supposed to make the application of the changesets all-or-none, right?  Any early return rolls it all back.

charithabandi: This loop tries to apply all the block heights for which all the chunks have been received. That's why we need to track the last applied block height

internal/abci/abci.go
```diff
@@ -177,9 +236,20 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
 		app.consensusParams.WithoutGasCosts = networkParams.DisabledGasCosts
 	}
 
+	networkParams.InMigration = app.migrator.InMigration(height)
+
+	var migrationParams *common.MigrationContext
+	if app.consensusParams.Migration.StartHeight != -1 && app.consensusParams.Migration.EndHeight != -1 {
+		migrationParams = &common.MigrationContext{
+			StartHeight: app.consensusParams.Migration.StartHeight,
```
brennanjl, line 244: Maybe error out if one of them is not -1 and the other is -1? This would suggest an error in the genesis file.

internal/migrations/changeset_listener.go
```diff
@@ -0,0 +1,323 @@
...
+func Start(ctx context.Context, service *common.Service, eventStore listeners.EventStore) error {
+	// Get the migration config from the service
+	cfg := &MigrationConfig{}
+	migrationCfg, ok := service.ExtensionConfigs[ListenerName]
+	if !ok {
+		service.Logger.Warn("no migration config provided, skipping migration listener")
+		return nil // no migration config, nothing to do
+	}
+
+	err := cfg.ExtractConfig(migrationCfg)
```
brennanjl, line 74: Is it possible to remove the need to configure a `start_height` and `end_height` by either:

a. passing the genesis params to `*common.Service`
or
b. querying the configured `listen_address`?

internal/migrations/migrator.go
```diff
@@ -54,30 +108,76 @@ type activeMigration struct {
 	ChainID string
 }
 
-// TODO: when we implement the constructor, we need to add a note that the snapshotter should
-// not be the normal snapshotter, but instead its own instance. This is needed to ensure we do not
-// delete the migration snapshot.
+func SetupMigrator(ctx context.Context, db Database, snapshotter Snapshotter, accounts SpendTracker, dir string, logger log.Logger) (*Migrator, error) {
+	// Set the migrator declared in migrations.go
+	migrator.snapshotter = snapshotter
```
brennanjl, line 113: For future safety, can you return an error here if this is called more than once? Since this is used to initialize the singleton, _but_ it is also returned like a constructor, we should make sure people don't treat it as a constructor that can be called many times

internal/migrations/migrations.go
```diff
@@ -62,42 +75,42 @@ var MigrationResolution = resolutions.ResolutionConfig{
...
-		if err != nil {
-			return err
-		}
-
-		block.ChainContext.NetworkParameters.InMigration = true
-
-		// TODO: there are certainly other things we need to do on activation. I am unsure how to handle this.
-		// For example, we need to snapshot the network at the activation block.
-		return nil
+		return migrator.startMigration(ctx, app, resolution, block)
```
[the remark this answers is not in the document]
brennanjl: @charithabandi ?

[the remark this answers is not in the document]
brennanjl: Never mind I am blind

jchappelow (dismissed):
LGTM.  I only did a very cursory review after the latest iteration.  Best to not squash post-review until review has seen the incremental changes.  NBD though.

jchappelow (approved):

PR #889, opened 2024-07-15
feat: raw JOIN defaults to INNER JOIN

brennanjl:
This PR adds support for raw JOIN, It will be parsed as an inner join, which is the same as how Postgres would treat this statement:

```sql
select * from tbl1 JOIN tbl2 on tbl1.id = tbl2.id
# is the same as
select * from tbl1 INNER JOIN tbl2 on tbl1.id = tbl2.id
```

I added this because I have seen a few different people run into this. Since it ends up being an ANTLR error, it isn't very helpful, and people tend to get stuck here (@MicBun ran into this here https://github.com/truflation/tsn/blob/3e24b3f455b28959d6d2c2756f54e68a8f2367b9/internal/contracts/primitive_stream_template.kf#L369-L370, however Luis ran into it as well a while back).


jchappelow (approved):
LGTM, although I'm curious what's the consideration for setting the type as inner vs adding a new default type and just allowing unspecified?  For instance `generate.WriteSQL` will insert INNER for seemingly no reason (as far as postgres is concerned).  Disambiguation and standardization seems like a good enough reason.  Just wondering.

Yaiba (approved):
LGTM

brennanjl:
> LGTM, although I'm curious what's the consideration for setting the type as inner vs adding a new default type and just allowing unspecified? For instance generate.WriteSQL will insert INNER for seemingly no reason (as far as postgres is concerned). Disambiguation and standardization seems like a good enough reason. Just wondering.

Mostly just for clarity. IMO, allowing an unspecified JOIN feels like more of a parser trick for convenience rather than something that has ramifications for downstream processes, so I figured it is just easier.

PR #891, opened 2024-07-15
chore: bump security policy for v0.8

KwilLuke:
Updating the security policy now that v0.8 is the official latest release.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -5,8 +5,8 @@
 | Version | Supported |
 | ------- | --------- |
 | Latest beta or release candidate    | ✅        |
-| v0.6.x  | ✅        |
-| < v0.6  | ❌        |
+| v0.8.x  | ✅        |
+| < v0.8  | ❌        |
 
 ## Reporting a Vulnerability
 


SECURITY.md
```diff
@@ -5,8 +5,8 @@
 | Version | Supported |
 | ------- | --------- |
 | Latest beta or release candidate    | ✅        |
-| v0.6.x  | ✅        |
-| < v0.6  | ❌        |
+| v0.8.x  | ✅        |
```
jchappelow, line 8: Should this be showing support for 0.7.x and 0.8.x?

```markdown
| Version | Supported |
| ------- | --------- |
| Latest beta or release candidate    | ✓        |
| v0.8.x  | ✓        |
| v0.7.x  | ✓        |
| other | ❌        |
```

KwilLuke: @jchappelow when we first added the security policy (here: https://github.com/kwilteam/kwil-db/pull/483) the coverage was only the latest stable release + any pre-releases advertised as stable(ish) (i.e. beta/rc). Should the security coverage include version n-1?

jchappelow: Oh, sorry maybe not 0.7.  We had talked about supporting it, but seem like all partners are migrating or have migrated.  We probably don't want the burden.

brennanjl: I agree, since all partners are upgraded to v0.8, I think we can say we do not support v0.7 anymore. If someone had an issue, we'd tell them to just use v0.8.

KwilLuke: Great, do one of you want click approve? 😄

jchappelow (approved):

PR #893, opened 2024-07-18
add parse_unix_timestamp

brennanjl:
Adds 2 functions:
- `parse_unix_timestamp(text, text) -> decimal(16,6)`: parses a datetime string to a UNIX timestamp. The first argument is the datetime string, the second is the formatting.
- `format_unix_timestamp(decimal(16,6), text) -> text`: formats a UNIX timestamp as a datetime string. The first argument is the UNIX timestamp, the second is the formatting for the result.

The formatting rules match Postgres's `to_timestamp` formatting: https://www.postgresql.org/docs/current/functions-formatting.html

This was requested by Truflation, and should be enough until we support a timestamp data type. It ended up being way easier to implement than I expected.

### DIFF
--- a/internal/sql/pg/sql.go
+++ b/internal/sql/pg/sql.go
@@ -62,6 +62,18 @@ END$$;`
 		WHEN duplicate_object THEN null;
 	END $$;`
 
+	// postgres returns EXTRACT as a double precision, but will only at most have 6
+	// decimal places of precision (to measure microseconds). We cast to numeric(16, 6)
+	// which should allow for up to 6 decimal places of precision.
+	// Since max unix timestamp is 2147483648, we can cast to numeric(16, 6) to allow 10 digits
+	// before the decimal and 6 after.
+	sqlCreateParseUnixTimestampFunc = `CREATE OR REPLACE FUNCTION parse_unix_timestamp(timestamp_string text, format_string text)
+	RETURNS NUMERIC(16, 6) AS $$
+	BEGIN
+		RETURN EXTRACT(EPOCH FROM TO_TIMESTAMP(timestamp_string, format_string))::numeric(16, 6);
+	END;
+	$$ LANGUAGE plpgsql;`
+
 	sqlCreateOrReplaceReplicaIdentity = `CREATE OR REPLACE FUNCTION set_replica_identity()
 RETURNS event_trigger
 LANGUAGE plpgsql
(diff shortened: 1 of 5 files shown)


brennanjl:
@outerlook @micbun do you think this will work for you?

internal/sql/pg/sql.go
```diff
@@ -62,6 +62,26 @@ END$$;`
...
 
+	// postgres returns EXTRACT as a double precision, but will only at most have 6
+	// decimal places of precision (to measure microseconds). We cast to numeric(16, 6)
+	// which should allow for up to 6 decimal places of precision.
+	// Since max unix timestamp is 2147483648, we can cast to numeric(16, 6) to allow 10 digits
+	// before the decimal and 6 after.
+	sqlCreateParseUnixTimestampFunc = `CREATE OR REPLACE FUNCTION parse_unix_timestamp(timestamp_string text, format_string text)
+	RETURNS NUMERIC(16, 6) AS $$
+	BEGIN
+		RETURN EXTRACT(EPOCH FROM TO_TIMESTAMP(timestamp_string, format_string))::numeric(16, 6);
```
jchappelow, line 73: It seems postgresql is very conservative about lose of information, so my main concern is largely addressed.  With `.US` in the format template:

```
psql://postgres@[local]:5432/postgres 
# select to_timestamp('2024-06-11 13:54:12.123456', 'YYYY-MM-DD HH24:MI:SS.US');
┌───────────────────────────────┐
│         to_timestamp          │
├───────────────────────────────┤
│ 2024-06-11 13:54:12.123456-05 │
└───────────────────────────────┘
(1 row)

Time: 0.442 ms
```

Same time string but with `.MS` in the template instead:

```
psql://postgres@[local]:5432/postgres 
# select to_timestamp('2024-06-11 13:54:12.123456', 'YYYY-MM-DD HH24:MI:SS.MS');
ERROR:  date/time field value out of range: "2024-06-11 13:54:12.123456"
Time: 0.245 ms
```

So it won't toss out any information, and we don't have to worry about whether it truncates or rounds or whatever... as long as it always errors like this.

I think you mentioned this in the meeting but I'm not certain.

The other thing to note is that we could go this route with trade-offs:

```
# select '2024-06-11 13:54:12.123456'::timestamptz;
┌───────────────────────────────┐
│          timestamptz          │
├───────────────────────────────┤
│ 2024-06-11 13:54:12.123456-05 │
└───────────────────────────────┘
(1 row)

Time: 0.423 ms
```

With this, there's no template to specify at all.  The trade-off is that while you don't get the possibly unexpected out-of-range error as easily if you've neglected something in your template, it only works with "most standard date/time formats" formats and that's probably a bigger issue at that could change and bite us.

jchappelow (approved):

PR #894, opened 2024-07-19
fix decimal cmp method

brennanjl:
Fixes a weird aspect of the decimal `Cmp` method where it modifies one of the decimals.


jchappelow (approved):
LGTM.  I'm still uncertain about NaNs, which is presently the only difference between the package-level `decimal.Cmp(a, b)` the method `a.Cmp(b)`.  Regardless, I think this change is right and it's what I have in my other branch.

PR #898, opened 2024-07-25
fix: extensions with sql whitespace fail (#897)

brennanjl:
backports https://github.com/kwilteam/kwil-db/pull/897

### DIFF
--- a/parse/parse.go
+++ b/parse/parse.go
@@ -282,6 +282,10 @@ func ParseSQL(sql string, schema *types.Schema) (res *SQLParseResult, err error)
 	if sql == "" {
 		return nil, fmt.Errorf("empty SQL statement")
 	}
+
+	// trim whitespace
+	sql = strings.TrimSpace(sql)
+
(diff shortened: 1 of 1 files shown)


jchappelow (approved):
Makes sense.  It does break consensus though, if a chain had deploy failures on account of trailing white space after a semicolon.

How important is it to include this change in behavior on 0.8?  The workaround for 0.8 users it to remove the whitespace themselves before deploy, right?

jchappelow:
Also, what's the relevance of extensions?  Is that mentioned because our code would be parsing kuneiform rather than directly sql directly, and thus would not be hit?

brennanjl:
> Also, what's the relevance of extensions? Is that mentioned because our code would be parsing kuneiform rather than directly sql directly, and thus would not be hit unless an extension were using engine.Execute directly in the manner of an ad hoc query?

Exactly

> Makes sense. It does break consensus though, if a chain had deploy failures on account of trailing white space after a semicolon.

Hmm that is true. I think I am inclined to not backport this then.

PR #899, opened 2024-07-29
use go 1.22

Yaiba:
This pr set go version to 1.22 in go.mod, to avoid unexpected 'backward compatiable' behavior,  learn more at https://go.dev/doc/godebug.

I encountered this unexpected behavior in kgw https://github.com/kwilteam/kgw/pull/78#issuecomment-2257163435

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/kwilteam/kwil-db
 
-go 1.21.0
+go 1.22
 
 replace (
 	github.com/kwilteam/kwil-db/core => ./core


jchappelow (commented):
Normally I would say that we should not do this since we aim to support `latest` and `latest - 1` versions of Go. (documented here https://github.com/kwilteam/kwil-db/blob/main/README.md#quickstart)
However, Go 1.22 is slated for release in August 2024, and I'm pretty sure that will beat Kwil DB 0.9.
In general, we should support `latest - 1` with workarounds as required (it is rare).

Yaiba:
>  (documented here https://github.com/kwilteam/kwil-db/blob/main/README.md#quickstart)

ah didn't think of that,  if so should I just close this issue?  
Or, should we provide a 'task' that make sure go toolchain use either 1.22 or 1.21 ?   Right now all our dev/staging/prod using go 1.22, so it's build in compatiable mode

Yaiba:
OK just confirmed that in kwild we don't have compatible behavior(cmd returns empty):
```
$ go list -f '{{.DefaultGODEBUG}}' cmd/kwild/main.go
$ go list -f '{{.DefaultGODEBUG}}' cmd/kwil-cli/main.go
$ go list -f '{{.DefaultGODEBUG}}' cmd/kwil-admin/main.go
```

jchappelow:
That's good.

In general though we should aim to support latest minus 1, so that would mean latest builds running with any backward compatible behaviors.  If there's a behavior we don't want, it's best to make code changes to work around it instead of hoisting the required version.

Issue #800, opened 2024-06-07
provide an advanced DB interface for internals


jchappelow:
The current DB methods scan in all rows at once and into `[][]any` since when dealing with user-authored queries we have no idea what concrete Go type to *scan* into.

However, there are many cases in internal code, including bookkeeping business of the engine, where it would be more straightforward and less error prone (we pick the type) if there were methods for this purpose.

This could be accomplished with a new Execish method that takes scan targets, or it could be done the traditional way with a row iterator (next, scan, err, etc).

jchappelow:
A specific case where this is problematic is with the `uint256` type which is domain and posgresql claims the OID of the underlying type when scanning it out.

jchappelow:
Related note, I've been making functions and methods to do said instantiation and scanning to support statistics collections.  We could expose some of that or even have the engine use that approach rather than relying so heavily on the pgx type map and then converting the `[]any` from `row.Values()`.

Related comment on the issue of efficiency: https://github.com/kwilteam/kwil-db/pull/837#discussion_r1645167702

In short, I indented these "advanced" features with caller instantiated scan variables to support internals like our stores, but user datasets could go this route if the engine instantiates the scan variables, typed based on the schema or just the column info from `information_schema.columns` when the select/projection is a simple column reference (otherwise an `any`).

PR #900, opened 2024-07-30
uint256 type decoding

brennanjl:
I tried to fix the uint256 type decoding issue, but unfortunately it appears to not really be possible. As noted [here](<https://www.postgresql.org/message-id/87fvoydtxx.fsf%40locaine.bese.it>), this is just part of Postgres.

Seems the only way to solve for this is with https://github.com/kwilteam/kwil-db/issues/800#issuecomment-2256723334

This PR removes one completely useless test, and changes query results to only read OIDs once per query, instead of once per row

### DIFF
--- a/internal/sql/pg/query.go
+++ b/internal/sql/pg/query.go
@@ -180,6 +180,7 @@ func query(ctx context.Context, oidToDataType map[uint32]*datatype, cq connQuery
 
 	// res := rows.CommandTag() // RowsAffected, bool for Select etc.
 	resSet := &sql.ResultSet{}
+	var oids []uint32
 	for _, colInfo := range rows.FieldDescriptions() {
 		// fmt.Println(colInfo.DataTypeOID, colInfo.DataTypeSize)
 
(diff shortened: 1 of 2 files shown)


internal/sql/pg/query.go
```diff
@@ -180,6 +180,7 @@ func query(ctx context.Context, oidToDataType map[uint32]*datatype, cq connQuery
 
 	// res := rows.CommandTag() // RowsAffected, bool for Select etc.
 	resSet := &sql.ResultSet{}
+	var oids []uint32
 	for _, colInfo := range rows.FieldDescriptions() {
 		// fmt.Println(colInfo.DataTypeOID, colInfo.DataTypeSize)
```
charithabandi, line 185: remove this comment.

charithabandi (approved):

internal/sql/pg/query.go
jchappelow: Will do in follow up that's touching this code

jchappelow (approved):

PR #603, opened 2024-03-13
[WIP] Feature/sql exec cost estimate

Yaiba:
This pr implements simple exec cost estimation feature, by doing following steps:
1. convert AST to logical plan
2. do simple logical plan optimization
3. convert logical plan to a virtual plan based on statics
4. calculate a cost from a virtual plan

Issue #410, opened 2023-12-04
query cost, gas, and fees


jchappelow:
See https://github.com/kwilteam/kwil-db/wiki/Kwil-v0.9 for v0.9 goals w.r.t. cost and pricing.
This issue breaks down into roughly a few main hurdles:

- Creating a logical query plan from SQL
- Given some table statistics create a virtual plan that can compute a unit-less cost estimate.
- Collecting and maintaining statistics (e.g. row count, MVCs, min/max/range, etc.). Full scan vs incremental.  Incremental updates using changeset info.
- For procedures
  - The wiki for 0.9 says "Therefore, we should use Postgres extensions to allow the procedural language to run with an interpreter that can track the gas while it executes."  This option refers to a postgresql extension that takes the place of pl/pgSQL, and we would be responsible for all flow control. @brennanjl Would you expand on the likely mechanics w.r.t. the pl ast and implementing an interpreter?  This sounds compelling, but had not even considered this all encompassing approach.
  - A different approach might keep using pl/pgSQL but would use an extension with functions to compute cost of each query *just before* executing them.    This may also to require postgresql tricks such as triggers, an audit table, or a novel use of a FWD (a special standard extension called a **f**oreign **d**ata **w**rapper).
  - In terms of detail, other reqs: the engine needs to know total used gas, postgresql or at least the procedure needs to be able to cancel in the event that the gas limit is reached, ...
- In the application (abci/txApp/engine), redesign execution to involve progressive reduction of an allotted gas amount (either from the tx or the user's balance)

## tasks

 - [ ] update query_planner for insert/update/delete.  **Currently only select is supported**. **In progress.**
 - [ ] support CTEs and subqueries.  buildTableSource needs to consider these from field of PlannerContext e.g. GetCTE
 - [ ] complete `costmodel` functions for computing cost from `LogicalPlan`, similar to virtual_plan pkg.  This is now working, but not completely accurate.  Selectivity needs to be computed to reduce cost.  Expression cost is now computed.  Accessing the scan/from plan's statistics in child plans is ready and will inform selectivity computations, which is most important in a join where one table's size can become small due to filter.  **In progress.**
 - [ ] implement a Catalog (GetDataSourcer for qp.NewPlanner) at the level of the engine, with the actual DB and any stored statistics to avoid full rescan. **In progress.**
 - [ ] wire up engine and txApp/abci with new progressive transaction cost from the logical plans.
 - [ ] collect and maintain statistics
[...]

PR #901, opened 2024-07-30
pg: advanced query feats, table inspection, and stats

### LINKED ISSUE #800: provide an advanced DB interface for internals

jchappelow:
This at least partially resolves #800, and provides some foundational components for table statistics collection (https://github.com/kwilteam/kwil-db/issues/410).  I have more work done on stats in both engine and pg for incremental updates, but the pieces in this PR were substantial, so I pulled them out for easier review.

This adds functionality to the `pg` package:

- Table inspection: the `ColumnInfo` function and the `ColInfo` and `ColType` types provide a way to inspect any postgresql table without knowledge of the scheme. The `(*ColInfo).ScanVal` method provides an instance of a suitable type into which a column expression may be scanned in one of the new advanced query methods described below.
- Advanced query execution:
    * the `sql.QueryScanner` interface is the advanced version of `Execute` that uses caller-provided scan values and a function to run for each scanned row:

        ```go
        // QueryScanner represents a type that provides the ability to execute an SQL
        // statement, where for each row:
        //
        //  1. result values are scanned into the variables in the scans slice
        //  2. the provided function is then called
        //
        // The function would typically capture the variables in the scans slice,
        // allowing it to operator on the values. For instance, append the values to
        // slices allocated by the caller, or perform reduction operations like
        // sum/mean/min/etc.
        //
        // NOTE: This method may end up being included in the Tx interface alongside
        // Executor since all of the concrete transaction implementations provided by
        // this package implement this method.
        type QueryScanner interface {
        	QueryScanFn(ctx context.Context, stmt string,
        		scans []any, fn func() error, args ...any) error
        }
        ```

    * each transaction type in the `pg` package satisfies the `sql.QueryScanner` interface
    * the `pg.QueryRowFunc` function executes an SQL statement, handling the rows and returned values as described by the `sql.QueryScanner` interface.
    * the `pg.QueryRowFuncAny` is similar to `pg.QueryRowFunc`, except that no scan values slice is provided. The provided function is called for each row of the result. The caller does not determine the types of the Go variables in the values slice. In this way it behaves similar to `Execute`, but providing "for each row" semantics so that every row does not need to be loaded into memory.

- Table statistics collection: beginning with a simplified `sql.Statistics` struct based on the types proposed in https://github.com/kwilteam/kwil-db/pull/603, the `pg` package provides the following new methods aimed at the (relatively expensive) collection of ground truth table statistics:
    * `RowCount` provides an exact row count
    * `colStats` computes column-wise statistics
    * `TableStats` uses the above functions to build a `*sql.Statistics` for a table.
    
    These methods are will not be used routinely. We will have incremental updates, but there are cases where a full scan may be needed to obtain the ground truth statistics.
- Use the `pgNumericToDecimal` helper to reuse the logic to convert from `pgtypes.Numeric` to either our `decimal.Decimal` or `types.Uint256` in the recent pgtype decoding added to the `query` helper for interpreting the values returned by `row.Values()` in `pgx.CollectRows`.

Note that this PR has three separate commits for easier review, but they can be squashed when merged.

### DIFF
--- a/common/sql/sql.go
+++ b/common/sql/sql.go
@@ -35,6 +35,25 @@ type Executor interface {
 	Execute(ctx context.Context, stmt string, args ...any) (*ResultSet, error)
 }
 
+// QueryScanner represents a type that provides the ability to execute an SQL
+// statement, where for each row:
+//
+//  1. result values are scanned into the variables in the scans slice
+//  2. the provided function is then called
+//
+// The function would typically capture the variables in the scans slice,
+// allowing it to operator on the values. For instance, append the values to
+// slices allocated by the caller, or perform reduction operations like
+// sum/mean/min/etc.
+//
+// NOTE: This method may end up being included in the Tx interface alongside
+// Executor since all of the concrete transaction implementations provided by
+// this package implement this method.
+type QueryScanner interface {
+	QueryScanFn(ctx context.Context, stmt string,
+		scans []any, fn func() error, args ...any) error
+}
+
 // TxMaker is an interface that creates a new transaction. In the context of the
 // recursive Tx interface, is creates a nested transaction.
 type TxMaker interface {
--- a/core/types/uint256.go
+++ b/core/types/uint256.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"database/sql"
 	"database/sql/driver"
 	"fmt"
 	"math/big"
@@ -12,12 +13,13 @@ import (
 // It is mostly a wrapper around github.com/holiman/uint256.Int, but includes
 // extra methods for usage in Postgres.
 type Uint256 struct {
-	uint256.Int
+	base uint256.Int // not exporting massive method set, which also has params and returns of holiman types
+	Null bool
 }
 
 // Uint256FromInt creates a new Uint256 from an int.
 func Uint256FromInt(i uint64) *Uint256 {
-	return &Uint256{Int: *uint256.NewInt(i)}
+	return &Uint256{base: *uint256.NewInt(i)}
 }
 
 // Uint256FromString creates a new Uint256 from a string.
@@ -26,7 +28,7 @@ func Uint256FromString(s string) (*Uint256, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &Uint256{Int: *i}, nil
+	return &Uint256{base: *i}, nil
 }
 
 // Uint256FromBig creates a new Uint256 from a big.Int.
@@ -40,8 +42,33 @@ func Uint256FromBytes(b []byte) (*Uint256, error) {
 	return Uint256FromBig(bigInt)
 }
 
+func (u Uint256) String() string {
+	return u.base.String()
+}
+
+func (u Uint256) Bytes() []byte {
+	return u.base.Bytes()
+}
+
+func (u Uint256) ToBig() *big.Int {
+	return u.base.ToBig()
+}
+
 func (u Uint256) MarshalJSON() ([]byte, error) {
-	return []byte(u.String()), nil
+	return []byte(u.base.String()), nil // ? json ?
+}
+
+func (u *Uint256) Clone() *Uint256 {
+	v := *u
+	return &v
+}
+
+func (u *Uint256) Cmp(v *Uint256) int {
+	return u.base.Cmp(&v.base)
+}
+
+func CmpUint256(u, v *Uint256) int {
+	return u.Cmp(v)
 }
 
 func (u *Uint256) UnmarshalJSON(b []byte) error {
@@ -50,16 +77,20 @@ func (u *Uint256) UnmarshalJSON(b []byte) error {
 		return err
 	}
 
-	u.Int = u2.Int
+	u.base = u2.base
 	return nil
 }
 
 // Value implements the driver.Valuer interface.
 func (u Uint256) Value() (driver.Value, error) {
+	if u.Null {
+		return nil, nil
+	}
 	return u.String(), nil
 }
 
 var _ driver.Valuer = Uint256{}
+var _ driver.Valuer = (*Uint256)(nil)
 
 // Scan implements the sql.Scanner interface.
 func (u *Uint256) Scan(src interface{}) error {
@@ -70,21 +101,28 @@ func (u *Uint256) Scan(src interface{}) error {
 			return err
 		}
 
-		u.Int = u2.Int
+		u.base = u2.base
+		u.Null = false
+		return nil
+
+	case nil:
+		u.Null = true
+		u.base.Clear()
 		return nil
 	}
 
 	return fmt.Errorf("cannot convert %T to Uint256", src)
 }
 
-var _ driver.Valuer = (*Uint256)(nil)
-var _ driver.Valuer = (*Uint256)(nil)
+var _ sql.Scanner = (*Uint256)(nil)
 
 // Uint256Array is an array of Uint256s.
 type Uint256Array []*Uint256
 
 // Value implements the driver.Valuer interface.
 func (ua Uint256Array) Value() (driver.Value, error) {
+	// Even when implementing pgtype.ArrayGetter we still need this, so that the
+	// pgx driver can use it's wrapSliceEncodePlan.
 	strs := make([]string, len(ua))
 	for i, u := range ua {
 		strs[i] = u.String()
@@ -95,21 +133,8 @@ func (ua Uint256Array) Value() (driver.Value, error) {
 
 var _ driver.Valuer = (*Uint256Array)(nil)
 
-// Scan implements the sql.Scanner interface.
-func (ua *Uint256Array) Scan(src interface{}) error {
-	switch s := src.(type) {
-	case []string:
-		*ua = make(Uint256Array, len(s))
-		for i, str := range s {
-			u, err := Uint256FromString(str)
-			if err != nil {
-				return err
-			}
-
-			(*ua)[i] = u
-		}
-		return nil
-	}
-
-	return fmt.Errorf("cannot convert %T to Uint256Array", src)
-}
+// Uint256Array is a slice of Scanners. pgx at least is smart enough to make
+// this work automatically!
+// Another approach is to implement pgx.ArraySetter and pgx.ArrayGetter like
+// similar in effect to:
+//   type Uint256Array pgtype.FlatArray[*Uint256]
--- a/internal/sql/pg/query.go
+++ b/internal/sql/pg/query.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/kwilteam/kwil-db/common/sql"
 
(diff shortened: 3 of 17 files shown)


internal/sql/pg/system.go
```diff
@@ -198,6 +202,452 @@ var settingValidations = map[string]settingValidFn{
...
+        FROM information_schema.columns
+        WHERE table_name = '` + tbl + `' AND table_schema = '` + schema + `'
+			AND table_catalog = '` + dbName + `'`
+
+	var worked bool
+
+	var pos int
+	var domainName pgtype.Text // null in Valid bool
+	var colName, dataType, typeOrArray, isNullable string
+	var colDefault any
```
brennanjl, line 593: This default is scanned out, but it will probably be pretty confusing to a consumer of this information. In the `ColInfo` struct, it is of type `any`, however a consumer would probably think that it would be one of Kwil's types (e.g. our numeric type). However, a numeric will actually be one of pgx's.

It seems like this (or the `queryRowFunc`) is a place that we could rely on the `*datatype` struct to convert this, using the `(*datatype.Decode())` method.

Alternatively, making the `*pg.ColInfo` struct unexported could make sense. Really, what I am getting at here is that the consumer of this package is left with ambiguity as to whether things are converted to be "Kwil native", or if they are "pgx native"

internal/sql/pg/system.go
```diff
@@ -198,6 +202,452 @@ var settingValidations = map[string]settingValidFn{
 	"server_encoding": wantStringFn("UTF8"),
 }
 
+func queryRowFunc(ctx context.Context, conn *pgx.Conn, sql string,
+	scans []any, fn func() error, args ...any) error {
+	rows, _ := conn.Query(ctx, sql, args...)
+	_, err := pgx.ForEachRow(rows, scans, fn)
```
brennanjl, line 208: We should add better tests for scanning to different types, because I am fairly confident they will not work. For example, I don't think `Uint256Array`'s `Scan()` implementation will work, because it only checks for a `[]string` type, and I would expected pgx would give us either `[]any` or `[]pgtype.UUID`.

internal/sql/pg/system.go
```diff
@@ -198,6 +202,452 @@ var settingValidations = map[string]settingValidFn{
...
+	case conner:
+		conn := ti.Conn()
+		return queryRowFunc(ctx, conn, stmt, scans, fn, args...)
+	}
+	return errors.New("cannot query with scan values")
+}
+
+// FieldDesc describes result value from a query. This is used to convey
+// information on the values passed to the closure given to QueryRowFuncAny.
+type FieldDesc struct {
```
brennanjl, line 232: Unexport? This gives the caller access to the OID, which for the uint256 domain, will be different on every system. Would hate for someone to forget this and rely on this functionality externally later

internal/sql/pg/system.go
```diff
@@ -198,6 +202,452 @@ var settingValidations = map[string]settingValidFn{
...
+	return "", false, false
+}
+
+// ColInfo is used when ingesting column descriptions from PostgreSQL, such as
+// from information_schema.column. Use the Type method to return a canonical
+// ColType. Use the ScanVal method to return a pointer to an instance of the
+// appropriate Go type to scan a row containing this column type in an SQL
+// statement.
+type ColInfo struct {
+	Pos      int
```
brennanjl, line 309: As mentioned in another comment, I think we should do one of two things:

1. Make the data type uniform with Kwil data types (*types.DataType), and make the `Default` a Kwil type as well (instead of a pgx type), since this is what other functions in this package return.
2. Make this struct unexported, in which case it doesn't matter that it doesn't follow the same semantics as the rest of the package.

common/sql/sql.go
```diff
@@ -35,6 +35,25 @@ type Executor interface {
...
+// slices allocated by the caller, or perform reduction operations like
+// sum/mean/min/etc.
+//
+// NOTE: This method may end up being included in the Tx interface alongside
+// Executor since all of the concrete transaction implementations provided by
+// this package implement this method.
+type QueryScanner interface {
+	QueryScanFn(ctx context.Context, stmt string,
+		scans []any, fn func() error, args ...any) error
+}
```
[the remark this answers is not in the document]
jchappelow: > Overall seems find, but what is the reason for not doing something more conventional, similar to the standard library's `sql.Rows` (containing `Next()`, `Scan()`, etc.)?

Basically just no real reason to introduce a rows type.  The for each helpers in the pgx package were also a more intuitive building block for this method.

internal/sql/pg/stats.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+
+	return &sql.Statistics{
+		RowCount:         count,
+		ColumnStatistics: colStats,
+	}, nil
+}
+
+type ColStatser interface {
+	ColStats(ctx context.Context, qualifiedTable string, colInfo []ColInfo) ([]sql.ColumnStatistics, error)
```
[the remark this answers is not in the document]
jchappelow: Yup, I have a branch building on it.

internal/sql/pg/system.go
brennanjl: We could also just get rid of `Default`, if we aren't relying on it

internal/sql/pg/system.go
jchappelow: I agree that exporting this `colDefault` is unhelpful at best and problematic for the reasons you mentioned.  I didn't give this field much thought, really it was just for completeness but it's not helpful outside this package without more constrained types.

I'll almost certainly unexport things rather than cast `ColInfo` and it's fields in terms of Kwil types.  The `*datatype` and it's `Decode` method are designed to address a different problem -- interpreting the `any` from the `row.Values()` return in `pgx.CollectRows`.  In this work we want to dictate the scan value's type so that it's predicatable and easy to work with, no fuss with OIDs and ambiguity about what type pgx might choose if left to its own devices (or postgres' in the case of domains).  The types from the ScanVal method are quite tightly coupled to the (unexported) `colStats`.  ` TableStats` is the only thing that really uses these.

internal/sql/pg/system.go
jchappelow: Commented in https://github.com/kwilteam/kwil-db/pull/901/files/a62d42727c0574985c662cd8ae94390810691ea5#r1699049173, but regarding `ColInfo`, as the doc comment says, this is really `ingesting column descriptions from PostgreSQL`, and the `Type()` method is intended to return the canonical enumerator of the type.  This is fine as it's well defined and just an enum, but things do get problematic with the `ScanVal() any` method as there's really nothing useful a consumer (outside the `pg` package) can do with the value which is just a type that other functions in the `pg` package want/need.  (You're :100: correct to note that most of this table inspection code is designed to power other functionality internal to `pg`.)

internal/sql/pg/system.go
jchappelow: Ah, yes!  So, the motivation is poorly explained, but this pertains to the main similarity between `QueryRowFuncAny` and the current `Execute`/`query` functions, which also use `row.Values()`.  Essentially `QueryRowFuncAny` should behave like `query` but does something for each row rather than loading all rows into memory.

This `FieldDesc` relates to the above since it is an argument to the function that is executed for each row.

So yeah, both `FieldDesc` and `QueryRowFuncAny` in their current forms should be unexported.  Any exported `QueryRowFuncAny` should not trouble the user with OIDs and such, and should ensure the `[]any` contains the same types of values that the exported `Execute` methods return. (needs to use the registered `datatype`s + `decodeFromPG` => the `Decode` methods)

internal/sql/pg/system.go
jchappelow: > We should add better tests for scanning to different types, because I am fairly confident they will not work. For example, I don't think `Uint256Array`'s `Scan()` implementation will work, because it only checks for a `[]string` type, and I would expected pgx would give us either `[]any` or `[]pgtype.UUID`.

Yeah, I will add some tests using the scanners.  But an `sql.Scanner` will always be given a primitive type. That's pretty much the purpose of such a type.  (Note the the types in `pgtypes` are `sql.Scanner` themselves).  pgx is just smart enough to recognize when it's scanning into a `sql.Scanner` and skip past it's own "smart" codecs, so afaik we are sidestepping all that complexity and dealing with lcd types.

Will test this a bit.

For reference, the std lib docs expand on this:

```go
// Scanner is an interface used by [Rows.Scan].
type Scanner interface {
	// Scan assigns a value from a database driver.
	//
	// The src value will be of one of the following types:
	//
	//    int64
	//    float64
	//    bool
	//    []byte
	//    string
	//    time.Time
	//    nil - for NULL values
	//
	// An error should be returned if the value cannot be stored
	// without loss of information.
	//
	// Reference types such as []byte are only valid until the next call to Scan
	// and should not be retained. Their underlying memory is owned by the driver.
	// If retention is necessary, copy their values before the next call to Scan.
	Scan(src any) error
}
```

jchappelow:
Getting uint256 and decimal and their array types scanning and valueing right was way more painful than expected.
There are some incomplete tests I will get to tonight, but this PR is mostly updated according to the review comments now.
The easier update from yesterday is that `QueryRowFuncAny` now converts to kwil types like the existing query that returns all the rows.  Also unexported a bunch of stuff that would not be useful outside of `pg`.

internal/sql/pg/query.go
```diff
@@ -234,3 +235,92 @@ func queryTx(ctx context.Context, oidToDataType map[uint32]*datatype, dbTx txBeg
...
+	for rows.Next() {
+		pgxVals, err := rows.Values()
+		if err != nil {
+			return err
+		}
+
+		// Decode the values into Kwil or native types.
+		decVals := make([]any, len(pgxVals))
+		for i, pgVal := range pgxVals {
+			decVal, err := decodeFromPGVal(pgVal, oids[i], oidTypes)
```
jchappelow, line 302: Since I  moved `queryRowFuncAny` to a new file, I'll call out this key change, which is converting the types to Kwil types according to our oid type map.

internal/sql/pg/system.go
```diff
@@ -624,26 +559,17 @@ func columnInfo(ctx context.Context, conn *pgx.Conn, schema, tbl string) ([]ColI
...
-
 // ColumnInfo attempts to describe the columns of a table in a specified
-// PostgreSQL schema.
+// PostgreSQL schema. The results are **as reported by information_schema.column**.
 //
 // If the provided sql.Executor is also a ColumnInfoer, its ColumnInfo method
 // will be used. This is primarily for testing with a mocked DB transaction.
 // Otherwise, the Executor must be one of the transaction types created by this
 // package, which provide access to the underlying DB connection.
 func ColumnInfo(ctx context.Context, tx sql.Executor, schema, tbl string) ([]ColInfo, error) {
```
jchappelow, line 575: Exported functions don't take any `ColInfo`, only return for general information purposes.  Docs for `ColInfo` elaborate.  This `ColumnInfo` won't be need needed outside of the `pg` packages, but I find it generally useful, so didn't go so far as to unexport it or `ColInfo`, but it wouldn't derail any of the work in engine or elsewhere.

core/types/uint256.go
```diff
@@ -95,21 +133,55 @@ func (ua Uint256Array) Value() (driver.Value, error) {
...
+	if dimensions == nil {
+		*ua = nil
+		return nil
+	}
+	*ua = make(Uint256Array, pgDimCard(dimensions))
+	return nil
+}
+
+func (ua Uint256Array) ScanIndex(i int) any { return &ua[i] }
+
```
brennanjl, line 185: are you still planning / debating to use this commented out code?

brennanjl (dismissed):
other than the commented code lgtm

core/types/uint256.go
jchappelow: Cleaned it out.  PR squashed.

brennanjl (approved):

PR #903, opened 2024-07-31
added @block_timestamp

brennanjl:
I added an `@block_timestamp` variable. Both Truflation and idOS requested this in the last 2 weeks.

We should make sure to note that this can be spoofed, since it is just based off of block time. A less-spoofable version could be added using cometbft's vote extensions, but that is something we can add at a later date.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -92,6 +92,9 @@ type TransactionData struct {
 
 	// Height is the block height of the incoming transaction.
 	Height int64
+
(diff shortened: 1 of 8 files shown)


jchappelow (approved):
LGTM.  Just one thing that I think we should consider with contextual variables is that we must be able to set them not only in `FinalizeBlock` but in `ProcessProposal`.  I checked and `RequestProcessProposal` has hash, height, and time, so I think this is good, assuming that time will be the same that gets sent to `FinalizeBlock` for approved blocks.

PR #905, opened 2024-07-31
feat: dbid util function and cli command

brennanjl:
While putting together an example, I ran into two things I wish I had: a function to generate DBIDs, and a CLI util to generate DBIDs.

### DIFF
--- a/cmd/kwil-cli/cmds/utils/dbid.go
+++ b/cmd/kwil-cli/cmds/utils/dbid.go
@@ -64,11 +64,7 @@ type dbidOutput struct {
 func (d *dbidOutput) MarshalJSON() ([]byte, error) {
 	// Alias is used to avoid infinite recursion when calling json.Marshal
 	type Alias dbidOutput
-	return json.Marshal(&struct {
-		Alias
-	}{
-		Alias: (Alias)(*d),
-	})
+	return json.Marshal((Alias)(*d))
 }
 
 func (d *dbidOutput) MarshalText() (text []byte, err error) {
--- a/cmd/kwil-cli/cmds/utils/dbid.go
+++ b/cmd/kwil-cli/cmds/utils/dbid.go
@@ -1,0 +1,76 @@
+package utils
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/kwilteam/kwil-db/cmd/common/display"
+	"github.com/kwilteam/kwil-db/core/utils"
+	"github.com/spf13/cobra"
+)
+
+var (
+	dbidLong = `"dbid" generates a dbid for a given schema name and deployer.`
+
+	dbidExample = `# Generate a dbid for a schema and deployer
+kwil-cli utils dbid --schema=myschema --deployer=0x1234567890abcdef
+
+# Maintain the exact deployer address, and do not trim off the 0x prefix
+kwil-cli utils dbid --schema=myschema --deployer=0xnot_an_eth_address --no-trim`
+)
+
+func dbidCmd() *cobra.Command {
+	var schema, deployer string
+	var noTrim bool
+
+	cmd := &cobra.Command{
+		Use:     "generate-dbid",
+		Short:   dbidLong,
+		Long:    dbidLong,
+		Example: dbidExample,
+		Args:    cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			deployerBts := []byte(deployer)
+			if strings.HasPrefix(deployer, "0x") && !noTrim {
+				deployer = deployer[2:]
+				var err error
+				deployerBts, err = hex.DecodeString(deployer)
+				if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/kwil-cli/cmds/utils/dbid.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (d *dbidOutput) MarshalJSON() ([]byte, error) {
+	// Alias is used to avoid infinite recursion when calling json.Marshal
+	type Alias dbidOutput
+	return json.Marshal(&struct {
+		Alias
+	}{
+		Alias: (Alias)(*d),
+	})
```
jchappelow, line 71: I'm pretty certain that the anonymous struct embedding `Alias` isn't needed.

```go
	type Alias dbidOutput
	return json.Marshal((Alias)(*d))
```

```go
func Example_dbidOutput_json() {
	display.Print(&dbidOutput{DBID: "asdf"}, nil, "json")
	// Output:
	// {
	//   "result": {
	//     "dbid": "asdf"
	//   },
	//   "error": ""
	// }
}
```

cmd/kwil-cli/cmds/utils/dbid.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			dbid := utils.GenerateDBID(schema, deployerBts)
+			return display.PrintCmd(cmd, &dbidOutput{DBID: dbid})
+		},
+	}
+
+	cmd.Flags().StringVar(&schema, "schema", "", "Schema name")
+	cmd.Flags().StringVar(&deployer, "deployer", "", "Deployer address")
+	cmd.Flags().BoolVar(&noTrim, "no-trim", false, "Maintain the exact deployer address")
```
jchappelow, line 52: Maybe mention that the default is to trim `0x` prefix

jchappelow (approved):
Makes sense.  One comment about the command result marshalling though.

cmd/kwil-cli/cmds/utils/dbid.go
brennanjl: I tested it without it, and I hit an overflow. It is because It is implementing a `MarshalJSON` method, so calling `json.Marshal(d)` would be infinitely recursive.

jchappelow: Right, I understand the recursion issue needs Alias, but you have this anonymous struct in addition to alias.  Can you try with just this?

```go
	type Alias dbidOutput
	return json.Marshal((Alias)(*d))
```

That works for me.  The alias is sufficient to drop the methods of the underlying type.

brennanjl: Oh whoops

jchappelow (approved):

jchappelow:
@Yaiba See any issues with a `generate_dbid` function that spits out the DBID given owner and db *name*.  I think we've pretty well established that DBID is not in internal detail at this point so I don't see the harm, but wanted to get your opinion.

cmd/kwil-cli/cmds/utils/dbid.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+
+			dbid := utils.GenerateDBID(schema, deployerBts)
+			return display.PrintCmd(cmd, &dbidOutput{DBID: dbid})
+		},
+	}
+
+	cmd.Flags().StringVar(&schema, "schema", "", "Schema name")
+	cmd.Flags().StringVar(&deployer, "deployer", "", "Deployer address")
+	cmd.Flags().BoolVar(&noTrim, "no-trim", false, "Do not trim off the 0x prefix of the deployer address")
```
Yaiba, line 52: This name 'no-trim' reminds me a word Luke mentioned before, "double negative".   
Is there a better name we can choose? something like `keep-prefix`?

Yaiba: I think maybe I lost some context,  in what case do we want `0x` in the address?

Yaiba:
> @Yaiba See any issues with a `generate_dbid` function that spits out the DBID given owner and db _name_. I think we've pretty well established that DBID is not an internal detail at this point so I don't see the harm, but wanted to get your opinion.

Yeah agree, I think we can expose this function

cmd/kwil-cli/cmds/utils/dbid.go
brennanjl: I don't think `no-trim` is a double negative. Trimming is something our tooling does by default, but we are telling it not to do that.

> I think maybe I lost some context, in what case do we want 0x in the address?

We identify all users by their ethereum addresses, without the 0x.

Yaiba: > We identify all users by their ethereum addresses, without the 0x

Yeah, so why would user want to generate a DBID with address starts with '0x'?

brennanjl: If they have a custom authenticator that does expect an address to have `0x`.

Yaiba: ah I see.

Yaiba (approved):

PR #908, opened 2024-07-31
fix: generate_dbid() generates syntax error

brennanjl:
The new `generate_dbid` function generates a syntax error if it is used with assignment (`:=`). I fixed this and added a test for it.


Yaiba (approved):
LGTM

PR #910, opened 2024-08-02
Brennan planner

brennanjl:
Putting up for tracking


brennanjl:
Just pushed up an initial version. I am still working on tests for it. It can be found in `parse/planner`. I moved it to parse because a lot of the validation logic that I hacked in the parser's SQL analyzer is better handled with relational algebra.

PR #912, opened 2024-08-07
fix: cte syntax bug

brennanjl:
While testing CTEs within the logical planner, I found a bug where our common table expression syntax required users to specify their return columns. This is typically optional in SQL.

We don't need to backport this since it isn't a security issue, but it was definitely unexpected and a bit confusing (I had to look at our ANTLR grammar to figure out why my CTE wasn't working).


Yaiba (approved):
LGTM

PR #916, opened 2024-08-12
removed unnecessary peer flag

brennanjl:
Removes an incorrectly documented flag `--peers`, as noted [here](<https://github.com/truflation/tsn/issues/393#issuecomment-2283364060>)

### DIFF
--- a/cmd/kwil-admin/cmds/setup/peer.go
+++ b/cmd/kwil-admin/cmds/setup/peer.go
@@ -16,7 +16,7 @@ var (
 It will automatically generate required directories and keypairs, and can be given a genesis file and peer list for an existing network.`
 
 	peerExample = `# Initialize a node as a peer to an existing network
-kwil-admin setup peer --root-dir ./kwil-node --genesis /path/to/genesis.json`
+kwil-admin setup peer --root-dir ./kwil-node --genesis /path/to/genesis.json --chain.p2p.persistent-peers peer1@ip:26656,peer2@ip:26656`
 )
 
 func peerCmd() *cobra.Command {
--- a/cmd/kwil-admin/cmds/setup/peer.go
+++ b/cmd/kwil-admin/cmds/setup/peer.go
@@ -16,7 +16,7 @@ var (
 It will automatically generate required directories and keypairs, and can be given a genesis file and peer list for an existing network.`
 
 	peerExample = `# Initialize a node as a peer to an existing network
-kwil-admin setup peer --root-dir ./kwil-node --genesis /path/to/genesis.json --peers`
+kwil-admin setup peer --root-dir ./kwil-node --genesis /path/to/genesis.json`
 )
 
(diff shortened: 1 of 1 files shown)


cmd/kwil-admin/cmds/setup/peer.go
```diff
@@ -16,7 +16,7 @@ var (
 It will automatically generate required directories and keypairs, and can be given a genesis file and peer list for an existing network.`
```
jchappelow, line 16: "and peer list for an existing network"  is that what the `--peers` flag was intending to do?  Is this part of the command's docs incorrect?

charithabandi (approved):

cmd/kwil-admin/cmds/setup/peer.go
brennanjl: I think so originally, but users should just use the `--chain.p2p.persistent-peers` flag instead. I have updated the example to reflect this

jchappelow:
I had never used this command before, and was surprised I had never noticed it.  So, I ran just `./kwil-admin setup peer` to see what would happen.  Nothing was printed.  It seems to have modified my `~/.kwild/config.toml`, but NOT genesis or anything else.  Is this right?  What should I expect it to have done?

jchappelow (approved):
LGTM although I didn't know about this command for some reason.
I'm wondering why we don't just have a single `setup node` command that is like the current `setup init` but can take a full `--genesis` to use verbatim (which use may have created manually for a new _or_ existing network).

brennanjl:
> LGTM although I didn't know about this command for some reason.
I'm wondering why we don't just have a single setup node command that is like the current setup init but can take a full --genesis to use verbatim (which use may have created manually for a new or existing network).

This is exactly what this command is for. If a user already has a network that they want to connect to, they can use this to configure their node with a pre-set genesis.

jchappelow:
> > LGTM although I didn't know about this command for some reason.
> > I'm wondering why we don't just have a single setup node command that is like the current setup init but can take a full --genesis to use verbatim (which use may have created manually for a new or existing network).
> 
> This is exactly what this command is for. If a user already has a network that they want to connect to, they can use this to configure their node with a pre-set genesis.

Yeah, I'm completely clear on that.  But I don't get why it is it's own command.  The `setup init` command could have had a `--genesis` flag to override all the other piecemeal flags and it would have been identical.  The `setup peer` command name doesn't make sense to me.  If I had a genesis file for a *new* network, I would be confused about which command to use, if any.  Most of the time I just want a default config.toml to go with the file, so based on name `setup peer` is not what I would reach for in that case, but it actually does just that.

brennanjl:
I see, yes that makes sense. This seems to be exactly the issue Truflation is running into.

Why don't we make this change to `setup init`, and we can mark `peer` as deprecated in the next release.

PR #919, opened 2024-08-14
support go 1.23

jchappelow:
This adds support for Go 1.23, including it's corresponding `govet` (at least the one shipped with `golangci-lint` with 1.23 support), which has some new rules we were violating. `task lint` on my machine immediately failed after upgrading my workstation's toolchain. e.g. `printf: non-constant format string in call ... (govet)`

Regarding the `govet` issues, the tldr is that it is almost never correct to do `printf(strVar)` (non-constant format specification with *no* arguments) because if the contents of the string variable contain a "%" character, the result is an error string.  See:
- https://github.com/golang/go/issues/60529
- https://go-review.googlesource.com/c/tools/+/585795

This PR also updates CI, the linter version, the README, and dockerfiles.

IMPORTANT: this does **not** increase the required compiler version from `go 1.21` to `go 1.22` in the `go.mod` files.  Although support policy is Go latest and latest-1 (1.22 and 1.23 now), I haven't yet changed to go.mod to break compilation with Go 1.21.  There are however a few things added in 1.22 that would be nice, such as `reflect.TypeFor[T any]()`, so we might make this change soon, but it's not happening here.


internal/services/grpc_gateway/middleware/cors/cors_test.go
```diff
@@ -5,10 +5,16 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
-
-	http2 "github.com/kwilteam/kwil-db/internal/services/grpc_gateway/middleware/http-dummy"
```
jchappelow, line 9: I figure I'll follow up with a PR to remove all grpc code and protos.  Since 0.9 introduced new RPC functionality on the JSON-RPC service that would be a waste of time to implement on the grpc server (and grpc http gateway), this means we're definitely removing it for 0.9

parse/analyze.go
```diff
@@ -426,21 +426,21 @@ var _ Visitor = (*sqlAnalyzer)(nil)
 // typeErr should be used when a type error is encountered.
 // It returns an unknown attribute and adds an error to the error listener.
 func (s *sqlAnalyzer) typeErr(node Node, t1, t2 *types.DataType) *types.DataType {
-	s.errs.AddErr(node, ErrType, fmt.Sprintf("%s != %s", t1.String(), t2.String()))
+	s.errs.AddErr(node, ErrType, "%s != %s", t1, t2)
```
jchappelow, line 429: There are three "*Err" functions with this `msg string, args ...any` pattern.  Sometimes used, sometimes not.  Honestly it would be better with just `msg string` leaving the caller to format if desired, as was the case in some places but not others, but I am leaving the function signatures unchanged for now.

.golangci.yml
```diff
@@ -13,7 +13,6 @@ issues:
 
 output:
   formats:
-    - format: github-actions
```
jchappelow, line 16: `WARN [config_reader] The output format `github-actions` is deprecated, please use `colored-line-number`  Been the case that github actions supports `colored-line-number` for a while, just now making the changes

jchappelow:
Resolves CI failures on other PRs since CI started pulling the updated linter automatically: https://github.com/kwilteam/kwil-db/actions/runs/10407057042

charithabandi (approved):
LGTM!

PR #922, opened 2024-08-16
fix: arithmetic and equality precedence

brennanjl:
Fixes a bug where equality and assignment operators take a higher precedence than arithmetic operators.

We shouldn't backport this because it will break consensus.


brennanjl:
converted back to draft as I recognized some other precedence issues

parse/grammar/KuneiformParser.g4
```diff
@@ -279,34 +279,41 @@ delete_statement:
...
-    | left=sql_expr (PLUS | MINUS) right=sql_expr                                   # arithmetic_sql_expr
-    // setting precedence for logical operations:
+
+    // the rest:
+    | sql_expr NOT? IN LPAREN (sql_expr_list|select_statement) RPAREN               # in_sql_expr
+    | left=sql_expr NOT? (LIKE|ILIKE) right=sql_expr                                # like_sql_expr
+    | element=sql_expr (NOT)? BETWEEN lower=sql_expr AND upper=sql_expr             # between_sql_expr
+    | left=sql_expr (EQUALS | EQUATE | NEQ | LT | LTE | GT | GTE) right=sql_expr    # comparison_sql_expr
+    | left=sql_expr IS NOT? ((DISTINCT FROM right=sql_expr) | NULL | TRUE | FALSE)  # is_sql_expr
+    | <assoc=right> (NOT) sql_expr                                                  # unary_sql_expr
```
Yaiba, line 311: nitpicking, no need for parentheses

parse/grammar/KuneiformParser.g4
```diff
@@ -279,34 +279,41 @@ delete_statement:
     (WHERE where=sql_expr)?
 ;
 
+// https://docs.kwil.com/docs/kuneiform/operators
 sql_expr:
-    literal type_cast?                                                              # literal_sql_expr
+    // highest precedence:
```
Yaiba, line 284: I suggest we put expr with same shape next to each other, different shape expr has no precedence issues

parse/parse_test.go
```diff
@@ -1952,18 +1973,46 @@ func mustNewDecimal(precision, scale uint16) *types.DataType {
...
+		if isNeg {
+			t *= -1
+		}
+
+		liter := &parse.ExpressionLiteral{
 			Type:  types.IntType,
 			Value: int64(t),
 		}
+
+		if isNeg {
```
Yaiba, line 1989: 👍

Yaiba (approved):
lgtm

parse/grammar/KuneiformParser.g4
```diff
@@ -347,22 +354,29 @@ procedure_block:
...
 ;
 
+// https://docs.kwil.com/docs/kuneiform/operators
 procedure_expr:
-    literal type_cast?                                                                          # literal_procedure_expr
+    // highest precedence:
+    LPAREN procedure_expr RPAREN type_cast?                                                     # paren_procedure_expr
+    | procedure_expr PERIOD IDENTIFIER type_cast?                                               # field_access_procedure_expr
+    | procedure_expr LBRACKET procedure_expr RBRACKET type_cast?                                # array_access_procedure_expr
+    | <assoc=right> (PLUS|MINUS|EXCL) procedure_expr                                            # unary_procedure_expr
```
Yaiba, line 363: is `!323` and `!age` valid in Postgres?

Yaiba: wait I don't see `!` in the operator table https://www.postgresql.org/docs/16/sql-syntax-lexical.html#SQL-PRECEDENCE-TABLE

brennanjl: No but we translate it to `not` when we generate code. Also it is only in procedures.

parse/grammar/KuneiformParser.g4
brennanjl: It's effectively the same thing

parse/grammar/KuneiformParser.g4
Yaiba: OIC

PR #923, opened 2024-08-19
Peer filtering based on the whitelisted peers

charithabandi:
Each node can now run in private mode which accepts connections from:
- the current validator set 
- persistent peers (`chain.p2p.persistent-peers`)
- seeds (`chain.p2p.seeds`)
- whitelisted peers  (`chain.p2p.whitelist-peers`)
Connections from  other sentry nodes can be allowed, either updating whitelist peers config and restart the node or by using `kwil-admin peers add/remove` commands.

When private mode is disabled, the node can accept the connections from any nodes.

### DIFF
--- a/cmd/kwild/config/config.go
+++ b/cmd/kwild/config/config.go
@@ -119,7 +119,11 @@ type P2PConfig struct {
 	// PersistentPeers is a comma separated list of nodes to keep persistent
 	// connections to.
 	PersistentPeers string `mapstructure:"persistent_peers"`
-	// WhitelistPeers is a comma seperated list of nodeIDs that can connect to this node.
+	// PrivateMode prevents other nodes from connecting to the node unless
+	// they are the current validators or a part of the whitelistPeers.
+	// If disabled, the node by default operates in public mode, where any node can connect to it.
+	PrivateMode bool `mapstructure:"private_mode"`
+	// WhitelistPeers is a comma separated list of nodeIDs that can connect to this node.
 	// This is excluding any persistent peers or seeds or current validators.
 	WhitelistPeers string `mapstructure:"whitelist_peers"`
 	// AddrBookStrict enforces strict address routability rules. This must be
@@ -595,6 +599,7 @@ func DefaultConfig() *KwildConfig {
 			P2P: &P2PConfig{
 				ListenAddress:       "tcp://0.0.0.0:26656",
 				ExternalAddress:     "",
+				PrivateMode:         false,
 				AddrBookStrict:      false, // override comet
 				MaxNumInboundPeers:  40,
 				MaxNumOutboundPeers: 10,
--- a/cmd/kwild/config/config.go
+++ b/cmd/kwild/config/config.go
@@ -119,6 +119,9 @@ type P2PConfig struct {
 	// PersistentPeers is a comma separated list of nodes to keep persistent
 	// connections to.
 	PersistentPeers string `mapstructure:"persistent_peers"`
+	// WhitelistPeers is a comma seperated list of nodeIDs that can connect to this node.
+	// This is excluding any persistent peers or seeds or current validators.
+	WhitelistPeers string `mapstructure:"whitelist_peers"`
 	// AddrBookStrict enforces strict address routability rules. This must be
 	// false for private or local networks.
 	AddrBookStrict bool `mapstructure:"addr_book_strict"`
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -695,8 +695,8 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 	}
 
 	// Join requests approved by this node are added to the peer list.
-	for _, join := range approvedJoins {
-		addr, err := cometbft.PubkeyToAddr(join.PubKey)
+	for _, pubKey := range approvedJoins {
+		addr, err := cometbft.PubkeyToAddr(pubKey)
 		if err != nil {
 			return nil, fmt.Errorf("failed to convert pubkey to address: %w", err)
 		}
@@ -708,8 +708,8 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 	}
 
 	// peers whose join requests have expired are removed from the peer list
-	for _, join := range expiredJoins {
-		addr, err := cometbft.PubkeyToAddr(join.PubKey)
+	for _, pubKey := range expiredJoins {
+		addr, err := cometbft.PubkeyToAddr(pubKey)
 		if err != nil {
 			return nil, fmt.Errorf("failed to convert pubkey to address: %w", err)
 		}
@@ -1500,6 +1500,7 @@ func (a *AbciApp) Query(ctx context.Context, req *abciTypes.RequestQuery) (*abci
 		switch paths[0] {
 		case "id":
 			if a.p2p.IsPeerWhitelisted(paths[1]) {
+				a.log.Info("Connection attempt accepted, peer is allowed to connect", zap.String("peerID", paths[1]))
 				return &abciTypes.ResponseQuery{Code: abciTypes.CodeTypeOK}, nil
 			}
 			// ID is not in the allowed list of peers, so reject the connection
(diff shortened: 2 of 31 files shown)


internal/abci/cometbft/peers.go
```diff
@@ -1,10 +1,183 @@
...
+	"sync"
+
 	"github.com/cometbft/cometbft/crypto/ed25519"
 	"github.com/cometbft/cometbft/p2p"
+	"github.com/kwilteam/kwil-db/common/sql"
+	"github.com/kwilteam/kwil-db/internal/sql/versioning"
+)
+
+const (
+	p2pSchemaName = "peers" // TODO: Umm, should this be part of the chain metadata schema?
```
brennanjl, line 15: > // TODO: Umm, should this be part of the chain metadata schema?


No, chain metadata is included as part of consensus (affects app-hash), but peers are a local db that is specific to each node

internal/abci/cometbft/peers.go
```diff
@@ -1,10 +1,183 @@
...
+
+	p2pStoreVersion = 0
+
+	initPeersTable = `CREATE TABLE IF NOT EXISTS ` + p2pSchemaName + `.peers (
+		peer_id TEXT PRIMARY KEY
+	);`
+
+	addPeer = `INSERT INTO ` + p2pSchemaName + `.peers ` + `VALUES ($1) ` + `ON CONFLICT (peer_id) DO NOTHING;`
```
brennanjl, line 23: I think if the peer already exists, we should return an error. If I was a user, I would expect some sort of error like `"%s" is already whitelisted as a peer`

internal/abci/cometbft/peers.go
```diff
@@ -1,10 +1,183 @@
...
+}
+
+// AddPeer adds a peer to the Peers object.
+func (p *Peers) AddPeer(ctx context.Context, peer string) error {
+	p.peerMtx.Lock()
+	defer p.peerMtx.Unlock()
+
+	_, ok := p.whitelistPeers[peer]
+	if ok {
+		return nil
```
brennanjl, line 91: Noted in the ON CONFLICT clause, but I think we should return an error here

internal/abci/cometbft/peers.go
```diff
@@ -1,10 +1,183 @@
...
+}
+
+// RemovePeer removes a peer from the Peers object.
+func (p *Peers) RemovePeer(ctx context.Context, peer string) error {
+	p.peerMtx.Lock()
+	defer p.peerMtx.Unlock()
+
+	_, ok := p.whitelistPeers[peer] // check if peer exists
+	if !ok {
+		return nil
```
brennanjl, line 117: Probably return error here

internal/abci/cometbft/peers.go
```diff
@@ -1,10 +1,183 @@
...
+	_, err = tx.Execute(ctx, removePeer, peer)
+	if err != nil {
+		return err
+	}
+
+	return tx.Commit(ctx)
+}
+
+// HasPeer checks if a peer is in the Peers object.
+func (p *Peers) HasPeer(ctx context.Context, peer string) bool {
```
brennanjl, line 144: No need for context?

internal/abci/cometbft/peers.go
```diff
@@ -1,10 +1,183 @@
...
 
+// Peers object is used to manage the set of peers that a node is allowed to connect to.
+// Static peers are those that are added in the genesis and are trusted by default. These
+// include the initial validators, persistent peers, and seed nodes.
+// Whitelist peers are the list of allowed sentry nodes that a node can connect to.
+// Whitelist peers can be updated dynamically using kwil-admin commands and are persisted
+// to disk.
+// The Peers gossiped using PEX are not automatically trusted, and to be added manually if needed.
+// Any new validators are automatically added to the whitelist peers and demoted validators are removed from the whitelist.
+type Peers struct {
```
brennanjl, line 36: This code all looks great, but maybe rename this to be a `PeerWhitelist` (and change the methods to match that logic)? If I was someone trying to understand what is going on here, I would read this believing that this is the ground truth for what peers are connected, rather than what peers are allowed to connect.

cmd/kwild/config/config.go
```diff
@@ -119,6 +119,9 @@ type P2PConfig struct {
 	// PersistentPeers is a comma separated list of nodes to keep persistent
 	// connections to.
 	PersistentPeers string `mapstructure:"persistent_peers"`
+	// WhitelistPeers is a comma seperated list of nodeIDs that can connect to this node.
+	// This is excluding any persistent peers or seeds or current validators.
+	WhitelistPeers string `mapstructure:"whitelist_peers"`
```
brennanjl, line 124: We probably want a config for running in private mode, right? Pretty much a `--private` flag that enforces this? The default action should still be public networks imo.

internal/abci/abci.go
```diff
@@ -1450,7 +1483,36 @@ func (a *AbciApp) Query(ctx context.Context, req *abciTypes.RequestQuery) (*abci
...
+
+		paths := strings.Split(req.Path[ABCIPeerFilterPathLen:], "/")
+		if len(paths) != 2 {
+			return &abciTypes.ResponseQuery{Code: 1}, fmt.Errorf("invalid path: %s", req.Path)
+		}
+
+		switch paths[0] {
+		case "id":
+			if a.p2p.IsPeerWhitelisted(paths[1]) {
+				return &abciTypes.ResponseQuery{Code: abciTypes.CodeTypeOK}, nil
```
brennanjl, line 1503: You had a log here letting the operator know that the connection was successful. Curious why you took it out?

internal/txapp/txapp.go
```diff
@@ -350,20 +350,61 @@ func (r *TxApp) Finalize(ctx context.Context, db sql.DB, block *common.BlockCont
...
+	}
+
+	var approvedJoins []*types.Validator
+	for _, join := range activeJoins {
+		var req voting.UpdatePowerRequest
+		if err := req.UnmarshalBinary(join.Body); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal join request: %w", err)
+		}
+
+		for _, voter := range join.Voters {
```
brennanjl, line 391: Why do we query the db here? All we need to do is recognize that, if a tx is a vote tx for a validator approval AND that was made by the local node, then we should approve the validator.

The reason I have an issue with this is because we are doing a full table scan here at the end of every block, and the resolutions table could be quite large for networks that use oracles.

brennanjl (commented):
Aside from the table scan, this looks good to me.

I think you can handle the table scan by having some extra logic in `*TxApp.Execute()`

internal/abci/abci.go
charithabandi: felt kinda unnecessary, as cometbft p2p layer would log the reason if we reject the request

brennanjl:
I seem to have found a bug while running it.

If I had a read node as a whitelist by using the flag `--chain.p2p.whitelist-peers`, and then restart my node without that peer, that node is still able to connect. I am assuming that this is being persisted somewhere, however we should only persist nodes that are set via admin RPC

brennanjl:
> If I had a read node as a whitelist by using the flag --chain.p2p.whitelist-peers, and then restart my node without that peer, that node is still able to connect. I am assuming that this is being persisted somewhere, however we should only persist nodes that are set via admin RPC

Confirimed, listing my peers showed this was the case.

Also, quite weirdly, it had my local node as a "peer". I was able to remove it (nothing happened). I think we should not add the local node as a peer.

brennanjl:
Another bug: I removed a peer, then restarted my node, and the peer I removed was back in my peer list.

brennanjl:
As discussed, fledging out peer addition and removal:

There are 4 ways to become a peer:
1. whitelist via config, set on node startup
1.1. This should not persist, but instead only be tracked in memory
1.2. The node operator should be able to remove this via `kwil-admin`. If the node restarts with this config again, the peer will be re-added
2. set using `kwil-admin`
2.1. This should be persisted, so that if a node restarts, they do not have to specify this peer
2.2. The node operator can remove this via `kwil-admin`, which permanently removes the peer.
3. a node votes to add a validator, and adds the potential new validator as a peer
3.1. This should not persist. If a node restarts, it should query the vote store to see any nodes that it has voted on that are not approved, and add them as peers.
3.1. A node can remove this peer via `kwil-admin`, but if they restart, they will need to remove again.
4. a node adds all current validators as their peers
4.1. This should not persist, since the validator list is already persisted.
4.2. Nodes can remove these peers via `kwil-admin`, but if they restart, they will be added as valid peers.

brennanjl (dismissed):

brennanjl (approved):

PR #929, opened 2024-08-23
statistics foudation -- MCVs and histograms for all types

jchappelow:
DON'T REVIEW YET.  Needs heaps of cleanup, docs, and a few more tests.


jchappelow:
Some of these test are atrocious and there is UUID vs *UUID breakage in multiple places.  :( fixing

jchappelow:
There are a few cherries to pick out of this, but overall we don't need the changes now.  Closing.

Issue #842, opened 2024-06-19
Feature: Fail Fast on missing resolution


brennanjl:
@sapience from the idOS team it an issue where resolutions are being stored and voted on, but are unable to be processed because their resolution type is not registered. On startup, we should verify that the set of resolution IDs registered in the internal voting schema exactly match the extensions registered. If they do not match, we should fail to start the node, as this indicates a change in consensus rules.

PR #938, opened 2024-08-27
catch any mismatches with the registered resolutions and resolutions in vote store at node startup

charithabandi:


jchappelow:
Was there an issue or discussion explaining what could go wrong without this?  For instance, installing a binary that lacks the compiled-in resolution types that are referenced by the `kwild_voting.resolutions` table.  Related: #952

brennanjl:
> Was there an issue or discussion explaining what could go wrong without this? For instance, installing a binary that lacks the compiled-in resolution types that are referenced by the kwild_voting.resolutions table. Related: https://github.com/kwilteam/kwil-db/issues/952

Yes: https://github.com/kwilteam/kwil-db/issues/842

A thread of when idOS hit this: https://kwilteam.slack.com/archives/C0578NPEU21/p1718791968399959

internal/voting/voting.go
```diff
@@ -56,9 +56,50 @@ func initializeVoteStore(ctx context.Context, db sql.TxMaker) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to validate resolution types: %w", err)
+	}
+
 	return tx.Commit(ctx)
 }
 
+// validateResolutionTypes verifies if the resolutions stored in the vote store
+// are the same as the resolutions registered. Any mismatch at the startup would
+// indicate unplanned consensus changes and should be treated as a critical error.
```
jchappelow, line 69: `validateResolutionTypes` is called *after* a loop around `createResolutionType`, so I believe only the length check could ever hit as it is currently used.  Did we want to detect resolutions an existing table that are not in the registered resolutions map from `ListResolutions`?

internal/voting/voting.go
```diff
@@ -41,13 +41,36 @@ func initializeVoteStore(ctx context.Context, db sql.TxMaker) error {
...
+	for _, name := range compiledResolutions {
+		resMap[name] = true
+	}
+
+	dbResolutions, err := getResolutionTypes(ctx, tx)
+	if err != nil {
+		return fmt.Errorf("failed to get resolution types: %w", err)
+	}
+
+	if len(dbResolutions) != 0 {
```
brennanjl, line 55: Why do you check if `len(dbResolutions) != 0`?. Seems unnecessary

brennanjl (commented):
Minor comment but otherwise lgtm

internal/voting/voting.go
charithabandi: Ah, sorry, relic from the past commits. removed it

brennanjl (approved):

Issue #902, opened 2024-07-31
improvement: Use `common.BlockContext` within `common.TransactionData`


brennanjl:
Currently, we have these two structs both defined in the `common` package:

```go
// TransactionData holds contextual data about the transaction that
// called the procedure.
type TransactionData struct {
	// Signer is the address of public key that signed the incoming
	// transaction.
	Signer []byte

	// Caller is a string identifier for the signer.
	// It is derived from the signer's registered authenticator.
	// It is injected as a variable for usage in the query, under
	// the variable name "@caller".
	Caller string

	// TxID is the transaction ID of the incoming transaction.
	TxID string

	// Height is the block height of the incoming transaction.
	Height int64

	// BlockTimestamp is the unix timestamp of the block, set by the block proposer.
	BlockTimestamp int64
}
```

```go
// BlockContext provides context for all block operations.
type BlockContext struct {
	// ChainContext contains information about the chain.
	ChainContext *ChainContext
	// Height gets the height of the current block.
	Height int64
	// BlockTimestamp is a timestamp of the current block.
	// It is set by the block proposer, and therefore may not be accurate.
	// It should not be used for time-sensitive operations where incorrect
	// timestamps could result in security vulnerabilities.
	BlockTimestamp int64
	// Proposer gets the proposer public key of the current block.
	Proposer []byte
}
```

We should remove the block information from `TransactionData`, and simply include a `BlockContext`. The `TransactionData` is created from the `BlockContext` within `TxApp`, so it is mostly just a useless layer of abstraction. It is a minor issue, but it is a breaking change.

jchappelow:
Resolved by https://github.com/kwilteam/kwil-db/pull/939

PR #939, opened 2024-08-27
unified common structs used throughout codebase

brennanjl:
Ok this was the last major change I wanted to make before we have another release.

This change removes/refactors some unnecessary structs in `common`, and unifies them. In particular, we had essentially two structs that do the same thing, and we often switch between them. The reason I felt this change was necessary is because there are currently some very awkward patterns used in extensions (resolutions in particular) that are the result of not properly duplicating fields/information in these structs.

Therefore, this is a breaking change, however it is trivially easy to port over old extensions, and extensions already received a breaking interface change in this release.

### DIFF
--- a/internal/services/jsonrpc/usersvc/service.go
+++ b/internal/services/jsonrpc/usersvc/service.go
@@ -583,10 +583,12 @@ func (svc *Service) Call(ctx context.Context, req *userjson.CallRequest) (*userj
 		Dataset:   body.DBID,
 		Procedure: body.Action,
 		Args:      args,
-		TransactionData: common.TransactionData{
-			Signer:        signer,
-			Caller:        caller,
-			Height:        -1, // not available
+		TxCtx: &common.TxContext{
+			Signer: signer,
+			Caller: caller,
+			BlockContext: &common.BlockContext{
+				Height: -1, // cannot know the height here.
+			},
 			Authenticator: msg.AuthType,
 		},
 	})
--- a/internal/services/jsonrpc/usersvc/service.go
+++ b/internal/services/jsonrpc/usersvc/service.go
@@ -207,10 +207,10 @@ func (svc *Service) Handlers() map[jsonrpc.Method]rpcserver.MethodHandler {
 }
 
 type EngineReader interface {
-	Procedure(ctx context.Context, tx sql.DB, options *common.ExecutionData) (*sql.ResultSet, error)
+	Procedure(ctx *common.TxContext, tx sql.DB, options *common.ExecutionData) (*sql.ResultSet, error)
 	GetSchema(dbid string) (*types.Schema, error)
 	ListDatasets(owner []byte) ([]*types.DatasetIdentifier, error)
-	Execute(ctx context.Context, tx sql.DB, dbid string, query string, values map[string]any) (*sql.ResultSet, error)
+	Execute(ctx *common.TxContext, tx sql.DB, dbid string, query string, values map[string]any) (*sql.ResultSet, error)
 }
 
 // NOTE:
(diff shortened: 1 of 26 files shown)


brennanjl:
Resolves https://github.com/kwilteam/kwil-db/issues/902

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -583,10 +583,12 @@ func (svc *Service) Call(ctx context.Context, req *userjson.CallRequest) (*userj
 		Dataset:   body.DBID,
 		Procedure: body.Action,
 		Args:      args,
-		TransactionData: common.TransactionData{
-			Signer:        signer,
-			Caller:        caller,
-			Height:        -1, // not available
+		TxCtx: &common.TxContext{
+			Signer: signer,
```
jchappelow, line 587: Is it intended to keep `Ctx` nil here?  Is `common.TxContext.Ctx` only used in consensus / block execution code?  The `Ctx` field itself does not any documented conventions so the assumption is it should be set, I think

brennanjl: Hmm that is an oversight. You are right that we are just using it to pass contextual vars. We could just duplicate the context to the TxContext? The other option is a larger breaking change to the engine interface by removing the first ctx parameter and replace it with tx context.

IMO the latter seems more logical.

jchappelow: > The other option is a larger breaking change to the engine interface by removing the first ctx parameter and replace it with tx context.
> 
> IMO the latter seems more logical.

Yah, kinda.  Although looking at `svc.engine.Procedure`, this `TxContext` is just a field of `common.ExecutionData`, which is... not really a context?  Sorta.  It's partially arguments (dataset/procedure/args) and partially request-scoped stuff.  Does it feel a bit like `ExecutionData` went a little wrong in an effort to dedup?

brennanjl: The way I see it:

An executions are within a transaction (there can be several executions in one tx in the case of a batch). Transactions are within a block, and blocks can be in a network (this is where the hierarchy comes from).

IMO, the most logical place for request scoping is within the `TxContext`. It seems like `ExecutionData` now should be `ExecutionContext`, and we replace the ctx.Context with these.

brennanjl: I do agree that it seems like these shifted a bit though.

brennanjl: Actually, we have a `ProcedureContext` which is essentially a duplicated version of `ExecutionData`. We only use it internally though

brennanjl: I'm trying a few things out, will get something fixed for this tonight.

jchappelow: > An executions are within a transaction (there can be several executions in one tx in the case of a batch). Transactions are within a block, and blocks can be in a network (this is where the hierarchy comes from).

Yeah, what you're doing along these lines makes perfect sense.  What feels less contexty is these:

```go
 		Dataset:   body.DBID,
 		Procedure: body.Action,
 		Args:      args,
```

I think it all works out fine. NBD.  As long as it's easy to manage passing of the actual `context.Context`.

brennanjl: Yeah, what I am doing is keeping `ExecutionData` as just that, and passing the `TxContext` in place of the `context.Context`.

brennanjl:
Ok @jchappelow I now use `TxContext` where we previously use both `context.Context` and `TxContext`

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -579,18 +584,18 @@ func (svc *Service) Call(ctx context.Context, req *userjson.CallRequest) (*userj
...
-	executeResult, err := svc.engine.Procedure(ctxExec, readTx, &common.ExecutionData{
+	executeResult, err := svc.engine.Procedure(&common.TxContext{
+		Ctx:    ctxExec,
+		Signer: signer,
+		Caller: caller,
+		BlockContext: &common.BlockContext{
+			Height: -1, // cannot know the height here.
+		},
+		Authenticator: msg.AuthType,
+	}, readTx, &common.ExecutionData{
```
jchappelow, line 595: Nice, LGTM this way.

jchappelow (approved):
LGTM.  update looks good. Did you want to do anything else?

PR #940, opened 2024-08-27
Create snapshots if snapshots are enabled and none exists

charithabandi:
node should create a snapshot if:
- none exists
- it is height 1 (genesis+1)
- it is not doing blocksync to some much larger height
- height is a multiple of snapshot duration

### DIFF
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -865,8 +865,17 @@ func (a *AbciApp) Commit(ctx context.Context, _ *abciTypes.RequestCommit) (*abci
 
 	a.txApp.Commit(ctx)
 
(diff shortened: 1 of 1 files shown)


charithabandi:
Brought this up with the cometbft guys, they agreed to the proposal to switch to blocksync if there are no snapshots in the network. But this change would help till we get that feature out, to unblock the node from syncing to the network,.

jchappelow (approved):
LGTM, although I'd try to order the logical operators to avoid pointless operations like `ListSnapshots` if it were for example replaying blocks.  Short-circuit behavior of logical operators like && and || is good to exploit.

PR #942, opened 2024-08-28
pg: elide unchanged tup vals in UPDATE changeset

jchappelow:
This PR has two closely related commits:

1. https://github.com/kwilteam/kwil-db/pull/942/commits/b1f9c903f69f9f92628ad705131ebed51ebd097d In changesets, the `ChangesetElentry` for an UPDATE includes both old and new values.  All of the old values are required, to be used with a fully-specified `WHERE` clause.  However, only the changed *new* values need to be included in the `SET` clause.  This commit deduplicates this unchanged data from the changesets.  This can drastically reduce the size of the serialized changesets (imagine only modifying an integer column and other large bytea/text columns unchanged), which must be stored and trasmitted cross-chain and then recorded permanently in governance transactions on the new chain.   The second reason it is helpful to flag these values as unchanged is that it makes statistics maintenance simpler when handling updates.
2. https://github.com/kwilteam/kwil-db/pull/942/commits/baaceb99e9cba683c2f180a12a71e9598a047a90 formalizes the changeset streaming format. Previously a special 5 byte prefixing convention and code was duplicated in several methods.  Plus the Serialize/Deserialize methods were not complementary, meaning they did not match and could not round trip data.  The `ChangeStreamer` interface and the `StreamElement` and `DecodeStreamPrefix` functions are now used to encasulate these details and conventions.  Each of the concrete types that are encoded/decoded to/from a stream of changeset elements implement the `ChangeStreamer`.  It is still awkward there is a single enum of prefix bytes but that both `internal/migrations` and `internal/sql/pg`, although this is still much clearer than before.

### DIFF
--- a/internal/migrations/changesets.go
+++ b/internal/migrations/changesets.go
@@ -2,7 +2,6 @@ package migrations
 
 import (
 	"context"
-	"encoding/binary"
 	"errors"
 	"math/big"
 
@@ -187,11 +186,10 @@ func applyChangeset(ctx context.Context, db sql.TxMaker, height int64, totalChun
 	defer tx.Rollback(ctx)
 
 	csReader := newChangesetReader(height, totalChunks)
-	var data []byte
 
 	var relations []*pg.Relation
 	for {
-		data, err = csReader.Read(ctx, tx, 5)
+		prefix, err := csReader.Read(ctx, tx, 5)
 		if err != nil {
 			// no more chunks or changesets not found
 			if errors.Is(err, ErrNoMoreChunksToRead) || errors.Is(err, ErrChangesetNotFound) {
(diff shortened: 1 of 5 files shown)


jchappelow:
Requesting review from you @charithabandi since you we talked about this yesterday and you know this code best.  I still need to run this on my statistics integration branch before I feel good about it though.
Are there any manual zdt migration tests to try, or do integration tests cover it all?

charithabandi:
> Requesting review from you @charithabandi since you we talked about this yesterday and you know this code best. I still need to run this on my statistics integration branch before I feel good about it though. Are there any manual zdt migration tests to try, or do integration tests cover it all?

there are unit tests in db_live_test.go under the name TestChangesetConflicts, also integration tests should cover.  
Will review this soon.

brennanjl (approved):
lgtm

PR #945, opened 2024-08-29
types,migrations,admin,rpc: cleanup and migrations type fix

jchappelow:
This PR does a few minor things pertaining to recent migrations code in various packages:

- In `core/types/MigrationMetadata` change the `GenesisConfig` and `SnapshotMetadata` fields from `[]byte` to `json.RawMessage`. This is the preferred type when delaying the unmarshal of raw JSON, which is exactly the case as we unmarshal these into internal types.  This avoids the RPC service/client from having to needlessly round trip the content, which is already text (JSON), to base64 and back.  This is very similar to changing the field to a `string`.  There are no visible changes unless using `curl` to see the raw response.
- In each of the `cmd/kwil-admin/cmds/...`, cleanup various typo, use the `cmd.Context()`, set arg counts, docs tweaks, etc.
- Rename `internal/migrations.ChangesetMetdata` to `ChangesetMetadata` (spelling)

### DIFF
--- a/cmd/kwil-admin/cmds/migration/approve.go
+++ b/cmd/kwil-admin/cmds/migration/approve.go
@@ -1,12 +1,11 @@
 package migration
 
 import (
-	"context"
+	"github.com/spf13/cobra"
 
 	"github.com/kwilteam/kwil-db/cmd/common/display"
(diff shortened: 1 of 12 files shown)


charithabandi (approved):
LGTM!

PR #946, opened 2024-08-29
Authenticate RPC queries: /call and /query

charithabandi:
This PR enables challenge based auth on the call and adhoc query endpoints when run with `--app.authenticate-rpcs=true`

Introduced /challenge endpoint to request for challenge and updated call and query endpoints to request a challenge and send call/query requests with a signed challenge. 

`SignedChallenge` struct includes challenge and its signature, which is verified by the JSON RPC handler in the  private mode.


internal/services/jsonrpc/usersvc/service.go
```diff
@@ -42,6 +43,14 @@ type Service struct {
 	chainClient BlockchainTransactor
 	pricer      Pricer
 	migrator    Migrator
+
+	// authenticate RPCs?
+	authenticateRPCs bool
+	// challenges issued to the clients
+	// TODO: have a goroutine that cleans up expired challenges for every few hrs?
```
jchappelow, line 50: I don't think we should punt on this as long as we are using this approach of allowing unlimited challenge requests.

core/rpc/json/user/commands.go
```diff
@@ -91,6 +91,8 @@ type EstimatePriceRequest struct {
 type QueryRequest struct {
 	DBID  string `json:"dbid"`
 	Query string `json:"query"`
+
+	Challenge *types.SignedChallenge `json:"challenge"`
```
jchappelow, line 95: `omitempty` in the JSON tag.  We don't want `"challenge": null` in every request that is not using authenticated RPCS.

internal/services/jsonrpc/usersvc/service.go
charithabandi: Yeah agreed.

cmd/kwil-cli/cmds/database/call.go
```diff
@@ -102,6 +111,7 @@ func callCmd() *cobra.Command {
 	cmd.Flags().StringP(dbidFlag, "i", "", "the target database id")
 	cmd.Flags().StringVarP(&action, actionNameFlag, "a", "", "the target action name (required)")
 	cmd.Flags().BoolVar(&authenticate, "authenticate", false, "authenticate signals that the call is being made to a gateway and should be authenticated with the private key")
+	cmd.Flags().BoolVarP(&requestChallenge, "challenge", "c", false, "request a challenge from the server to authenticate the call")
```
jchappelow, line 114: I find this flag's name and description to be a little too implementation oriented.  The point here is that we are indicating that we expect to be making a call to a server that requires a signature in the call request so server can authenticate us.  Using a challenge is a detail of our approach to preventing replay.
Can this work another name like `auth` or `signed`?  I know it's confusign with the gateway's `authenticate`, but I don't think `challenge` is to the point.

cmd/kwild/config/flags.go
```diff
@@ -24,6 +24,8 @@ func AddConfigFlags(flagSet *pflag.FlagSet, cfg *KwildConfig) {
 	flagSet.StringVar(&cfg.AppCfg.TLSCertFile, "app.tls-cert-file", cfg.AppCfg.TLSCertFile, "TLS certificate file path for the admin and consensus RPC server (optional)")
 	flagSet.StringVar(&cfg.AppCfg.TLSKeyFile, "app.tls-key-file", cfg.AppCfg.TLSKeyFile, "TLS key file path for the admin and consensus RPC servers (optional)")
 	flagSet.StringVar(&cfg.AppCfg.Hostname, "app.hostname", cfg.AppCfg.Hostname, "kwild Server hostname")
+	flagSet.BoolVar(&cfg.AppCfg.AutheticateRPCs, "app.authenticate-rpcs", cfg.AppCfg.AutheticateRPCs, "Authenticate JSON-RPC requests using the private key")
```
jchappelow, line 27: What does "using the private key" mean here?  This is a `kwild` flag, but I think you're referring to the client using their private key to create a signature to include in their requests.  For this `kwild` flag, I think the description should be more like "Authenticate call and query RPC request, requiring clients to include a signature in their requests."

core/client/client.go
```diff
@@ -295,7 +295,7 @@ func (c *Client) CallAction(ctx context.Context, dbid string, action string, inp
 }
 
 // Call calls a procedure or action. It returns the result records.
-func (c *Client) Call(ctx context.Context, dbid string, procedure string, inputs []any) (*clientType.CallResult, error) {
+func (c *Client) Call(ctx context.Context, dbid string, procedure string, inputs []any, challenge []byte) (*clientType.CallResult, error) {
```
jchappelow, line 298: Just a thought, but I don't see a compelling reason to expose the challenge or even knowledge of the authentication mechanism to the consumer of `Client`.  i.e. couldn't this be `Call(..., signReq bool)` and the method would do the challenge request internally?  Just for sake of discussion...  What is gained from explorting a `Challenge()` method?

core/types/types.go
```diff
@@ -100,3 +102,9 @@ type MigrationMetadata struct {
 	GenesisConfig    []byte `json:"genesis_config"`    // GenesisConfig is the genesis file data
 	SnapshotMetadata []byte `json:"snapshot_metadata"` // SnapshotMetadata is the snapshot metadata
 }
+
+type SignedChallenge struct {
+	Challenge []byte          `json:"challenge"`
+	Signature *auth.Signature `json:"signature"`
+	Sender    []byte          `json:"sender"`
```
jchappelow, line 109: As with `CallMessage.Sender`, this should be `HexBytes`.

BTW, it's a little unfortunate that a `CallMessage` has will repeat sender twice, once int the struct itself, and once in this field.  Thoughts about avoiding that?

core/types/transactions/message.go
```diff
@@ -27,6 +27,9 @@ type CallMessage struct {
 
 	// Sender is the public key of the sender
 	Sender types.HexBytes `json:"sender"`
+
+	// Challenge is the signed challenge to prove the sender's identity
+	Challenge *types.SignedChallenge `json:"challenge"`
```
jchappelow, line 32: Dang, I'm not really uncertain about signed challenge being part of the call request body or the JSON RPC service at all.  The *server*, not the RPC *service* may be the place to handle authentication. 

Specifically, in `(*Server) handlerV1`, we are dealing with the HTTP request, checking headers like http Basic auth (for admin server), then unmarshalling the request, and dispatching to the registered method handler.  In the server we could be issuing the challenges, and in this *server* method, we could be getting the the auth data from an http header, verifying it, and then calling into the service methods with the authenticated sender in the context.

The goal is to keep the authentication as far from the business of the *service* methods as possible.

I gotta think about it some more.

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -519,6 +592,13 @@ func (svc *Service) Call(ctx context.Context, req *userjson.CallRequest) (*userj
 
 	}
 
+	// If authenticateRPCs is enabled, we need to check the challenge
+	if svc.authenticateRPCs {
+		if err := svc.verifyChallenge(msg.Challenge); err != nil {
```
jchappelow, line 597: It seems like the most important detail to verify here (besides validity of signature) is that the sender in the challenge is the sender in the call message.  But we're missing that

In context of https://github.com/kwilteam/kwil-db/pull/946/files#r1739103177, this is the only concern of this Call method, if it were the job of the server to verify the signature and the pass the verified identity to the service method to do what it will: (1) in call, ensure that the call.sender is equal to challenge.sender, and (2) in query, just make sure that there is an authenticated sender.

BTW, what's the point of authentication on `query` if it doesn't matter who signed, only it was signed?  Don't we want a whitelist or something?  Or it it supposed to only allow the schema owner to do ad hoc queries?

core/client/client.go
jchappelow: Ah, so one step further: make `signReq bool` a client config.  No new `Challenge` method and no change to `Call` or `Query`?  I'm inclined to hide away the challenge details.

internal/services/jsonrpc/usersvc/service.go
charithabandi: talked to brennan about this. Apparently, the expectation is, we shouldn't allow adhoc queries in private mode.

core/types/transactions/message.go
jchappelow: @Yaiba @charithabandi and I just talked through a few related points:

- a template that includes (a) the challenge, (b) plain text explaining purpose and making it obvious it is not a transaction, and (c) and perhaps what json-rpc method it is intended for -- sign the template -- server generates the exepected template on the fly in order to verify the signuature
- segregating authentication API data from application data.  Namely, the server handling authentication while application/service method only needs the call data (schema method and action inputs etc).  I think @Yaiba has a good perspective on this with kgw.  AFAICT (correct me if I'm wrong) this is a close analogy but in this case we don't have a session cookie, instead a one-time authentication challenge that is discarded by server after use.  In kgw we didn't need to add the auth details the challenge to the `CallMessage`, and I'm not convinced we do now either.  Is private mode somehow conceptually more part of the application to the point that the service methods or engine need to be aware?  I don't think so.

Regarding the second point above, another consideration is if we change the authentication scheme, we would would have to change the core types.  Separating these pieces of information would provide flexibility.  Further, imagine if we devised a way to have transport layer authentication (e.g. mTLS with custom certification creation and verification to serve the same purpose), it would entirely vanish from the request body.

internal/services/jsonrpc/usersvc/service.go
jchappelow: Ahhh, that makes more sense.  As it was, verifying that there was a valid signature with no requirement that signer match anything in particular... well that didnt' do much. :)

core/types/client/client.go
```diff
@@ -31,10 +31,12 @@ type Client interface {
 	GetSchema(ctx context.Context, dbid string) (*types.Schema, error)
 	ListDatabases(ctx context.Context, owner []byte) ([]*types.DatasetIdentifier, error)
 	Ping(ctx context.Context) (string, error)
-	Query(ctx context.Context, dbid string, query string) (*Records, error)
+	Query(ctx context.Context, dbid string, query string, challenge []byte) (*Records, error)
```
brennanjl, line 34: As noted in discussions, query should just be disabled in private mode

cmd/kwild/config/flags.go
brennanjl, line 27: Instead of `authenticate-rpcs`, maybe do `private-rpc`? Simply because we only authenticate one of the RPC methods (`call`), we block another (`query`), and modify but don't authenticate the contents of a third (`tx status`).

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -64,26 +75,46 @@ func WithReadTxTimeout(timeout time.Duration) Opt {
...
+		log:              logger,
+		readTxTimeout:    cfg.readTxTimeout,
+		engine:           engine,
+		nodeApp:          nodeApp,
+		pricer:           pricer,
+		chainClient:      chainClient,
+		db:               db,
+		migrator:         migrator,
+		authenticateRPCs: cfg.authenticateRPCs,
+		challengeExpiry:  cfg.challengeExpiry,
```
brennanjl, line 116: You need to modify `tx-query` because it returns the full transaction body. In private mode, we should make sure to not do this, since it defeats the point.

internal/services/jsonrpc/usersvc/service.go
brennanjl: +1. Also, this map can be called concurrently, so you should use `internal/utils/sync_map`

brennanjl: Ah, nvm I see you have a mutex. Syncmap hides this / ensures that we don't forget it, but I'll leave it up to you to decide.

brennanjl (commented):
Ok I left a few comments, but overall it seems like @jchappelow has a better idea on the direction to take this.

One more thing I noticed while going through it relating to our convo:

We need to remove the transaction body for `tx-query` if we are in private mode. I left a note of this, but we should keep this in mind when figuring out how to include authentication within the HTTP server instead of the user service. Maybe it simply means we have some `private` flag in the service, but I have no strong opinion.

jchappelow:
With @charithabandi now completely swamped with [new migration issues](https://github.com/kwilteam/kwil-db/issues/952) that came out of the trial zdt migration of staging this weekend, I'll do next steps on this with a legible serialized message signing scheme for CallMessage.

As per the [understanding of `CallMessage` as a pseudo transaction](https://github.com/kwilteam/kwil-db/pull/949#issuecomment-2327453534) with `ActionCall` as it's acceptable binary-serializable payload, I'll do quickly implement the approach taking the first view:

1. call method specific, with the challenge like a server-provided nonce, which puts this data firmly in the `CallMessage` struct where it's serialize method will include it as part of a legible signed message.
2. http server wide authentication that functions above all service methods, providing the authenticated user id (if auth'd) into the service methods to do as they wish,

Again, quickly doing option 1 as an 11th hour reinterpretation of the call message and the method handling.

I do intend to demo option 2 as I think it is more versatile and allows the RPC method parameter shapes to stabilize with authentication managed by the communications layer instead.  The templated serialization would of course look different to the user.

jchappelow:
https://github.com/kwilteam/kwil-db/pull/946/commits/19174ea352432d3491a8b98d3d565b47477c3258:
- maintains the pseudo-transaction representation of `CallMessage`, building a signature field into the struct along with a Challenge field with the payload.  This mirrors the `Transaction` type now, although I would prefer to have fully separated the authentication details from the JSON-RPC request parameters shape, but we can reconsider after this week's release.
- adds a goroutine to clear out expired challenges
- adds a rate limiter to prevent abuse even with the cleanup goroutine

I need to review my own changes and do manual tests, but acceptance and integration pass.

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -511,14 +567,56 @@ func resultMap(r *sql.ResultSet) []map[string]any {
...
+		}
+		// Ensure we issued the message's challenge.
+		if err := svc.verifyCallChallenge([32]byte(msg.Body.Challenge)); err != nil {
+			return nil, err
+		}
+		sigtxt := transactions.CallSigText(body.DBID, body.Action,
+			msg.Body.Payload, msg.Body.Challenge)
+		err = ident.VerifySignature(msg.Sender, []byte(sigtxt), msg.Signature)
+		if err != nil {
+			return nil, jsonrpc.NewError(jsonrpc.ErrorInvalidCallSignature, "invalid signature on call message", nil)
```
brennanjl, line 616: Since we have `transactions.CallMessage.AuthType` and `transactions.CallMessage.Signature.AuthType`, we need to have some conditional logic here. Either:

- The `AuthType` we pass to the engine is  `transactions.CallMessage.AuthType` in the case of public mode, and  `transactions.CallMessage.Signature.AuthType` in case of private mode.
OR
- We check that  `transactions.CallMessage.Signature.AuthType` and  `transactions.CallMessage.AuthType` are equal if we are in private mode.

Right now, we only pass  `transactions.CallMessage.AuthType` to the engine, which doesn't necessarily match the authtype used in private mode

cmd/kwild/config/flags.go
jchappelow: Will make this change to `private-rpc`.  I agree with your points, just missed this comment.

jchappelow:
OK this almost good but I really need to make the challenge request limits configurable...

core/client/client.go
```diff
@@ -330,6 +341,10 @@ func (c *Client) Call(ctx context.Context, dbid string, procedure string, inputs
 
 // Query executes a query.
 func (c *Client) Query(ctx context.Context, dbid string, query string) (*clientType.Records, error) {
+	// if c.authCallRPC {
```
Yaiba, line 344: do we still need this?

core/rpc/json/errors.go
```diff
@@ -41,6 +41,14 @@ const (
 	ErrorValidatorsInternal ErrorCode = -800
 	ErrorValidatorNotFound  ErrorCode = -801
 
+	ErrorCallChallengeNotFound ErrorCode = -1001
```
Yaiba, line 44: just nitpicking, move this block below KGW?

core/types/transactions/message.go
```diff
@@ -27,18 +41,58 @@ type CallMessage struct {
...
+	}
+
+	if signer != nil { // for @caller
+		msg.AuthType = signer.AuthType()
+		msg.Sender = signer.Identity()
+	}
+
+	if len(challenge) > 0 {
+		sigText := CallSigText(ac.DBID, ac.Action, bts, challenge)
+		sig, err := signer.Sign([]byte(sigText))
```
Yaiba, line 89: `signer` could be nil here

internal/services/jsonrpc/usersvc/service.go
```diff
@@ -736,3 +849,37 @@ func (svc *Service) ListPendingMigrations(ctx context.Context, req *userjson.Lis
 		Migrations: pendingMigrations,
 	}, nil
 }
+
+func (svc *Service) expireChallenges() {
```
Yaiba, line 853: Since the ratelimit configuration is fairly high(or maybe the Operator configured high), we might have a big challenge map, maybe worth more consideration?

I'm thinking maybe only loop certain number(use ratelimit configure as a factor to adjust this number) of entries, thus reduce the time the lock is held.

Yaiba: Or maybe run checkExpire and delete in different batch, we release the lock between the batches

Yaiba (approved):
LMTM.  only concern about that we might have a huge challenge map(also depends on user's configuration), which could block user make Call requests

core/types/transactions/message.go
brennanjl: I don't think so. I assume you're referring to `signer` sometimes being nil if it was not configured in a `client`. However, I feel like any checks for that should be done at a higher level, since this is simply a function in the transactions package.

internal/services/jsonrpc/usersvc/service.go
brennanjl: Seems maybe a bit premature IMO.

brennanjl (approved):
@Yaiba already approved but overall LGTM

core/types/transactions/message.go
jchappelow: Yeah, this would be vanilla misuse of a the function.  I've opted for docs:

```go
// CreateCallMessage creates a new call message from a ActionCall payload. If a
// signer is provided, the sender and authenticator type are set. If a challenge
// is also provided, it will also sign a serialization of the request that
// includes the challenge for replay protection. Thus, if a challenge is
// provided, a signer must also be provided.
```

internal/services/jsonrpc/usersvc/service.go
jchappelow: > I'm thinking maybe only loop certain number(use ratelimit configure as a factor to adjust this number) of entries, thus reduce the time the lock is held.

Let's chat about this.  I'm not sure how to do this as envisioned, but my feeling is that a `range` over a map` should just go over the entire map as long as you've already acquired the lock.  This loop should be quite quick.

Yaiba (approved):
LGTM

PR #947, opened 2024-08-29
fix: parse bug with string literals

brennanjl:
This PR fixes a bug I found where string literals that end in an escaped single quote (e.g. `'hello world \''`) incorrectly remove the escaped quote.


Yaiba (approved):
LGTM.    We need exhaustive tests, maybe starting by add test on this case

brennanjl:
Will add a test

Yaiba (approved):

PR #949, opened 2024-08-30
types,rpc,client: extract ActionCall from transactions, not a payload

### LINKED ISSUE #904: core: remove PayloadTypeCallAction
`PayloadTypeCallAction` is misleading, since `PayloadType` is only used in transaction, and 'call' an action is not a transaction

jchappelow:
Resolves #904


jchappelow:
This was a quick one, so going back to draft so I can look at it again, but I want CI to run at least once.

brennanjl:
IMO this feels weird, since we are moving things into `types` that aren't needed. Previously, we had things in `types` if it was needed by end-users using the Go client. However, with this change, structs such as `ActionCall` and `EncodedValue`, which were only used internally by the client, are now boosted to this upper level.

I don't feel strongly opposed, but it did jump out to me as a big change in choosing what goes into `types`

jchappelow:
Yeah, I pretty much felt the same, and why I kept this one in limbo.  The competing awkwardness was `transactions.CallMessage` rather than `types.CallMessage` since it's got nothing to do with a transaction other than some superficial similarities.

I suppose you actually envisioned `CallMessage` becoming *more* similar to a transaction rather than less?  I feel like that's the confusing that lead to this false `Payload` classification for a `CallMessage` however.  Even the RLP `ActionCall` being treated as a "body" with a single "payload" field is a somewhat hard to swallow since unlike a transaction there's only one type ever in this "body".  There are many types of transactions and possible RLP-encoded content, but an action call is much more narrowly purposed.

I'll just put up a simple version that unpayloads the `CallMessage`, but leaves everything in the same place.

brennanjl:
> I suppose you actually envisioned CallMessage becoming more similar to a transaction rather than less?

Yeah exactly. In my mind, it is essentially a transaction that is not processed by consensus.

jchappelow:
> Yeah exactly. In my mind, it is essentially a transaction that is not processed by consensus.

I suspect this is an instance where I'm stuck in more conventional thinking (probably wrong or not quite meeting our design goals), namely an open network model.  Still, would you expand on your view of a call message as a "transaction that is not processed by consensus"?

On this subject, I'm really just considering Ethereum's "public methods" vs. those that require a transaction to use.  For an actual transaction, consensus level crypto is used to ensure the validity of the `sender` field.  For public methods you can just go to etherscan and call the method, or use the geth console with a forged caller, with no issues since it doesn't touch state.  Essentially, a simulation.

I acknowledge, we are more than Ethereum, and need to restrict such read-only methods, and consensus with nonces can't assist in preventing replay.  So we are working towards a secure solution to access control in https://github.com/kwilteam/kwil-db/pull/946 and it mostly make sense to me.   But I'm not sure that modeling a call request like a transaction will encourage a good view of the practical differences between the two (or make it straightforward and stable to use), where the guarantees of authenticity are established in very different places.

PR #953, opened 2024-09-03
moved kwild config and added it to common.Service

brennanjl:
This PR moves the main `kwild` config struct to `common/config`, and makes it accessible via `common.Service`. It also deprecates `common.Service.ExtensionConfig`, in favor of `common.Service.LocalConfig.AppConfig.Extensions`. Previously, we simply shallow copied between these two.

Finally, it changes how we pass `common.Service` around. We used to pass it around or build new ones haphazardly, and there were a few places where I found that we were not copying structs (e.g. the genesis config) properly. This has been an issue in the past, and so I figured we should just make it standard.


jchappelow:
LGTM, but will you include this commit with import renames and a little docs added: https://github.com/kwilteam/kwil-db/commit/c3c6f2d62141db280db644e354a8149e47ecddf2

jchappelow:
Also, what about the DefaultConfig function in common/config?  Does that go with the custom builds

brennanjl:
> Also, what about the DefaultConfig function in common/config? Does that go with the custom builds

I left it in `kwild/config` simply because I felt it is of no interest to the user. There is a function in https://github.com/kwilteam/kwil-db/pull/958 for custom builds that allows users to edit the default config.

jchappelow (approved):

PR #956, opened 2024-09-05
Array additions

brennanjl:
@outerlook highlighted some serious deficiencies with Kuneiform's arrays. This PR addresses those. There are 3 things specifically this PR adds:

- It adds "slices", e.g. `$my_arr[1:3]`, `$my_arr[1:]`, `$my_arr[3]`, etc.
- It adds Postgres's `array_remove` function
- It fixes a bug when comparing arrays of numeric types, e.g:
```
$arr := [1, 2, 3];

// this caused a type error at runtime
if $arr == [1, 2, 3] {
   // ...
}
```


parse/analyze.go
```diff
@@ -747,12 +755,31 @@ func (s *sqlAnalyzer) VisitExpressionArrayAccess(p0 *ExpressionArrayAccess) any
...
+		if !ok {
+			return s.expressionTypeErr(p0.Index)
+		}
+		if !idxAttr.Equals(types.IntType) {
+			return s.typeErr(p0.Index, idxAttr, types.IntType)
+		}
+	} else {
+		// if multiple indexes, result is an array
+		isArray = true
+		for _, idx := range p0.FromTo {
```
Yaiba, line 771: I think we need one more check, `leftIndex < rightIndex`, but it seems we cannot enforce it here?

parse/functions.go
```diff
@@ -300,6 +300,28 @@ var (
 				return fmt.Sprintf("array_length(%s, 1)", inputs[0]), nil
 			},
 		},
+		"array_remove": {
```
Yaiba, line 303: Can we construct tests again the functions? At least test `PGFormat`

Yaiba (commented):
Over all looks good. 
I'm not sure if we can force check leftIndex < rightIndex on slice operation, we should if we can

parse/analyze.go
brennanjl: 1. We can't enforce that
2. Postgres actually allows this, so it is a non-issue

parse/functions.go
brennanjl: We do, it is covered in an integration test. The integration test covers it. I felt this was most fitting because it tests that the generated postgres syntax is valid

parse/analyze.go
Yaiba: cool, then it's fine

parse/functions.go
Yaiba: Can you show me where this test is? I couldn't locate it.

Yaiba:
> It fixes a bug when comparing arrays of numeric types

Also, I don't know what change fix this, can you point it out?

internal/engine/integration/procedure_test.go
```diff
@@ -264,6 +264,35 @@ func Test_Procedures(t *testing.T) {
...
+				if $arr2 != [2, 3, 4] {
+					error('array slice failed');
+				}
+
+				$arr3 := array_prepend(0, $arr);
+				if $arr3 != [0, 1, 2, 3, 4] {
+					error('array_prepend failed');
+				}
+
+				if array_remove($arr3, 3) != [0, 1, 2, 4] {
```
brennanjl, line 287: @Yaiba the test is here

brennanjl:
> Also, I don't know what change fix this, can you point it out?

It is fixed by always typecasting literals when generating postgres. Previously, the Kuneiform array `[1, 2, 3]` would be generated as `ARRAY[1, 2, 3]`, and Postgres would freak out because it believes it is an INT4, while our ints are all INT8. Now, it generates `ARRAY[1::INT8, 2::INT8, 3::INT8]`

internal/engine/integration/procedure_test.go
Yaiba: Ok, so this test is against acutal postges.

Not a big deal now, but what about some tests on `PGFormat`?

brennanjl: This tests PGFormat.

PGFormat generates Postgres code for the function. The best way to test this IMO is to actually run the generated code against Postgres (which we are doing)

Yaiba:
> > Also, I don't know what change fix this, can you point it out?
> 
> It is fixed by always typecasting literals when generating postgres. Previously, the Kuneiform array `[1, 2, 3]` would be generated as `ARRAY[1, 2, 3]`, and Postgres would freak out because it believes it is an INT4, while our ints are all INT8. Now, it generates `ARRAY[1::INT8, 2::INT8, 3::INT8]`

I see, so it's the changes in `VisitExpressionLiteral` fix this.  

I just realized since we support slice, we should have one more test like `$arr := [1,2,3]; $arr2 := $arr1[:2];  THEN $arr1 != $arr2 `?  Or maybe you've already have tests like this?

internal/engine/integration/procedure_test.go
Yaiba: hmm, this is not obvious to me since we changed the output of some function's PGFormat in this PR, but no tests are changed.  I guess I'm talking about unit test.

Yaiba: I guess it's not a big issue, we do have this 'integration' tests

Yaiba (approved):

PR #962, opened 2024-09-09
deps (1/2): update go.mod core deps, next is parse

jchappelow:
We aren't branching or removing `replace` statements for the `v0.9.0-beta` release, but to assist custom builds, this updates the `core` entries in each go.mod with a *pseudo-version*.  This can also be done manually in the consuming project, but this provides the hint to get a non-broken build.

After this is merged, one more PR with updates for `parse` should follow.

As a reminder, with the `replace` statements remaining in the go.mod, files: (1) it is an error if trying to `go install` or `go run` one of the `cmd/...` apps, and (2) any `replace` statements get **ignored** if you are importing one of our packages like `core` or the main  module to build a custom app, like `tsn`.  Hence, `replace` statements will be removed for final/candidate release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/jackc/pglogrepl v0.0.0-20240307033717-828fbfe908e9
 	github.com/jackc/pgx/v5 v5.6.0
 	github.com/jpillora/backoff v1.0.0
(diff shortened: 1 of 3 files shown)


Yaiba (approved):
LGTM

PR #970, opened 2024-09-11
added latest snapshot height querying

brennanjl:
As discussed earlier, we hit a sync issue with TSN testnet due to Comet trying to verify too many block headers. Since our snapshot providers are trusted anyways, this isn't necessary. I added the ability to query ABCI for the latest snapshot from a node, so that we can avoid this.

This is untested, but I just wanted to get something for it because I need to get going and I didn't want to lose my train of thought on it. There is a TODO in there asking about why we don't return errors in a certain part of ABCI (CC @charithabandi you might be able to answer this for me)

### DIFF
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -1594,12 +1594,10 @@ func (a *AbciApp) ProcessProposal(ctx context.Context, req *abciTypes.RequestPro
 
 func (a *AbciApp) Query(ctx context.Context, req *abciTypes.RequestQuery) (*abciTypes.ResponseQuery, error) {
 	a.log.Debug("ABCI Query", zap.String("path", req.Path), zap.String("data", string(req.Data)))
-
 	switch {
-	// TODO: why do we not return errors in case of the snapshotter not being configured?
 	case req.Path == statesync.ABCISnapshotQueryPath:
 		if a.snapshotter == nil {
-			return &abciTypes.ResponseQuery{}, nil
+			return nil, fmt.Errorf("this node is not configured to serve snapshots")
 		}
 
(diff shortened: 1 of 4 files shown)


jchappelow:
@charithabandi for background: https://github.com/cometbft/cometbft/issues/742
The gap between trusted height and the latest available snapshot height determines how many block headers get pulled for validation.  I'm pretty certain trusted height (like a checkpoint height in btc) is expected to be below the available snapshot heights.  What we found is that if the trusted height was the current block (higher than latest snapshot height), for some reason it would pull and ungodly amount of data when verifying headers.

internal/abci/abci.go
```diff
@@ -1595,7 +1595,9 @@ func (a *AbciApp) ProcessProposal(ctx context.Context, req *abciTypes.RequestPro
 func (a *AbciApp) Query(ctx context.Context, req *abciTypes.RequestQuery) (*abciTypes.ResponseQuery, error) {
 	a.log.Debug("ABCI Query", zap.String("path", req.Path), zap.String("data", string(req.Data)))
 
-	if req.Path == statesync.ABCISnapshotQueryPath { // "/snapshot/height"
+	switch {
+	// TODO: why do we not return errors in case of the snapshotter not being configured?
```
charithabandi, line 1599: Umm, I don't see a major issue with erroring out here. Let's add that.

charithabandi (dismissed):
LGTM!

brennanjl:
I will add in errors

brennanjl:
Ok I added errors to abci query @charithabandi. I tested them out and it works as expected (it just returns the error to the client)

charithabandi (approved):

PR #972, opened 2024-09-11
add RPC health check, require Go 1.22

jchappelow:
This PR has two distinct but related commits (this should merge via rebase, not squash):

https://github.com/kwilteam/kwil-db/commit/07fb514cfdbbfa5be0ba69bdc0c6b8be2eda0466

```
go mod: require at least go 1.22, use reflect.TypeFor, etc.

This updates the go.mod files and other CI config to use Go 1.22,
which is now the minimum required version.

Code changes:

- use reflect.TypeFor
- use math/rand/v2 in core/utils/random, and add New2()
- use range over integer where applicable
- change "exportloopref" linter to "copyloopvar" since loop variables
  are no longer reused
```

https://github.com/kwilteam/kwil-db/commit/36cff189e3764716aeaab04d3524454949c84937
```
rpc: add health check endpoints and service methods

This adds health check endpoints and JSON-RPC service methods:

- GET /api/v1/health/{svc}

  This endpoint returns the health status of the service in {svc}.
  For example, "/api/v1/health/user".
  The response code is 200 if healthy, otherwise 503 (unavailable).
  The response body is JSON serialization of a HealthResponse type
  in the corresponding core/rpc/{svc} Go package.

- GET /api/v1/health

  This RESTish endpoint returns a core/rpc/json.HealthResponse.
  The code is 200 if all registered services report healthy and
  503 if any of the services are not healthy.
  This response body includes a fingerprint boolean ("kwil_alive")
  and the aggregate health ("healthy"), which corresponds to the
  response code.
  There is a "services" field that is an object keyed by service
  names. The individual service health status objects are those
  returned by the service-specific "/api/v1/health/{svc}" endpoint
  describe above.

- 'rpc.health' JSON-RPC method. This method in the reserved 'rpc'
  namespace corresponds to the aggregate health endpoint.

- 'user.health' and 'admin.health' JSON-RPC methods. These correspond
  to the "/api/v1/health/user" and "/api/v1/health/admin" endpoints.

Reminder: The HTTP server on port 8484, which is used for user facing
RPCs, has only the "user" and "function" services registered. Thus,
and request for admin health on this server will result in a 404 or
method not found.  The admin server, which is on a unix socket by
default or port 8485, will have all of "user", "function", and "admin"
health statuses reported. While it's less likely to have a health
check from a cloud provider for this secure server, it is available
for secure monitoring setups on the host machine.
```


jchappelow:
Still testing this with AWS Route 53 health checks.  Also need a few test in the code for this, maybe acceptance but I don't want CI to get even slower.

Yaiba (approved):
LGTM

jchappelow:
@Yaiba Could I trouble you for a re-approval?

Yaiba:
> @Yaiba Could I trouble you for a re-approval?

oh My bad, forgot about this. I'll do a quick scan on new changes.

jchappelow:
```
// HealthResponse is the response for MethodHealth. This determines the
// serialized response for the Health method required by the rpcserver.Svc
// interface. This is the response with which most health checks will be concerned.
type HealthResponse struct {
	// Healthy is is based on several factors determined by the service and it's
	// configuration, such as the maximum age of the best block and if the node
	// is still syncing (in catch-up or replay).
	Healthy bool `json:"healthy"`

	// Version is the service API version.
	Version string `json:"version"`

	ChainInfoResponse
	BlockTimestamp int64          `json:"block_time"` // epoch millis
	BlockAge       int64          `json:"block_age"`  // milliseconds
	Syncing        bool           `json:"syncing"`
	AppHeight      int64          `json:"app_height"` // may be less than block store best block
	AppHash        types.HexBytes `json:"app_hash"`
	PeerCount      int            `json:"peer_count"`

	// Mode is an oddball field as it pertains to the service config rather than
	// state of the node. It is provided here as a convenience so applications
	// can discern node state and the mode of interaction with one request.
	Mode ServiceMode `json:"mode"` // e.g. "private"
}
```

PR #973, opened 2024-09-11
log: rotate and compress log file

jchappelow:
The idea is this:

```
-rw-r--r--   1 jon jon   18433 Sep 12 11:43 kwild.log
-rw-r--r--   1 jon jon    3296 Sep 11 17:50 kwild.log.1.gz
-rw-r--r--   1 jon jon    3596 Sep 11 17:52 kwild.log.2.gz
-rw-r--r--   1 jon jon    3323 Sep 11 17:55 kwild.log.3.gz
-rw-r--r--   1 jon jon    3296 Sep 11 17:57 kwild.log.4.gz
-rw-r--r--   1 jon jon    3301 Sep 11 18:00 kwild.log.5.gz
-rw-r--r--   1 jon jon    3351 Sep 11 18:03 kwild.log.6.gz
-rw-r--r--   1 jon jon    3309 Sep 11 18:06 kwild.log.7.gz
etc.
```

There is one new setting: `--log.file-roll-size`, which is a number in KB at which the current log file will be compressed, named the next gz in the sequence, and a new uncompressed log file started.  All are retained, but the rotator I'm using has a configurable limit that we could expose if need be.


core/log/logger.go
```diff
@@ -120,6 +135,14 @@ const (
 	TimeEncodingEpochFloat   = "epochfloat"
 )
 
+const (
+	// defaults for log data retained:
+	// - 6 GB total uncompressed in 100 (minus) gzipped files
+	// - up to 60 MB in current uncompressed log
+	maxLogRolls  = 100
```
brennanjl, line 142: Just to confirm, if we hit 101 rolls, will the 1st be deleted?

If so, should we just make this a configurable value off the bat?

Otherwise lgtm.

jchappelow: > Just to confirm, if we hit 101 rolls, will the 1st be deleted?

That's correct.  If I set the threshold to 1KB and this max rolls to 5, the oldest is deleted when the 6th gz is created:

```
-rw-r--r--   1 jon jon   773 Sep 16 16:06 kwild.log
-rw-r--r--   1 jon jon   783 Sep 16 16:06 kwild.log.2.gz
-rw-r--r--   1 jon jon   121 Sep 16 16:06 kwild.log.3.gz
-rw-r--r--   1 jon jon   512 Sep 16 16:06 kwild.log.4.gz
-rw-r--r--   1 jon jon   556 Sep 16 16:06 kwild.log.5.gz
-rw-r--r--   1 jon jon   654 Sep 16 16:06 kwild.log.6.gz
```

> If so, should we just make this a configurable value off the bat?

Yeah, happy to do that.  I had started this draft with it as a configurable flag, but removed it for simplicity, and I wasn't entirely sure old log deletion would be broadly desired.  That said, my thinking about this was to have 0 mean "retain all" (and be the default).

brennanjl: > Yeah, happy to do that. I had started this draft with it as a configurable flag, but removed it for simplicity, and I wasn't entirely sure old log deletion would be broadly desired. That said, my thinking about this was to have 0 mean "retain all" (and be the default).

I would imagine it would be, simply because Kwil produces a lot of logs, and it will probably get annoying for operators.

brennanjl (approved):

PR #974, opened 2024-09-12
Optimize the migration process on the new network

charithabandi:
This PR optimizes the migration process on the new network, by not submitting the resolutions with empty changesets thereby reducing the number of resolutions to be processed and voted on the new network.

Also fixes any issues concerning if the changeset migration resolution can fit within a block. All the changeset chunks are readjusted to the blocksize/3 on the new network.

### DIFF
--- a/internal/migrations/changesets.go
+++ b/internal/migrations/changesets.go
@@ -23,7 +23,7 @@ var (
 	ErrNoMoreChunksToRead = errors.New("no more chunks to read")
 	ErrChangesetNotFound  = errors.New("changeset not found")
 
-	migrationCompleted = "migration completed. `migration` section of `genesis.json` and `migrate_from` config in `config.toml` is no longer relevant and should be removed"
+	migrationCompleted = "Migration completed. The migration section in genesis.json and the migrate_from configuration in config.toml are no longer relevant and should be removed."
 )
 
 func init() {
--- a/internal/migrations/changesets.go
+++ b/internal/migrations/changesets.go
@@ -58,6 +58,9 @@ type changesetMigration struct {
 
 	// Changeset is the serialized changeset chunk.
 	Changeset []byte
+
+	// PreviousBlock is the last block with non empty changesets.
+	PreviousBlock uint64
 }
 
 // MarshalBinary marshals the ChangesetMigration into a binary format.
@@ -119,7 +122,6 @@ func applyChangesets(ctx context.Context, app *common.App, blockCtx *common.Bloc
 	}
 	defer tx.Rollback(ctx)
 
-	var currentHeight int64
 	// get the last changeset height applied
 	lastChangeset, err := getLastChangeset(ctx, tx)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


internal/voting/sql.go
```diff
@@ -81,16 +81,42 @@ const (
 		height INT NOT NULL
 	);`
 
-	// upsertResolution is the sql statement used to ensure a resolution is
+	sqlInsertResolution = `CREATE OR REPLACE FUNCTION ` + votingSchemaName + `.insert_resolution(
```
jchappelow, line 84: i don't know if this is more or less silly, but:

```sql
WITH already_processed AS (
    SELECT EXISTS (
        SELECT 1 
        FROM kwild_voting.processed 
        WHERE id = v_id
    ) AS v_already_processed
)
-- Attempt to insert the resolution only if it's not already processed
INSERT INTO kwild_voting.resolutions (id, body, type, expiration, vote_body_proposer)
SELECT
    v_id,
    v_body,
    (SELECT id FROM kwild_voting.resolution_types WHERE name = v_type),
    v_expiration,
    v_vote_body_proposer
FROM already_processed
WHERE NOT v_already_processed;
```

charithabandi: This doesn't error out if already processed. Do we want it to error out? Event store currently just ignores if processed. But if its an explicit transaction, we probably wanna fail right?

jchappelow: Ah, true.  This might as well have been an on conflict do nothing, except that it is a legit error condition.

internal/migrations/changesets.go
```diff
@@ -132,49 +134,54 @@ func applyChangesets(ctx context.Context, app *common.App, blockCtx *common.Bloc
...
 			break
 		}
 
-		// Check if all chunks have been received for the current height, if not, break
-		totalChunks, chunksReceived, err := getChangesetMetadata(ctx, tx, currentHeight)
-		if err != nil {
-			return err
+		if height != startHeight && prevHeight != lastChangeset {
+			// If the previous changeset is not applied, break
+			break
```
jchappelow, line 154: Is this normal to happen?  if the previous changeset was not applied, when would it be applied?  After it is retrieved and in some later block's end block hook?

internal/migrations/migrator.go
```diff
@@ -687,9 +688,11 @@ func (m *Migrator) StoreChangesets(height int64, changes <-chan any) {
 	bs := &BlockSpends{
 		Spends: m.accounts.GetBlockSpends(),
 	}
-	if pg.StreamElement(chunkWriter, bs); err != nil {
-		m.errChan <- err
-		return
+	if bs.Spends != nil {
```
jchappelow, line 691: a `len(bs.Spends) > 0` check would be safer here.

internal/migrations/sql.go
```diff
@@ -195,7 +195,8 @@ var (
 	tableChangesetsMetadataSQL = `CREATE TABLE IF NOT EXISTS ` + migrationsSchemaName + `.changesets_metadata (
 		height INT PRIMARY KEY,
 		total_chunks INT, -- total number of chunks in the changeset
-		received INT -- number of chunks received
+		received INT, -- number of chunks received
+		prev_height INT -- height of the previous changeset
```
jchappelow, line 199: We should probably have heights use bigint / int8.  A 4 byte integer is risky.

internal/migrations/changesets.go
charithabandi: Yeah, it will be applied once we receive all the chunks for that block.

internal/migrations/sql.go
charithabandi: oh yeah, makes sense.

internal/migrations/migrator.go
```diff
@@ -640,7 +641,7 @@ func (cw *chunkWriter) SaveMetadata() error {
 	filename := formatChangesetMetadataFilename(cw.dir, cw.height)
 	metadata := &ChangesetMetadata{
 		Height:     cw.height,
-		Chunks:     cw.chunkIdx + 1,
+		Chunks:     cw.chunkIdx + 1, // + 1 ?
```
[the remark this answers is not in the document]
charithabandi: Yeah, looks like a stale comment, I changed the indexing bunch of times.

jchappelow:
For merged commit message when CI is done:

```
migrations,txapp,voting: optimize migration by omitting empty changesets

This optimizes the migration process on the new network by not submitting
the resolutions with empty changesets thereby reducing the number of
resolutions to be processed and voted on the new network.

Also fixes any issues concerning if the changeset migration resolution
can fit within a block. All the changeset chunks are readjusted to the
blocksize/3 on the new network.
```

jchappelow (approved):
Migration from staging network X to staging network Y complete successfully (except the final changeset edge case that is now fixed).  :partying_face:

PR #975, opened 2024-09-12
added kwil-cli utils generate-key

brennanjl:
Adds a utility to kwil-cli that allows users to generate ECDSA keys

### DIFF
--- a/cmd/common/generate/generate.go
+++ b/cmd/common/generate/generate.go
@@ -22,23 +22,23 @@ func WriteDocs(cmd *cobra.Command, dir string) error {
 		}
 
 		return os.Create(fullPath)
-	})
+	}, true)
 }
 
-func writeCmd(cmd *cobra.Command, dir string, idx int, write writeFunc) error {
+func writeCmd(cmd *cobra.Command, dir string, idx int, write writeFunc, first bool) error {
 	subCommands := cmd.Commands()
 	if len(subCommands) == 0 {
 		return createCmdFile(cmd, dir, idx, write)
 	}
 
-	err := createIndexFile(cmd, dir, write)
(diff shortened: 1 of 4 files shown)


brennanjl:
I added this because our docs currently tell users to go to a website that generates private keys for you: https://docs.kwil.com/docs/getting-started/create-a-database

This is very insecure, and so to avoid a user then using this key elsewhere, I figured it's best to just add this.

jchappelow:
I have been using https://github.com/jchappelow/kwilkeygen/blob/main/keygen.go

```
$  go run github.com/jchappelow/kwilkeygen@main
Key: 463273bc5457d883831db7dd3287b72bf8a0155485b44f89e6abb38e432cde37
Pub: 04222ef1da12668ae5bea19f54c7b3c957f6dc31ea70f3f11efe52a7a61ea0560e6125c35a90f30d3007482b0843c7426d89eb5664956af7de77b95e474dcb8d68
Identity: 038da88d40af7862832f27379e01faf75f9655b5
```

So this command is a welcome addition.

Only comment is that it is documented as producing a key **pair** but it only shows the private key, no public key or address/identity.

brennanjl:
> Only comment is that it is documented as producing a key pair but it only shows the private key, no public key or address/identity.

Ah. I'll have it log the address too

brennanjl:
I made some additions to our doc autogeneration. Previously we had to make these changes manually after each generation, but now it is automatic.

jchappelow (approved):

jchappelow (approved):

jchappelow:
When merging this one, will use rebase button for so the generate commit remains separate.

PR #977, opened 2024-09-12
deprecates --action, should now be passed as positional argument

brennanjl:
While going through the docs, I noticed that we liberally use `--action` everywhere with `kwil-cli`. This isn't quite accurate, since anywhere we specify `--action` we can also use a procedure. Therefore, I deprecated `--action`, and added a `--target` flag.

### DIFF
--- a/cmd/kwil-cli/cmds/database/call.go
+++ b/cmd/kwil-cli/cmds/database/call.go
@@ -18,17 +18,17 @@ import (
 )
 
 var (
-	callLong = `Call a ` + "`" + `view` + "`" + ` action, returning the result.
+	callLong = `Call a ` + "`" + `view` + "`" + ` procedure or action, returning the result.
 
-` + "`" + `view` + "`" + ` actions are read-only actions that do not require gas to execute.  They are
+` + "`" + `view` + "`" + ` procedure are read-only procedure that do not require gas to execute.  They are
 the primary way to query the state of a database. The ` + "`" + `call` + "`" + ` command is used to call
-a ` + "`" + `view` + "`" + ` action on a database.  It takes the action name as a required flag, and the
-action inputs as arguments.
+a ` + "`" + `view` + "`" + ` procedure on a database.  It takes the procedure name as a required flag, and the
+procedure inputs as arguments.
 
-To specify an action input, you first need to specify the input name, then the input value, delimited by a colon.
-For example, for action ` + "`" + `get_user($username)` + "`" + `, you would specify the action as follows:
+To specify a procedure input, you first need to specify the input name, then the input value, delimited by a colon.
+For example, for procedure ` + "`" + `get_user($username)` + "`" + `, you would specify the procedure as follows:
 
-` + "`" + `username:satoshi` + "`" + ` --action=get_user
+` + "`" + `username:satoshi` + "`" + ` --target=get_user
 
 You can either specify the database to execute this against with the ` + "`" + `--name` + "`" + ` and ` + "`" + `--owner` + "`" + `
 flags, or you can specify the database by passing the database id with the ` + "`" + `--dbid` + "`" + ` flag.  If a ` + "`" + `--name` + "`" + `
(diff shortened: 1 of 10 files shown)


jchappelow:
Makes sense to deprecate based on outdated terminology.  Not sure about `--target` though, yet I don't have a better suggestion.  does method or procedure not work?

Yaiba:
> Makes sense to deprecate based on outdated terminology. Not sure about `--target` though, yet I don't have a better suggestion. does method or procedure not work?

I feel the same about `--target`, target feels more about a DB, not a procedure/action.

brennanjl:
I thought about `--procedure`; originally, I just added a `--procedure` flag that was interoperable with `--action`. However, I felt it was sort've weird to have two of the same. As for `--method`, I'm hesitant to use it since we don't use it anywhere else.

If you guys like having interchangeable `--action` and `--procedure` flags, we could do that.

Yaiba:
> If you guys like having interchangeable --action and --procedure flags, we could do that.

I'm not sure about this, IMO this could be more clear or more confusing to different users :/

brennanjl:
I'm now coming around to standardizing the term `method` as either a procedure or action. Do you think this works @Yaiba?

Yaiba:
> I'm now coming around to standardizing the term `method` as either a procedure or action. Do you think this works @Yaiba?

It kind of works for me, if you mean use `method` in docs and kwil-cli

brennanjl:
> It kind of works for me, if you mean use method in docs and kwil-cli

Yes, exactly.

brennanjl:
Ok, it now takes the procedure/action name as the first positional param. It is backwards compatible with `--action`

cmd/kwil-cli/cmds/database/call.go
```diff
@@ -18,33 +18,32 @@ import (
...
 
 You can either specify the database to execute this against with the ` + "`" + `--name` + "`" + ` and ` + "`" + `--owner` + "`" + `
 flags, or you can specify the database by passing the database id with the ` + "`" + `--dbid` + "`" + ` flag.  If a ` + "`" + `--name` + "`" + `
 flag is passed and no ` + "`" + `--owner` + "`" + ` flag is passed, the owner will be inferred from your configured wallet.
 
 If you are interacting with a Kwil gateway, you can also pass the ` + "`" + `--authenticate` + "`" + ` flag to authenticate the call with your private key.`
 
-	callExample = `# Calling the ` + "`" + `get_user($username)` + "`" + ` action on the "mydb" database
-kwil-cli database call --action get_user --name mydb --owner 0x9228624C3185FCBcf24c1c9dB76D8Bef5f5DAd64 username:satoshi
+	callExample = `# Calling the ` + "`" + `get_user($username)` + "`" + ` procedure on the "mydb" database
```
Yaiba, line 39: NBD, maybe `get_user($username) view`?

cmd/kwil-cli/cmds/database/read_schema.go
```diff
@@ -49,8 +49,6 @@ func readSchemaCmd() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringP(nameFlag, "n", "", "the target database name")
-	cmd.Flags().StringP(ownerFlag, "o", "", "the target database owner")
-	cmd.Flags().StringP(dbidFlag, "i", "", "the target database id")
+	bindFlagsTargetingDatabase(cmd)
```
Yaiba, line 52: the description of this command is indented, better keep it the same as other cmds.

Yaiba (dismissed):
LGTM, I left commnets two minor issues.

I think we should make the similiar changes to other `database` sub-cmds, to use positional arguments instead of flags(well, flags should still work, but deprecated).  IMO, , those changes are gonna happen in a future version , so why not just do it in this version

cmd/kwil-cli/cmds/database/call.go
brennanjl: Eh, I think it is less clear

cmd/kwil-cli/cmds/database/read_schema.go
brennanjl: good catch

brennanjl:
> IMO, , those changes are gonna happen in a future version , so why not just do it in this version

I'll do this for deploy

cmd/kwil-cli/cmds/database/call.go
Yaiba: It's calling a view action/procedure, but nbd.

cmd/kwil-cli/cmds/database/execute.go
```diff
@@ -15,55 +15,56 @@ import (
...
+		Short:   "Execute a procedure or action against a database.",
 		Long:    executeLong,
 		Example: executeExample,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return common.DialClient(cmd.Context(), cmd, 0, func(ctx context.Context, cl clientType.Client, conf *config.KwilCliConfig) error {
-				dbId, err := getSelectedDbid(cmd, conf)
+				dbid, err := getSelectedDbid(cmd, conf)
 				if err != nil {
-					return display.PrintErr(cmd, fmt.Errorf("target database not properly specified: %w", err))
+					return display.PrintErr(cmd, fmt.Errorf("error getting selected procedure and dbid: %w", err))
```
Yaiba, line 47: seems `error getting dbid` is more accurate?

cmd/kwil-cli/cmds/database/flags.go
```diff
@@ -15,6 +16,7 @@ const (
 	nameFlag       = "name"
 	ownerFlag      = "owner"
 	actionNameFlag = "action"
+	targetFlag     = "target"
```
Yaiba, line 19: remove this?

Yaiba (changes requested):

Yaiba (dismissed):
LGTM

brennanjl:
@Yaiba re-approve?

Yaiba (approved):

PR #980, opened 2024-09-14
Fixes the pg_dump issues with create functions

charithabandi:


cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -233,8 +233,28 @@ func pgDump(ctx context.Context, dbName, dbUser, dbPass, dbHost, dbPort string,
 				Name:   fmt.Sprintf("validator-%d", validatorCount),
 			})
 			validatorCount++
+		} else if inFunctionBlock {
+			if strings.HasPrefix(line, "$$;") {
```
jchappelow, line 237: I'm trying to find documentation about this:
- can there be other functions that end in something other than `$$`;?  like functions that happen to be in the dump that we didn't define for procedures.  We might scan all the way to the end, or just too far (until the next function end).
- Why prefix vs suffix vs equals (ignoring leading/trailing space)?
- we often define functions ending with `$$ LANGUAGE plpgsql;` but I guess that gets simplified.

jchappelow: https://www.postgresql.org/docs/current/plpgsql-structure.html

jchappelow: Fortunately postgresql will hang/error if trying to do this:

```
CREATE FUNCTION test_func() RETURNS text AS $$
BEGIN
    RETURN 'This string contains $$ inside the body';
END;
$$ LANGUAGE plpgsql;
```

The only way to do that is to change the delimiter (`$$`) to something else:

```
CREATE FUNCTION test_func() RETURNS text AS $delim$
BEGIN
    RETURN 'This string contains $$ inside the body';
END;
$delim$ LANGUAGE plpgsql;
```

When we are crafting the `CREATE FUNCTION` in kwil, we use `$$`, so the function body cannot contain "$$" anywhere.

---

The remaining question I have is will pg_dump include any other system created functions that might use a different delimiter.  (It creates a bunch of selects and sets to get things right like locate, so who knows if it's also going to create a function.

I think this will work for now, but we should document this pretty clearly.

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -233,8 +233,28 @@ func pgDump(ctx context.Context, dbName, dbUser, dbPass, dbHost, dbPort string,
...
+
+			// Write the line to the output file
+			n, err := multiWriter.Write([]byte(line + "\n"))
+			if err != nil {
+				return -1, nil, fmt.Errorf("failed to write to gzip writer: %w", err)
+			}
+			totalBytes += int64(n)
 		} else {
-			if line == "" || strings.TrimSpace(line) == "" { // Skip empty lines
+			if strings.HasPrefix(line, "CREATE FUNCTION") {
```
jchappelow, line 248: We'd miss `CREATE OR REPLACE FUNCTION`, right?

cmd/kwil-admin/cmds/snapshot/create.go
jchappelow:
```
// create_user creates a user in the database.
// It is assigned a unique uuid.
// MUTABLE
procedure create_user($name text, $age int) public {
    // we will generate a uuid from the txid
    notice('stuff $$ adsf');
    INSERT INTO users (id, name, age, address)
    VALUES (uuid_generate_v5('985b93a4-2045-44d6-bde4-442a4e498bc6'::uuid, @txid), $name, $age, @caller);
}
```

on deploy:

`failed to deploy database, code = 4294967295, log = "ERROR: syntax error at or near \"adsf\" (SQLSTATE 42601)"`

cmd/kwil-admin/cmds/snapshot/create.go
jchappelow:
```
procedure create_user($name text, $age int) public {
    // we will generate a uuid from the txid
    notice('stuff $$; create function adsf() returns text AS $asd$ select 1::int8; ');
    // malice here... how?
    notice('$asd$;');
}
```

Can someone get creative and create an arbitrary *valid* function like this @brennanjl ?

brennanjl: Yes. https://github.com/kwilteam/kwil-db/pull/981 fixes this. Since this edge case exists on preview right now, I think we should keep it since it is a breaking consensus change, and only put the fix on v0.9.

brennanjl:
Still reviewing, but is there any way to consolidate the sanitization logic between the two pgdump calls? It seems like they are _very_ similar. If not, nbd, it just feels weird to have them be so similar and be duplicated.

brennanjl:
Discussed in-person with @charithabandi but just for completeness:

We are going to go ahead with replicating the internal tables as-is. This means:
1. Internal tables cannot be changed in a patch version.
2. Users will get error warnings during replication since it will fail to create the internal tables (which already exist).

brennanjl (approved):

PR #981, opened 2024-09-16
check for invalid delimiter

brennanjl:
This adds a delimiter `$kwil_reserved_delim$`, and checks that users aren't trying to maliciously use it in procedure or table creation.

We should not backport this to the TSN fix, since it is a breaking change with how snapshots are created.

### DIFF
--- a/internal/engine/generate/generate.go
+++ b/internal/engine/generate/generate.go
@@ -22,6 +22,12 @@ func GenerateDDL(pgSchema string, table *types.Table) ([]string, error) {
 	}
 	statements = append(statements, createIndexStatements...)
 
+	for _, stmt := range statements {
+		if containsDisallowedDelimiter(stmt) {
+			return nil, fmt.Errorf("statement contains disallowed delimiter: %s", stmt)
+		}
+	}
+
 	return statements, nil
 }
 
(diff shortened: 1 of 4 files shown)


brennanjl:
One pretty big issue we might have here is upgradability.

Any old schema will be pgdumped with the `$$` delimiter. I guess its not too big a deal since it won't break anything, but its something we should keep in mind.

jchappelow (approved):
LGTM

jchappelow:
> Any old schema will be pgdumped with the `$$` delimiter. I guess its not too big a deal since it won't break anything, but its something we should keep in mind.

Mmmm, no ideal but I also imagine that's fine, particularly given there has been no real working snapshots yet anyway.

Still, just throwing this out there -- dump sanitize could do a string replacement, but we risk messing up the dump if not done just so.

brennanjl:
> Mmmm, not ideal but I also imagine that's fine, particularly given there has been no real working snapshots yet anyway.

Still, just throwing this out there -- dump sanitize could do a string replacement, but we risk messing up the dump if not done just so.

The thing is, we only want to replace the string for dumps that are using the current `$$` delimiter, instead of `$kwil_reserved_delim$`. IMO it seems more error prone, especially since dumps using the current delimiter would work find anyways.

PR #982, opened 2024-09-16
CI: use app health check endpoint

### LINKED ISSUE #967: flakey CI -- chain_info not found
We have been living with frequent CI failures for quite some time.  The most common failure that appears to be the fault of the CI system is `wrap client: chain_info: not found`.  For instance: https://github.com/kwilteam/kwil-db/actions/runs/10780424528/job/29896165355#step:22:1874

There have been many attempts to fix this to no avail.  We got this with the old gRPC/HTTP server and with the new JSON-RPC server.

I had experimented with a workaround, but it's not really the right fix: https://github.com/kwilteam/kwil-db/pull/641

jchappelow:
This uses the new health check endpoints with CI to get a more meaningful idea of node container readiness.
This is an effort to resolve #967


jchappelow:
:crossed_fingers:

PR #985, opened 2024-09-17
Cleanup snapshots, statesync and whitelist peers

charithabandi:
- Remove max-row-size config and update snapshot creation process to handle any row data sizes
- unexpose(removed) trust options in the statesync config as they are not user configurable. 
- Rename `kwil-admin peers` to `kwil-admin whitelist`

### DIFF
--- a/cmd/kwil-admin/cmds/setup/init.go
+++ b/cmd/kwil-admin/cmds/setup/init.go
@@ -66,6 +66,20 @@ func initCmd() *cobra.Command {
 			}
 			cfg.RootDir = expandedDir
 
+			// saves genesis state snapshot file in the root directory under the name "genesis_state.sql.gz"
+			if genesisState != "" {
+				if genesisState, err = common.ExpandPath(genesisState); err != nil {
+					return display.PrintErr(cmd, err)
+				}
+
+				stateFile := filepath.Join(expandedDir, config.GenesisStateFileName)
+				if err = copyFile(genesisState, stateFile); err != nil {
+					return display.PrintErr(cmd, err)
+				}
+
+				cfg.AppConfig.GenesisState = config.GenesisStateFileName
+			}
+
 			// saves config and private key files in the root directory
 			pub, err := nodecfg.GenerateNodeFiles(expandedDir, cfg, true)
 			if err != nil {
@@ -99,18 +113,6 @@ func initCmd() *cobra.Command {
 					if err = genesisCfg.SaveAs(genFile); err != nil {
 						return display.PrintErr(cmd, err)
 					}
-				}
-			}
-
-			// saves genesis state snapshot file in the root directory under the name "genesis_state.sql.gz"
-			if genesisState != "" {
-				if genesisState, err = common.ExpandPath(genesisState); err != nil {
-					return display.PrintErr(cmd, err)
-				}
-
-				stateFile := filepath.Join(expandedDir, config.GenesisStateFileName)
-				if err = copyFile(genesisState, stateFile); err != nil {
-					return display.PrintErr(cmd, err)
 				}
 			}
 
--- a/cmd/kwil-admin/cmds/setup/init.go
+++ b/cmd/kwil-admin/cmds/setup/init.go
@@ -29,7 +29,7 @@ kwil-admin setup init -o ~/.kwild-new`
 )
 
 func initCmd() *cobra.Command {
-	var out, chainId, genesisPath string
+	var out, chainId, genesisPath, genesisState string
 	var blockInterval time.Duration
 	var joinExpiry int64 // block height
 	var withGas bool
@@ -71,6 +71,7 @@ func initCmd() *cobra.Command {
 				return display.PrintErr(cmd, err)
 			}
 
+			// copies the genesis.json file from the --genesis-file to the root directory
 			genFile := filepath.Join(expandedDir, cometbft.GenesisJSONName)
 			if genesisPath != "" {
 				if genesisPath, err = common.ExpandPath(genesisPath); err != nil {
@@ -105,6 +106,25 @@ func initCmd() *cobra.Command {
 					}
 				}
 			}
+
+			// saves genesis state snapshot file in the root directory under the name "genesis_state.sql.gz"
+			if genesisState != "" {
+				if genesisState, err = common.ExpandPath(genesisState); err != nil {
+					return display.PrintErr(cmd, err)
+				}
+
+				file, err := os.ReadFile(genesisState)
+				if err != nil {
+					return display.PrintErr(cmd, err)
+				}
+
+				stateFile := filepath.Join(expandedDir, config.GenesisStateFileName)
+				err = os.WriteFile(stateFile, file, 0644)
+				if err != nil {
+					return display.PrintErr(cmd, err)
+				}
+			}
+
 			return display.PrintCmd(cmd, display.RespString("Initialized node in "+expandedDir))
 		},
 	}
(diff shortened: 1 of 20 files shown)


cmd/kwil-admin/cmds/whitelist/cmd.go
```diff
@@ -1,14 +1,13 @@
...
+package whitelist
 
 import "github.com/spf13/cobra"
 
 var peersCmd = &cobra.Command{
-	Use:     "peers",
-	Short:   "manages the node's peers",
-	Aliases: []string{"peer"},
+	Use:   "whitelist",
+	Short: "manages the node's whitelist peers",
```
brennanjl, line 7: Small nit, but can you make sure you have consistent grammar for all of your user-facing docs? For example, in this command, you don't capitalize the first letter of the sentence, nor do you have punctuation. In another command, you capitalize the first letter, but don't have punctuation, and in another command, you do have punctuation and you capitalize the first letter.

We should make sure to have proper grammar is user facing docs like this, so please capitalize the first letter, add punctuation, etc. Also, make sure that they are complete sentences: e.g. "The `peer` command is used to manage a node's peer whitelist".

brennanjl (commented):
Other than the grammar, lgtm. @jchappelow no rush but when convenient, could you take a quick scan at the scanner stuff? It seems ok to me, but I'm not totally sure on the right way to do it.

cmd/kwil-admin/cmds/setup/init.go
```diff
@@ -105,6 +106,25 @@ func initCmd() *cobra.Command {
...
+				file, err := os.ReadFile(genesisState)
+				if err != nil {
+					return display.PrintErr(cmd, err)
+				}
+
+				stateFile := filepath.Join(expandedDir, config.GenesisStateFileName)
+				err = os.WriteFile(stateFile, file, 0644)
+				if err != nil {
+					return display.PrintErr(cmd, err)
+				}
```
brennanjl, line 125: We should also set the `kwild` command's genesis state config to point to the newly copied config

charithabandi: genesis-state has now a default value set to `genesis-state.sql.gz`

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -191,32 +193,37 @@ func pgDump(ctx context.Context, dbName, dbUser, dbPass, dbHost, dbPort string,
...
 
-	// Pass the output of pg_dump through scanner to sanitize it
-	buf := make([]byte, maxRowSize)
-	scanner := bufio.NewScanner(pgDumpOutput)
-	scanner.Buffer(buf, maxRowSize)
+	// Sanitize the output of pg_dump to include only the necessary tables and data
+	reader := bufio.NewReader(pgDumpOutput)
+
+	for {
+		line, err := reader.ReadString('\n')
```
jchappelow, line 200: It'll be important that the text dump is newline-terminated, otherwise we'll lose the last line.

Although, didn't we just need to give a very large max to the `scanner.Buffer` call?

Other thing is that I was under the impression that a `bufio.Reader` would also return a `bufio.ErrBufferFull` if it's buffer size was exceeded.  Did we test this behavior?


```go
// NewReader returns a new [Reader] whose buffer has the default size.
func NewReader(rd io.Reader) *Reader {
	return NewReaderSize(rd, defaultBufSize)
}
```

cmd/kwil-admin/cmds/snapshot/create.go
```diff
@@ -191,32 +193,37 @@ func pgDump(ctx context.Context, dbName, dbUser, dbPass, dbHost, dbPort string,
...
-	for scanner.Scan() {
-		line := scanner.Text()
 		trimLine := strings.TrimSpace(line)
 
 		// Remove whitespaces, set and select statements, process voters table
 		if inVotersBlock {
 			// Example voter: \\xdae5e91f74b95a9db05fc0f1f8c07f95	\\x9e52ff636caf4988e72e4ac865e6ef83a1e262d1a6376a300f3db8884e1f2253	1
 
-			if line == "\\." { // End of voters block
+			if trimLine == "\\." { // End of voters block
```
jchappelow, line 214: Backticks can make this read easier so we don't have to escape the backslash.

cmd/kwild/server/build.go
```diff
@@ -359,6 +363,20 @@ func (c *closeFuncs) closeAll() error {
...
 
+func verifyDependencies(d *coreDependencies) {
+	// Check if pg_dump is installed, which is necessary for snapshotting during migrations and when snapshots are enabled
+	if _, err := exec.LookPath("pg_dump"); err != nil {
+		failBuild(err, "pg_dump is not installed. Please install it to use snapshots or migrations.")
+	}
+
+	if d.cfg.ChainConfig.StateSync.Enable {
+		// Check if psql is installed, which is required for state-sync
+		if _, err := exec.LookPath("psql"); err != nil {
```
jchappelow, line 374: I'm curious what the output of the following is on the mac versions:

```
$  psql -V
psql (PostgreSQL) 16.4 (Ubuntu 16.4-1.pgdg22.04+1)

$  pg_dump -V
pg_dump (PostgreSQL) 16.4 (Ubuntu 16.4-1.pgdg22.04+1)
```

cmd/kwil-admin/cmds/snapshot/create.go
jchappelow: Well, nevermind!  It works without raising the buffer size:

```go
func TestXxx(t *testing.T) {
	var b [1000]byte
	b[0] = 2
	b[800] = 1
	b[801] = 2
	b[999] = 1
	r := bytes.NewReader(b[:])

	br := bufio.NewReaderSize(r, 500)
	line, err := br.ReadBytes(1)
	if err != nil {
		t.Fatal(err)
	}
	t.Logf("%x (%d)", line, len(line))
}
```

cmd/kwild/server/build.go
brennanjl:
```
$ psql -V
psql (PostgreSQL) 16.4
```

cmd/kwil-admin/cmds/setup/init.go
```diff
@@ -105,6 +106,25 @@ func initCmd() *cobra.Command {
...
 				}
 			}
+
+			// saves genesis state snapshot file in the root directory under the name "genesis_state.sql.gz"
+			if genesisState != "" {
+				if genesisState, err = common.ExpandPath(genesisState); err != nil {
+					return display.PrintErr(cmd, err)
+				}
+
+				file, err := os.ReadFile(genesisState)
```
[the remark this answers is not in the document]
charithabandi: Oh Dang, totally forgot about the size of these.

cmd/kwil-admin/cmds/snapshot/create.go
charithabandi: It internally adjusts the buffer sizes.

cmd/kwild/server/build.go
charithabandi: Added version checks to ensure that everyone is running 16.x.

cmd/kwild/server/build.go
```diff
@@ -790,6 +805,18 @@ func buildStatesyncer(d *coreDependencies, db sql.ReadTxMaker) *statesync.StateS
 		d.cfg.ChainConfig.StateSync.RPCServers += "," + providers[0]
 	}
 
+	// create state syncer
+	return statesync.NewStateSyncer(d.ctx, dbCfg, d.cfg.ChainConfig.StateSync.SnapshotDir,
+		providers, db, *d.log.Named("state-syncer"))
+}
+
+func retrieveLightClientTrustOptions(d *coreDependencies) (height int64, hash string, err error) {
```
brennanjl, line 813: Add a comment here just to explain what this does

cmd/kwild/server/build.go
```diff
@@ -358,6 +361,20 @@ func (c *closeFuncs) closeAll() error {
 	return err
 }
 
+func verifyDependencies(d *coreDependencies) {
```
brennanjl, line 364: Add a short comment here to explain what this function does

brennanjl (commented):
Two more requests asking for commets on functions. I find too often I look at code in our repo and there isn't any comments explaining what it does, so we should get in the habit of having some documentation for any function that isn't painfully obviously self explanatory.

jchappelow (dismissed):

brennanjl (approved):

PR #986, opened 2024-09-17
minor documentation updates, config name changes

brennanjl:
While going through the docs, some commands or documentation that I don't like or are incorrect I am making minor changes to. Still WIP.

# Notable Changes

## New flags

I've added a few new flags for configs that were set-able via `config.toml`, but not with flags.
- `--chain.p2p.dial-timeout` flag
- `--chain.p2p.handshake-timeout` flag

I've also fixed the `--log.consensus-level` flag, which previously used an underscore.

## Config Rename

I renamed the `app.rpc-req-limit` config because I felt that it was terribly confusing. The name makes me think that there is some configurable limit to the requests that the RPC can process, but in reality, it is a max request size config. I renamed it to `app.max-req-size`, and have deprecated the old one.

### DIFF
--- a/common/config/config.go
+++ b/common/config/config.go
@@ -61,7 +61,7 @@ type AppConfig struct {
 	DBName string `mapstructure:"pg_db_name"`
 
 	RPCTimeout         Duration                     `mapstructure:"rpc_timeout"`
-	RPCMaxReqSize      int                          `mapstructure:"rpc_req_limit"`
+	RPCMaxReqSize      int                          `mapstructure:"max_req_size"`
 	PrivateRPC         bool                         `mapstructure:"private_rpc"`
 	ChallengeExpiry    Duration                     `mapstructure:"challenge_expiry"`
 	ChallengeRateLimit float64                      `mapstructure:"challenge_rate_limit"`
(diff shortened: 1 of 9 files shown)


common/config/config.go
```diff
@@ -61,7 +61,7 @@ type AppConfig struct {
 	DBName string `mapstructure:"pg_db_name"`
 
 	RPCTimeout         Duration                     `mapstructure:"rpc_timeout"`
-	RPCMaxReqSize      int                          `mapstructure:"rpc_req_limit"`
```
jchappelow, line 64: There are still a number of occurrences of "rpc_req_limit", like in `/home/jon/kwil/git/kwil-db/cmd/kwil-admin/nodecfg/toml.go` and some actual toml files.

Other thing is that `rpc_req_limit` was in v0.8, so any config files with this will have the setting ignored in 0.9.  That doesn't bother me too much, but ideally we'd keep recognizing it and have similar deprecation handling where we first process what we read from viper or whatever.

brennanjl: Ah that's a good point. Ok, I will do that.

jchappelow (approved):

PR #991, opened 2024-09-18
rpc,abci: use abci Info Data field for app info

jchappelow:
This adds a `GasEnabled` field to the `core/types.ChainInfo` response and to `core/types/admin.AppInfo`.

This info is provided through the cometbft RPC client through the ABCI `Info` method in it's opaque `Data string` field.


jchappelow:
I think we can live without this so closing for now, unless anyone else would either (a) like gasEnabled in the chain_info response or (b) the ability to encode kv data in the ABCI Info response.

PR #996, opened 2024-09-18
Remove callauth flag from kwil-cli call subcommand

charithabandi:


cmd/kwil-cli/cmds/common/roundtripper.go
```diff
@@ -54,18 +55,23 @@ func DialClient(ctx context.Context, cmd *cobra.Command, flags uint8, fn RoundTr
...
 
 	// if not using the gateway, then we can simply create a regular client and return
 	if flags&UsingGateway == 0 {
-		clientConfig.AuthenticateCalls = (flags&AuthenticatedCalls != 0)
 		client, err := client.NewClient(ctx, conf.Provider, &clientConfig)
 		if err != nil {
 			return err
 		}
 
+		if client.PrivateMode() && flags&AuthenticatedCalls != 0 && conf.PrivateKey == nil {
```
brennanjl, line 71: This isn't going to work. You do:

```go
if flags&UsingGateway == 0 {
// ...
}
```

but the only way the CLI knows it is talking to the gateway is if the `--authenticate` flag was passed. Therefore, you would never be in this condition if the user was trying to authenticate.

I think the larger issue is that the `UsingGateway` flag is no longer applicable / accurate to what the application is doing, because it will always be true if a user passes `--authenticate`. I really think this code should be refactored, because at this point what some of these variables means is entirely innacurate.

brennanjl: This might have to turn this PR into a larger fix of consolidating gateway client and regular client. This isn't ideal simply because it is a lot of work, but it might be unavoidable.

brennanjl (approved):
lgtm but will wait for Jon to review

cmd/kwil-cli/cmds/common/roundtripper.go
```diff
@@ -54,18 +55,23 @@ func DialClient(ctx context.Context, cmd *cobra.Command, flags uint8, fn RoundTr
 		if needPrivateKey { // only check chain ID if signing something
 			clientConfig.ChainID = conf.ChainID
 		}
-	} else if flags&WithoutPrivateKey == 0 {
+	} else if flags&WithoutPrivateKey == 0 && flags&AuthenticatedCalls == 0 {
```
jchappelow, line 58: Bitmask logic plus double negatives is dizzing.  Let's do above:

```go
needPrivateKey := flags&WithoutPrivateKey == 0
authCalls := flags&AuthenticatedCalls != 0
```

and then write all the conditions in this function in terms of these local variables.

I'm pretty confused about this current condition.

cmd/kwil-cli/cmds/database/call.go
```diff
@@ -44,25 +43,20 @@ kwil-cli database call get_user --dbid 0x9228624C3185FCBcf24c1c9dB76D8Bef5f5DAd6
...
-			// We only need a private key if using gateway auth or signing the
-			// call message.
-			dialFlags := common.WithoutPrivateKey
+			// AuthenticatedCalls specifies that the call should be authenticated with the private key
+			// if the call is made to the Kwild node with private mode enabled. Else, no authentication is required.
+			dialFlags := common.AuthenticatedCalls
 			if gwAuth {
-				// overwrite the WithoutPrivateKey flag, and add the UsingGateway flag
+				// If the call is made to a gateway, the call should be authenticated with the private key.
 				dialFlags = common.UsingGateway
```
jchappelow, line 59: I think we should rename `AuthenticatedCalls`.  Now it indicates that the following context: (a) using `DialClient` for a call, AND (b) NOT using gateway.  So, it's like... `NonGatewayCall`?  :laughing:

core/types/types.go
```diff
@@ -129,3 +129,40 @@ type MigrationMetadata struct {
...
+// service is in private mode (where calls are authenticated, query is disabled,
+// and raw transactions cannot be retrieved).
+type ServiceMode string
+
+const (
+	ModeOpen    ServiceMode = "open"
+	ModePrivate ServiceMode = "private"
+)
+
+// HealthResponse is the response for MethodHealth. This determines the
```
jchappelow, line 143: in docs: `HealthResponse` -> `Health`

jchappelow (changes requested):
Generally LGTM, just a few small nits.

jchappelow (approved):

PR #1004, opened 2024-09-19
statesync should error before applying a snapshot if the DB is alread…

charithabandi:


cmd/kwild/server/build.go
```diff
@@ -848,9 +848,15 @@ func buildStatesyncer(d *coreDependencies) *statesync.StateSyncer {
 		failBuild(nil, "failed to configure state syncer, failed to fetch trust options from the remote server.")
 	}
 
+	db, err := d.poolOpener(d.ctx, d.cfg.AppConfig.DBName, 3)
+	if err != nil {
+		failBuild(err, "failed to build db for state syncer")
+	}
+	closer.addCloser(db.Close, "closing state syncer db")
```
jchappelow, line 855: I think we should just close it here and not wait for closers to do it after kwild shutdown.

jchappelow: Oh, I thought it only used the db in the constructor.... dang, nvm

cmd/kwild/server/build.go
```diff
@@ -763,7 +763,7 @@ func buildSnapshotter(d *coreDependencies) *statesync.SnapshotStore {
 	return ss
 }
 
-func buildStatesyncer(d *coreDependencies) *statesync.StateSyncer {
+func buildStatesyncer(d *coreDependencies, closer *closeFuncs) *statesync.StateSyncer {
```
jchappelow, line 766: I think this can be `func buildStatesyncer(d *coreDependencies, db sql.ReadTxMaker)` and take the `db` from `buildServer`.

I recall that `buildEventStore` need its own instance because it perform writes to the events table outside of the consensus txn, but statesyncer just needs a read tx, so it's fine.

cmd/kwild/server/build.go
```diff
@@ -172,7 +172,7 @@ func buildServer(d *coreDependencies, closers *closeFuncs) *Server {
 
 	// these are dummies, but they might need init in the future.
```
jchappelow, line 173: can rm this comment

jchappelow (changes requested):

jchappelow (approved):

Issue #993, opened 2024-09-18
CI: occasional TestKwildValidatorRemoval failure


jchappelow:
https://github.com/kwilteam/kwil-db/actions/runs/10927860773/job/30335011124#step:22:334

I have seen this a few times recently.

In      `validator_ops.go:85: Executing remove non-validator specification`: `Messages:   	tx should fail - status: , hash a83a878b2ce58ee0d3fe47884642625926b25dc1b050903b6c32841f58a672fa`

PR #1005, opened 2024-09-20
Remove Validator transaction should fail if the node is not a validator

charithabandi:


jchappelow:
Resolves https://github.com/kwilteam/kwil-db/issues/993

brennanjl (approved):

PR #1006, opened 2024-09-20
ci: non-overlapping net for gh actions

jchappelow:
@Yaiba had the idea that perhaps the integration test's network needs to be different, in particular, not overlapping with the subnet used by the github actions environment

### DIFF
--- a/test/go.mod
+++ b/test/go.mod
@@ -12,6 +12,7 @@ replace (
 
 require (
 	github.com/cometbft/cometbft v0.38.12
+	github.com/docker/docker v26.1.5+incompatible
 	github.com/docker/go-connections v0.5.0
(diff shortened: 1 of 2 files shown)


jchappelow:
@Yaiba that might have worked 🤔
Worth trying this for a while?

Yaiba (approved):
LGTM.  let's roll

test/go.mod
```diff
@@ -12,6 +12,7 @@ replace (
 
 require (
 	github.com/cometbft/cometbft v0.38.12
+	github.com/docker/docker v26.1.5+incompatible
```
Yaiba, line 15: lol just noticed this version number

PR #1007, opened 2024-09-20
Make exposed ports unique between admin and json rpc endpoints for te…

charithabandi:


Yaiba (approved):
LGTM.  Turns out mapping two container ports to one host port would cause collision.

test/integration/docker-compose-migration.yml.template
```diff
@@ -6,7 +6,7 @@ services:
     image: {{ .DockerImage }}
     ports:
       - "{{with .ExposedRPCPorts}}{{index . 0}}:{{end}}8484"
-      - "{{with .ExposedRPCPorts}}{{index . 0}}:{{end}}8485"
+      - "1{{with .ExposedRPCPorts}}{{index . 0}}:{{end}}8485"
```
Yaiba, line 9: So both docker 8484 and 8485 is mapped to same host port, this won't start on local machine due to port collision. I don't know why this would work on CI.

jchappelow: In the other compose files , I do stuff like `{{index . 0 | plus 123}}` to tweak this.  But yeah, I think something strange was going on on docker / CI such that it didn't fail just starting the containers like this

charithabandi: Ah! I think CI passes only because we don't use integration.WithExposedRPCPorts() on any test-cases other than in  task dev:testnet:up. In all other testcases, the exposed ports are different.

test/integration/docker-compose-migration.yml.template
```diff
@@ -6,7 +6,7 @@ services:
     image: {{ .DockerImage }}
     ports:
       - "{{with .ExposedRPCPorts}}{{index . 0}}:{{end}}8484"
-      - "{{with .ExposedRPCPorts}}{{index . 0}}:{{end}}8485"
+      - "{{with .ExposedRPCPorts}}{{index . 0 | plus 100}}:{{end}}8485"
```
Yaiba, line 9: haha I don't see how this is different than the previous approach, other tests still need to change the port being used right?

charithabandi: Nah, looks like it's not required, we always request the address based on the unexposed port which is always 8485.

Yaiba: Oh my bad, I thought the CI failed the other day is due to admin test fail because of the port changes. Just checked the output, it's the 'pool overlap' issue - -

Yaiba (approved):
LGTM.  
@jchappelow is this good to you? I think as long as no admin test is relying on this port we are good

jchappelow (approved):

Issue #999, opened 2024-09-19
`app.tls_cert_file` and `app.tls_key_file` are incorrectly documented


brennanjl:
The `app.tls_cert_file` and `app.tls_key_file` configs are not correctly documented, and should be renamed.

Previously, these configs allowed users to configure TLS for all of the RPCs provided by Kwil. However, when we moved the user service to JSON RPC, we removed the ability to configure TLS, since most users should just use nginx anyways. Now, the TLS configs only apply to the admin service and to comet BFT's RPC (which I am not sure why we ever had it do that in the first place).

Therefore, there are 3 changes we need to make:
- [ ] rename the configs to be `app.admin_tls_cert_file` and `app.admin_tls_key_file`
- [ ] remove the ability for these to be used to configure CometBFT's RPC
- [ ] update the [documentation](<https://github.com/kwilteam/docs>) to make it clear these are only for admin

jchappelow:
Doing this.  Assuming you don't have this already done @brennanjl ?

brennanjl:
I have not done this yet. Thanks.

PR #1013, opened 2024-09-23
config,cmds: prefix admin- to tls-{cert,key}-file settings

jchappelow:
Resolves https://github.com/kwilteam/kwil-db/issues/999 (except for docs repo, which I'll address too)

### DIFF
--- a/cmd/kwil-admin/nodecfg/toml.go
+++ b/cmd/kwil-admin/nodecfg/toml.go
@@ -115,7 +115,9 @@ private_key_path = "{{ .AppConfig.PrivateKeyPath }}"
 # TCP address for the KWILD App's JSON-RPC server to listen on
 jsonrpc_listen_addr = "{{ .AppConfig.JSONRPCListenAddress }}"
 
-# Unix socket or TCP address for the KWILD App's Admin GRPC server to listen on
+# Unix socket or TCP address for the KWILD App's Admin GRPC server to listen on.
+# For a non-loopback TCP address, TLS is automatically enabled unless
+# admin_notls is set to true.
 admin_listen_addr = "{{ .AppConfig.AdminListenAddress }}"
 
 # Timeout on requests on the user RPC servers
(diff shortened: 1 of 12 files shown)


jchappelow:
I just recalled that I need to put some deprecation code for these settings for config.toml how was done for the RPC request limit setting recently.

jchappelow:
> I just recalled that I need to put some deprecation code for these settings for config.toml how was done for the RPC request limit setting recently.

This is done now.

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -115,7 +115,9 @@ private_key_path = "{{ .AppConfig.PrivateKeyPath }}"
 # TCP address for the KWILD App's JSON-RPC server to listen on
 jsonrpc_listen_addr = "{{ .AppConfig.JSONRPCListenAddress }}"
 
-# Unix socket or TCP address for the KWILD App's Admin GRPC server to listen on
+# Unix socket or TCP address for the KWILD App's Admin GRPC server to listen on.
```
brennanjl, line 118: Admin service does not use GRPC anymore, so it might be worth updating this comment

cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -166,20 +168,17 @@ pg_db_name = "{{ .AppConfig.DBName }}"
 # any TCP listen address.
 admin_notls = {{ .AppConfig.NoTLS }}
```
brennanjl, line 169: Isn't this totally unnecessary now since tls only gets configured for admin?

At what point would anyone set this to true?

jchappelow: So this can change, but right now TLS automatically gets enabled if the listen address is not either unix or a loopback tcp.  The cert/key being empty in the config does not disable it (as it does with the user service) as the defaults to use in the aforementioned case are just rpc.cert/rpc.key in the root directory.
Thus, @charithabandi uses `admin_notls` to bypass this so test containers' admin service can be accessed via non-loopback TCP.

I'll take another a look but I think we want to enable TLS if it's listening on an interface that looks like it's going to be accessed from a remote machine. Two types of authentication can be used: password with `admin_pass`, which dangerous if the transport is not encrypted, or mTLS.

Will look again.

brennanjl (approved):

PR #1014, opened 2024-09-23
test: fail fast in expectTxSuccess and expectTxFail

jchappelow:
This is part of investigating the common "transaction not found" responses that we often receive briefly after broadcast (we should only ever get "not confirmed" or an execution result).

This updates the `expectTxSuccess` and `expectTxFail` helper functions in the test module to use `require.EventuallyWithT` that in theory allows a fail-fast mode.  This would be very helpful when:

- in `expectTxFail` we find the transaction mined and executed with code 0 (success) since there's no longer a need to keep checking
- in `expectTxSuccess` when the error is something other than "not confirmed" (or for now, "not found")

However, there's a major issue with `EventuallyWithT` that prevents it from returning cleanly. The fix needs to be released: https://github.com/stretchr/testify/issues/1396

We can easily work around this with our own "eventually" helper function, and I may do that.


jchappelow:
Taking out of draft just to run it through CI

jchappelow:
This PR was 90% for debugging.  Plus it requires an unreleased testify and some other cruft just to support fail-fast and making us aware of the "transaction not found" issue that will shortly be a moot point.  Closing.

Issue #995, opened 2024-09-18
feat: Metrics


brennanjl:
A few times now, we have had users request the ability for metrics. We added a basic sort of metrics to `kgw` (this was to allow Truflation to perform analytics on their network and customer usage), but we also should support metrics on `kwild` such that node operators can ensure their nodes are running properly. This was suggested in the TSN node operator discord today by @mo-husseini-vc from [Validation Cloud](<https://www.validationcloud.io/>), who mentioned that it would be very helpful to have a metrics endpoint that can be integrated with Prometheus.

CometBFT has an [endpoint for this](<https://docs.cometbft.com/v0.38/core/metrics>), which definitely contains some information that would be relevant to us. Some metrics here that I think would be particularly helpful to us:
- consensus_validator_missed_blocks
- consensus_missing_validators
- consensus_missing_validators_power
- consensus_byzantine_validators_power
- consensus_block_interval_seconds
- consensus_rounds
- consensus_num_txs
- consensus_block_size_bytes
- mempool_tx_size_bytes
- mempool_recheck_times
- state_block_processing_time
- state_block_processing_time

Additionally, Mo had some suggestions on what else we could track:

- Postgres round trip time
- RPC request time

A few others that could obviously be helpful:

- Simple RPC req/s
- Events written per second (within the event store)

I don't think we can (or should) add this into v0.9, however we should backlog it for v0.10. Additionally, we should use the TSN testnet as an opportunity to get feedback from validators to see what they expect from this endpoint / what would be helpful for them.

brennanjl:
As discussed in Slack with @outerlook, we have a need to expose the CometBFT metrics as a part of v0.9. This won't require anything more than adding some configurations to Kwil, however we should be careful how we do it, such that we can avoid making breaking changes when we add more Kwil-specific instrumentation.

Below, I have the (potential) new configurations that we can add to Kwil to enable this. With these, we can enable usage of the current CometBFT metrics endpoint, and later add new metrics for Kwil. These generally match [CometBFT's config](<https://github.com/cometbft/cometbft/blob/1ecadfc52659ae8d7eaaa315294e26b8f6d1c52c/config/config.go#L1456>), however I removed the ability to configure namespace, since we will likely want to enforce separate namespaces for comet and Kwil

```toml
[instrumentation]
# when true, prometheus metrics are served under /metrics
prometheus = true

# listen address for prometheus metrics
prometheus_listen_addr = "tcp://0.0.0.0:26660"

[...]

PR #1015, opened 2024-09-23
add prometheus metrics server and config

jchappelow:
I want to play with prometheus a bit before saying this is right, but it addresses exposing cometbft's metrics as in https://github.com/kwilteam/kwil-db/issues/995#issuecomment-2369289713

Prometheus server + Grafana can talk to this and see cometbft's metrics, so I think this is properly exposing CometBFT's metrics endpoint.

Metrics: https://docs.cometbft.com/v0.38/core/metrics

image

prometheus config

```yml
global:
  scrape_interval: 15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
  evaluation_interval: 15s # Evaluate rules every 15 seconds. The default is every 1 minute.
  # scrape_timeout is set to the global default (10s).

# Alertmanager configuration
alerting:
  alertmanagers:
    - static_configs:
        - targets:
          # - alertmanager:9093

# Load rules once and periodically evaluate them according to the global 'evaluation_interval'.
rule_files:
  # - "first_rules.yml"
  # - "second_rules.yml"

# A scrape configuration containing exactly one endpoint to scrape:
# Here it's Prometheus itself.
scrape_configs:
  # The job name is added as a label `job=<job_name>` to any timeseries scraped from this config.
  - job_name: "prometheus"

    # metrics_path defaults to '/metrics'
    # scheme defaults to 'http'.

    scrape_interval: 5s

    static_configs:
      - targets: ['localhost:26660']
```

### DIFF
--- a/internal/services/jsonrpc/server.go
+++ b/internal/services/jsonrpc/server.go
@@ -76,6 +76,7 @@ type serverConfig struct {
 	specInfo   *openrpc.Info
 	reqSzLimit int
 	proxyCount int
+	namespace  string
 }
 
 type Opt func(*serverConfig)
@@ -100,6 +101,13 @@ func WithTLS(cfg *tls.Config) Opt {
 func WithTrustedProxyCount(trustedProxyCount int) Opt {
 	return func(c *serverConfig) {
 		c.proxyCount = trustedProxyCount
+	}
+}
+
+// WithMetricsNamespace enables metrics with the provided namespace.
+func WithMetricsNamespace(namespace string) Opt {
+	return func(c *serverConfig) {
+		c.namespace = namespace
 	}
 }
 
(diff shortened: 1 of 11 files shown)


internal/services/jsonrpc/server.go
```diff
@@ -56,6 +58,14 @@ type Server struct {
...
 	authSHA        []byte
 	tlsCfg         *tls.Config
+
+	// UNSTABLE: this is not much more than a placeholder to ensure we can add
+	// our own metrics to the global prometheus metrics registry.
+	metrics map[string]Metrics
+}
+
+type Metrics interface {
+	Inc()
```
jchappelow, line 68: Here's some PoC stuff.  I marked this counter-type metric as unstable as we should fully specify all of kwild's metrics and organize them better within a hierarchy of namespaces.

brennanjl (approved):
LGTM, thanks a ton for doing this

PR #1017, opened 2024-09-24
abci: detect inconsistent state on startup

jchappelow:
WIP: more will probably come of this, perhaps reworking chain state with a staggered design or another row to contain the "previous" or pending app hash.  Needs design.


jchappelow:
More complete fixes incoming from @charithabandi

PR #1018, opened 2024-09-24
persist app state to disk for clean restarts when the node end up wit…

charithabandi:
Reproduce:
```go
func (a *AbciApp) Commit(ctx context.Context, _ *abciTypes.RequestCommit) (*abciTypes.ResponseCommit, error) {
	err := a.consensusTx.Commit(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to commit transaction app: %w", err)
	}

	panic("Panic here to put the node in inconsistent state")
}
```

CometBFT persists the FinalizeBlock results, so if a crash occurs in the Commit method, CometBFT doesn't replay FinalizeBlock during restart and directly applies the Commit() method. This breaks the chain if the failure occurs as pointed in the above code, where the database state is committed but not the app hash (currently its []byte{42}). Therefore the replay commits block with incorrect app state and proceeds to the next block, which makes the chain non replay-able. 

This PR fixes this by storing the app state at the end of Finalize Block (cometBFT expects app to persist this info) (This would be way simpler if cometBFT can give us the appHash with the commit method :/)

### DIFF
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -59,6 +59,8 @@ type AbciConfig struct {
 	GenesisAllocs      map[string]*big.Int
 	GasEnabled         bool
 	ForkHeights        map[string]*uint64
+
+	ABCIDir string
 }
 
 type AbciApp struct {
@@ -126,7 +128,7 @@ type AbciApp struct {
 }
 
 func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule, statesyncer StateSyncModule,
-	txRouter TxApp, consensusParams *chain.ConsensusParams, peers WhitelistPeersModule, migrator MigratorModule, db DB, dir string, log log.Logger) (*AbciApp, error) {
+	txRouter TxApp, consensusParams *chain.ConsensusParams, peers WhitelistPeersModule, migrator MigratorModule, db DB, logger log.Logger) (*AbciApp, error) {
 	app := &AbciApp{
 		db:                     db,
 		cfg:                    *cfg,
@@ -137,8 +139,8 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
 		consensusParams:        consensusParams,
 		appHash:                cfg.GenesisAppHash,
 		p2p:                    peers,
-		log:                    log,
-		lastCommitInfoFileName: filepath.Join(dir, lastCommitInfoFile),
+		log:                    logger,
+		lastCommitInfoFileName: filepath.Join(cfg.ABCIDir, lastCommitInfoFile),
 
 		validatorAddressToPubKey: make(map[string][]byte),
 		verifiedTxns:             make(map[chainHash]struct{}),
@@ -179,20 +181,20 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
 		// CometBFT seems to persist the FinalizeBlock responses, so during restart it calls "Commit"
 		// directly without calling FinalizeBlock and assumes that the application has the correct state
 		// persisted, whereas the application has the app hash as "42". This will cause the current block
-		// to be committed with an invalid app hash (sentry node blocksyncing will fail at this point).
+		// to be committed with an invalid app hash (any node doing a blocksync will fail at this point).
 		// To fix this, we will store the last commit info in a file and use that as the application's
 		// app hash whenever we find the chain state to be inconsistent signified by the app hash as "42".
 		lc, err := loadLastCommitInfo(app.lastCommitInfoFileName)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to load last commit info due to error: %v. Please drop PostgresDB and rebuild state from existing block data", err)
 		}
 
 		if height != lc.Height {
-			return nil, fmt.Errorf("height mismatch between chain state and last commit info: %d != %d", height, lc.Height)
+			return nil, fmt.Errorf("height mismatch between chain state and last commit info: %d != %d. Please drop PostgresDB and rebuild state from existing block data", height, lc.Height)
 		}
 
 		appHash = lc.AppHash
-		log.Info("Recovered last commit info", zap.Int64("height", height), zap.String("appHash", hex.EncodeToString(appHash)))
+		logger.Warn("Recovered last commit info", log.Int("height", height), log.String("appHash", hex.EncodeToString(appHash)))
 	}
 
 	app.appHash = appHash
@@ -216,7 +218,7 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
 
 		// Update transaction payloads.
 		for _, newPayload := range fork.TxPayloads {
-			log.Infof("Registering transaction route for payload type %v", fork.Name)
+			logger.Infof("Registering transaction route for payload type %v", fork.Name)
 			if err := txapp.RegisterRouteImpl(newPayload.Type, newPayload.Route); err != nil {
 				return nil, fmt.Errorf("failed to register route for payload %v: %w", newPayload.Type, err)
 			}
@@ -814,7 +816,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 			delete(a.validatorAddressToPubKey, addr)
 			if err = a.p2p.RemovePeer(ctx, addr); err != nil {
 				if !errors.Is(err, cometbft.ErrPeerNotWhitelisted) {
-					a.log.Warn("failed to remove demoted validator from peer list", zap.String("address", addr), zap.Error(err))
+					a.log.Warn("failed to remove demoted validator from peer list", log.String("address", addr), log.Error(err))
 				}
 			}
 		} else {
@@ -822,7 +824,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 			// Add the validator to the peer list
 			if err = a.p2p.AddPeer(ctx, addr); err != nil {
 				if !errors.Is(err, cometbft.ErrPeerAlreadyWhitelisted) {
-					a.log.Warn("failed to whitelist promoted validator", zap.String("address", addr), zap.Error(err))
+					a.log.Warn("failed to whitelist promoted validator", log.String("address", addr), log.Error(err))
 				}
 			}
 		}
@@ -834,12 +836,12 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 	for _, pubKey := range approvedJoins {
 		addr, err := cometbft.PubkeyToAddr(pubKey)
 		if err != nil {
-			a.log.Warn("failed to convert pubkey to address", zap.Error(err))
+			a.log.Warn("failed to convert pubkey to address", log.Error(err))
 			continue
 		}
 		if err = a.p2p.AddPeer(ctx, addr); err != nil {
 			if !errors.Is(err, cometbft.ErrPeerAlreadyWhitelisted) {
-				a.log.Warn("failed to whitelist new validator", zap.String("address", addr), zap.Error(err))
+				a.log.Warn("failed to whitelist new validator", log.String("address", addr), log.Error(err))
 			}
 		}
 	}
@@ -852,7 +854,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		}
 		if err = a.p2p.RemovePeer(ctx, addr); err != nil {
 			if !errors.Is(err, cometbft.ErrPeerNotWhitelisted) {
-				a.log.Warn("failed to remove expired validator from peer list", zap.String("address", addr), zap.Error(err))
+				a.log.Warn("failed to remove expired validator from peer list", log.String("address", addr), log.Error(err))
 			}
 		}
 	}
@@ -874,7 +876,7 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		AppHash: newAppHash[:],
 	}
 	if err = lc.saveAs(a.lastCommitInfoFileName); err != nil {
-		return nil, fmt.Errorf("failed to save last commit info: %w", err)
+		a.log.Warn("failed to save last commit info", log.Error(err))
 	}
 
 	return res, nil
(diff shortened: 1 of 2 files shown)


jchappelow (commented):
Crash issues to check are addressed/addressable:

- [ ] in `FinalizeBlock`, *after* Precommit, *before* save last commit info
- [ ] in `FinalizeBlock`, *after* Precommit, *after* save last commit info, but *before* returned to (or recorded by) CometBFT
- [ ] between `FinalizeBlock` and `Commit` but after cometbft recorded our finalize response
- [ ] in `app.Commit`, *after* `consensus.Commit`, *before* `SetChainState`
- [ ] in `app.Commit`, *after* `consensus.Commit`, *after* `SetChainState`, *before* returning response to CometBFT

charithabandi:
> Crash issues to check are addressed/addressable:
> 
> * [x]  in `FinalizeBlock`, _after_ Precommit, _before_ save last commit info
> * [x]  in `FinalizeBlock`, _after_ Precommit, _after_ save last commit info, but _before_ returned to (or recorded by) CometBFT
> * [x]  _after_ `FinalizeBlock` and _before_ `Commit`, but after cometbft recorded our finalize response
> * [x]  in `app.Commit`, _after_ `consensus.Commit`, _before_ `SetChainState`
> * [x]  in `app.Commit`, _after_ `consensus.Commit`, _after_ `SetChainState`, _before_ returning response to CometBFT

Tested out all these scenarios and with a sentry node doing block sync for block validation.

jchappelow:
> Crash issues to check are addressed/addressable:
> 
>     * [x]  in `FinalizeBlock`, _after_ Precommit, _before_ save last commit info

In above, the orphaned tx is rolled back.  last commit info is fine

>     * [x]  in `FinalizeBlock`, _after_ Precommit, _after_ save last commit info, but _before_ returned to (or recorded by) CometBFT

In above, the orphaned tx is rolled back.  last commit info is wrong, but... ignored?
 
>     * [x]  _after_ `FinalizeBlock` and _before_ `Commit`, but after cometbft recorded our finalize response

In above, what?  On restart, Cometbft skips `FinalizeBlock`, straight to `Commit`.  How does the prepared tx get committed?

>     * [x]  in `app.Commit`, _after_ `consensus.Commit`, _before_ `SetChainState`

In above, ?

>     * [x]  in `app.Commit`, _after_ `consensus.Commit`, _after_ `SetChainState`, _before_ returning response to CometBFT

In above, ?

internal/abci/abci.go
```diff
@@ -820,13 +848,17 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
...
+				a.log.Warn("failed to remove expired validator from peer list", zap.String("address", addr), zap.Error(err))
 			}
 		}
 	}
 
 	// wait for all logs to be received
-	wg.Wait()
+	select {
+	case <-doneLogs:
+	case <-ctx.Done():
```
jchappelow, line 859: Well crud, this isn't working as I expected.  If I remove the `case <-doneLogs`, and shutdown PostgreSQL, it does stop everything, but comet doesn't cancel the context it passed:

```
2024-09-24T17:52:39.35-05:00	info	kwild.server	stop listeners
2024-09-24T17:52:39.35-05:00	info	kwild.server	stop comet server
2024-09-24T17:52:39.35-05:00	info	kwild.admin-jsonrpc-server	JSON-RPC server shutting down...
2024-09-24T17:52:39.35-05:00	info	kwild.listener-manager	ListenerManager stopped.
2024-09-24T17:52:39.35-05:00	info	kwild.user-jsonrpc-server	JSON-RPC server shutting down...
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"msg": "Stopping Node service", "impl": "Node"}
2024-09-24T17:52:39.35-05:00	info	kwild.user-jsonrpc-server	JSON-RPC listener done for [::]:8484
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	Stopping Node
2024-09-24T17:52:39.35-05:00	info	kwild.user-jsonrpc-server	JSON-RPC server shutdown complete
2024-09-24T17:52:39.35-05:00	info	kwild.admin-jsonrpc-server	JSON-RPC listener done for /tmp/kwild.socket
2024-09-24T17:52:39.35-05:00	info	kwild.admin-jsonrpc-server	JSON-RPC server shutdown complete
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "events", "msg": "Stopping EventBus service", "impl": "EventBus"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "events", "module": "pubsub", "msg": "Stopping PubSub service", "impl": "PubSub"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "txindex", "msg": "Stopping IndexerService service", "impl": "IndexerService"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "p2p", "msg": "Stopping P2P Switch service", "impl": "P2P Switch"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "pex", "msg": "Stopping PEX service", "impl": "PEX"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "p2p", "book": "/home/jon/.kwild-junk3/abci/addrbook.json", "msg": "Stopping AddrBook service", "impl": "AddrBook"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "mempool", "msg": "Stopping Mempool service", "impl": "Mempool"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	Saving AddrBook to file	{"module": "p2p", "book": "/home/jon/.kwild-junk3/abci/addrbook.json", "size": 0}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "blocksync", "msg": "Stopping Reactor service", "impl": "Reactor"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "consensus", "msg": "Stopping Consensus service", "impl": "ConsensusReactor"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "consensus", "msg": "Stopping State service", "impl": "ConsensusState"}
2024-09-24T17:52:39.35-05:00	info	kwild.cometbft	service stop	{"module": "consensus", "msg": "Stopping TimeoutTicker service", "impl": "TimeoutTicker"}
```

Rethinking this...

internal/abci/abci.go
```diff
@@ -783,15 +811,15 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
...
+					a.log.Warn("failed to remove demoted validator from peer list", zap.String("address", addr), zap.Error(err))
 				}
 			}
 		} else {
 			a.validatorAddressToPubKey[addr] = up.PubKey // there may be new validators we need to add
 			// Add the validator to the peer list
 			if err = a.p2p.AddPeer(ctx, addr); err != nil {
 				if !errors.Is(err, cometbft.ErrPeerAlreadyWhitelisted) {
-					return nil, fmt.Errorf("failed to whitelist promoted validator %s: %w", addr, err)
+					a.log.Warn("failed to whitelist promoted validator", zap.String("address", addr), zap.Error(err))
```
jchappelow, line 822: There are two more above for approved and expired joins

charithabandi:
>     * [x]  _after_ `FinalizeBlock` and _before_ `Commit`, but after cometbft recorded our finalize response
This can lead to orphaned tx or not depending on the nature of the crash.  If its an orphaned tx, it is rollbacked in the start. But in both the cases, our chain state has the last block's state, so we start replay from the previous block instead. 
 
>     * [x]  in `app.Commit`, _after_ `consensus.Commit`, _before_ `SetChainState`

In above, ?
It's like previous, except there's no orphaned tx.  There's also no consensusTx.
The added 42 check in constructor does detect this though.

::edit
correct, this is the only problematic scenario which causes invalid blocks. This is where the state is read from the last_commit_info instead as cometBFT doesn't do FinalizeBlock again.

internal/abci/abci.go
charithabandi: OKay, I am gonna remove these changes from this PR.

jchappelow:
> > ```
> > * [x]  _after_ `FinalizeBlock` and _before_ `Commit`, but after cometbft recorded our finalize response
> > ```
> 
> 
> This can lead to orphaned tx or not depending on the nature of the crash.  If its an orphaned tx, it is rollbacked in the start. But in both the cases, our chain state has the last block's state, so we start replay from the previous block instead.

After FinalizeBlock but before Commit is an orphaned prepared transaction.  

The problem I'm referring to is where we gave a FinalizeBlock response, so cometbft skips to Commit on restart.  How can are we handling that?

internal/abci/abci.go
```diff
@@ -164,6 +172,29 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
...
+		// state but before the commit of the app hash. The chain state will have the app hash as "42".
+		// CometBFT seems to persist the FinalizeBlock responses, so during restart it calls "Commit"
+		// directly without calling FinalizeBlock and assumes that the application has the correct state
+		// persisted, whereas the application has the app hash as "42". This will cause the current block
+		// to be committed with an invalid app hash (sentry node blocksyncing will fail at this point).
+		// To fix this, we will store the last commit info in a file and use that as the application's
+		// app hash whenever we find the chain state to be inconsistent signified by the app hash as "42".
+		lc, err := loadLastCommitInfo(app.lastCommitInfoFileName)
+		if err != nil {
+			return nil, err
```
jchappelow, line 187: I think we should decorate this error as it might be some cryptic complaint from the json decoder about an invalid utf8 character.  Provide context and even resolution (drop pg DB and rebuild state from existing block data)

internal/abci/abci.go
```diff
@@ -164,6 +172,29 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
...
 		height = 0 // negative means first start (no state table yet), but need non-negative for below logic
 	}
+
+	if bytes.Equal(appHash, []byte{0x42}) {
+		// This is an interesting corner case where the node crashes after the commit of block execution
+		// state but before the commit of the app hash. The chain state will have the app hash as "42".
+		// CometBFT seems to persist the FinalizeBlock responses, so during restart it calls "Commit"
+		// directly without calling FinalizeBlock and assumes that the application has the correct state
+		// persisted, whereas the application has the app hash as "42". This will cause the current block
+		// to be committed with an invalid app hash (sentry node blocksyncing will fail at this point).
```
jchappelow, line 182: The sentry node comment is confusing.  Any node block sync will fail, including this one if it tries to resync.  You are alluding to the case where if this nodes is the only validator, there will be no other nodes with which it will disagree to cause a consensus failure, resulting in silent corruption that will only be detected on resync (or other node sync)

internal/abci/abci.go
```diff
@@ -164,6 +172,29 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
...
+		// To fix this, we will store the last commit info in a file and use that as the application's
+		// app hash whenever we find the chain state to be inconsistent signified by the app hash as "42".
+		lc, err := loadLastCommitInfo(app.lastCommitInfoFileName)
+		if err != nil {
+			return nil, err
+		}
+
+		if height != lc.Height {
+			return nil, fmt.Errorf("height mismatch between chain state and last commit info: %d != %d", height, lc.Height)
```
jchappelow, line 191: A brief hint at resolution would be good here (drop pg DB and resync)

internal/abci/abci.go
```diff
@@ -164,6 +172,29 @@ func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule
...
+		}
+
+		appHash = lc.AppHash
+		log.Info("Recovered last commit info", zap.Int64("height", height), zap.String("appHash", hex.EncodeToString(appHash)))
```
jchappelow, line 195: This should be rare, and a sign of crash, so Warn would be appropriate and helpful.

Also, in log statements, it would be best to be in the habbit of using our generic helpers like `log.Int` instead of the zap ones.  We want to minimally couple our packages with zap and leave that detail to the our `log` package

internal/abci/abci.go
```diff
@@ -835,6 +868,15 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		result.Result.Log += logs.logs
 	}
 
+	// Persist app hash and height to the disk for recovery purposes.
+	lc := &lastCommitInfo{
+		Height:  req.Height,
+		AppHash: newAppHash[:],
+	}
+	if err = lc.saveAs(a.lastCommitInfoFileName); err != nil {
+		return nil, fmt.Errorf("failed to save last commit info: %w", err)
```
jchappelow, line 877: I'm undecided if this should halt chain.  If nothing else errors, the node is generally OK.  If there is a recovery scenario that would need this file and it's either missing or for the wrong height, it will fail then.
If this ave fails because of a permissions issue (for example), that could be resolved by the operator without even stopping the node.  Thoughts?

internal/abci/abci.go
```diff
@@ -959,6 +1001,38 @@ func (a *AbciApp) Commit(ctx context.Context, _ *abciTypes.RequestCommit) (*abci
...
 
+// lastCommitInfo is a struct to store the last commit info such as
+// height and apphash at the end of the FinalizeBlock method.
+type lastCommitInfo struct {
+	AppHash types.HexBytes `json:"app_hash"`
+	Height  int64          `json:"height"`
+}
+
+// saveAs saves the last commit info to root_dir/abci/last_commit_info.json.
+func (lc *lastCommitInfo) saveAs(filename string) error {
```
jchappelow, line 1012: Just a nit, but these method docs are misleading.  They are not aware of the path the docs mention at all.  They save/load blindly from the *absolute* path provided in `filename`

internal/abci/abci.go
```diff
@@ -116,21 +119,26 @@ type AbciApp struct {
...
 	migrator MigratorModule
+
+	// lastCommitInfoFileName is the file name of the last commit info file
+	// which stores the app hash and height at the end of FinalizeBlock.
+	lastCommitInfoFileName string
 }
 
 func NewAbciApp(ctx context.Context, cfg *AbciConfig, snapshotter SnapshotModule, statesyncer StateSyncModule,
-	txRouter TxApp, consensusParams *chain.ConsensusParams, peers WhitelistPeersModule, migrator MigratorModule, db DB, log log.Logger) (*AbciApp, error) {
+	txRouter TxApp, consensusParams *chain.ConsensusParams, peers WhitelistPeersModule, migrator MigratorModule, db DB, dir string, log log.Logger) (*AbciApp, error) {
```
jchappelow, line 129: I'm not sure think we need to or should add this `dir` param.  We already have a log of parameters, and the purpose and scope of this one is unclear from the function signuature.   Could be in `AbciConfig` or manage some other way?

jchappelow (approved):
LGTM.

PR #1019, opened 2024-09-24
fix: kwil-admin setup reset does not use custom pg config

brennanjl:
Fixes a bug where `kwil-admin setup reset` was not properly reading in Postgres configs from the `config.toml` in the root directory.


charithabandi (approved):
Works !

jchappelow (approved):

PR #1020, opened 2024-09-24
Rename snapshots enabled to enable

charithabandi:


cmd/kwild/config/flags.go
```diff
@@ -119,4 +119,6 @@ to instead run a dedicated seeder like https://github.com/kwilteam/cometseed.`))
 	flagSet.MarkDeprecated("app.snapshots.snapshot-dir", "this value is no longer configurable")
 	flagSet.String("chain.statesync.snapshot-dir", "", "Chain state sync snapshot directory")
 	flagSet.MarkDeprecated("chain.statesync.snapshot-dir", "this value is no longer configurable")
+	flagSet.BoolVar(&cfg.AppConfig.Snapshots.DEPRECATED_Enabled, "app.snapshots.enabled", cfg.AppConfig.Snapshots.DEPRECATED_Enabled, "Enable snapshots")
```
jchappelow, line 122: For config.toml, we need something like this: https://github.com/kwilteam/kwil-db/pull/1013/files#diff-f23562fd52a0d43d17897596ae698df69cbb537eebd01410825af78beaa19759R154

brennanjl, line 122: It looks like you still need to use the DEPRECATED_Enabled config. If a user sets this to true, it should set the `enable` config. If a user sets both, it should error.

brennanjl (approved):

PR #1021, opened 2024-09-25
pg,abci: unblock notice subscribers on DB shutdown

jchappelow:
This resolves a possible hang in `FinalizeBlock` if the DB shuts down / dies while waiting for the completion of the notice stream at `wg.Wait()`.  When the `pg.DB` instance closes up in response to the replication connection terminating, we also make sure to close the subscriber channels to unblock the receivers.  To ensure the receivers handle this as an error rather than assuming all notices have been received successfully, an empty string is first sent on the channel.  Since all notice messages have a special prefix (`pgtx:`), this is easily recognized as an exception.

### DIFF
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -606,21 +606,28 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 	defer done(ctx)
 
 	// wait group to wait at the end of the function for all logs to be received
-	var wg sync.WaitGroup
-	wg.Add(1)
+	logsDone := make(chan error, 1)
 	go func() {
+		defer close(logsDone)
+
 		// we enforce that the cumulative size of logs is less than 1KB
 		// per tx. This is a work-around until we have gas costs to protect
 		// against log spam.
 		for {
 			log, ok := <-logs
-			if !ok {
-				break
+			if !ok { // empty and closed (normal completion)
+				return // logsDone receiver gets nil error
+			}
+			if log == "" {
+				// The DB has shut down with active subscribers. Fail.
+				logsDone <- errors.New("premature notice stream termination")
+				return
 			}
 			txid, notice, err := parse.ParseNotice(log)
 			if err != nil {
-				// should be deterministic so nbd to not halt here
-				a.log.Errorf("failed to parse notice: %v", err)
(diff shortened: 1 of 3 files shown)


internal/abci/abci.go
```diff
@@ -573,21 +573,28 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
...
+				logsDone <- errors.New("premature notice stream termination")
+				return
 			}
 			txid, notice, err := parse.ParseNotice(log)
 			if err != nil {
-				// should be deterministic so nbd to not halt here
-				a.log.Errorf("failed to parse notice: %v", err)
+				// will still be deterministic so nbd to not halt here
+				a.log.Errorf("failed to parse notice (%.20s...): %v", log, err)
+				continue // since txid is invalid and won't match any result.TxHash
```
jchappelow, line 597: Pls double check me on this `continue` @brennanjl .  I tend to agree it not necessary to halt, but I don't think we should attempt to use the `txid` as a map key below, just continue to receive in the next iteration of the loop

brennanjl: Yeah this seems correct.

internal/abci/abci.go
```diff
@@ -825,8 +828,18 @@ func (a *AbciApp) FinalizeBlock(ctx context.Context, req *abciTypes.RequestFinal
 		}
 	}
 
-	// wait for all logs to be received
-	wg.Wait()
+	// wait for all logs to be received, or a premature shutdown
+	select {
+	case <-ctx.Done():
```
brennanjl, line 833: Do we need to worry about this not being supported in comet v0.38? I don't think so because we close the notice subscribers when the database shuts down, but just want to double check that you see it the same

jchappelow: No it's ok.  With 0.38 we only have one way to break out of here, and that's the notice signaling that the DB dies.  With 1.0 we can also get a signal from cometbft, but that's just going to be a bonus.

brennanjl (approved):
lgtm but seems to be a merge conflict

jchappelow:
gofmt'd

jchappelow:
Resolved conflicts.

brennanjl (approved):

PR #1023, opened 2024-09-25
test: randomized non-overlapping networks on 10.A.B.0/20

jchappelow:
Since we cannot set Docker's `default-address-pools` in daemon.json on Github Actions like we did our our old self-hosted running, this PR attempts to pick random /20 networks in 10.A.B.0/20 (there are 4096 subnets each with 4096 addresses) rather than use "10.9.0.0/16", which can cause trouble if the network isn't taken down fast enough or if a test fails and decides not to clean up the network, which can break subsequent tests without different networks.

### DIFF
--- a/test/utils/container.go
+++ b/test/utils/container.go
@@ -3,10 +3,12 @@ package utils
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 
 	dockerNet "github.com/docker/docker/api/types/network"
 	"github.com/docker/go-connections/nat"
+	"github.com/kwilteam/kwil-db/core/utils/random"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/network"
 )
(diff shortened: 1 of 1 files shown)


jchappelow:
closing, but keeping my branch handing in case CI starts hitting `failed to create docker network for TestStatesync/cli_driver: Error response from daemon: Pool overlaps with other one on this address space: failed to create network`, or we can remove the IPAM section entirely to go back to docker managed random networks, but it seems that was clashing with github's services networks somehow.

jchappelow:
Closing but not deleting branch in case we want to resurrect this.

jchappelow:
I think we should at least try this before entirely reverting the ipam section of docker network config.  I'm seeing frequent docker errors in CI related to overlapping networks.  Usually it's just following a failed test that is not cleaned up, but it is misleading when trying to diagnose CI failure.

Yaiba (approved):
LGTM

charithabandi (approved):

PR #1027, opened 2024-09-25
Poller to download the genesis state. Node start to delay until the g…

charithabandi:
Integration tests might fail, need to adjust the config there. But tested locally

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -63,6 +63,9 @@ func defaultBinaryConfig() binaryConfig {
 var DefaultConfig = func() *commonConfig.KwildConfig {
 	return &commonConfig.KwildConfig{
 		RootDir: defaultKwildRoot(),
+		MigrationConfig: &commonConfig.MigrationConfig{
+			Enable: false,
+		},
 		AppConfig: &commonConfig.AppConfig{
 			JSONRPCListenAddress: "0.0.0.0:8484",
 			AdminListenAddress:   "/tmp/kwild.socket", // Or, suggested, 127.0.0.1:8485
--- a/cmd/kwil-admin/nodecfg/toml.go
+++ b/cmd/kwil-admin/nodecfg/toml.go
@@ -108,11 +108,12 @@ time_format = "{{ .Logging.TimeEncoding }}"
 #######################################################################
 
 [migration]
-# Setting this to true will enable the zero-downtime migration process on the node.
+# Enables migration mode
 enable = {{ .MigrationConfig.Enable }}
 
-# The listening address of the node to migrate the app state from.
-# mandatory if the start_height and end_height are provided in the genesis file.
+# The listening address of the trusted node from the old network. This is used to
+# fetch the genesis state and the block changes from the old network during migration.
+# This is a mandatory field if migration is enabled.
 migrate_from = "{{ .MigrationConfig.MigrateFrom }}"
 
 #######################################################################
--- a/cmd/kwil-admin/nodecfg/toml.go
+++ b/cmd/kwil-admin/nodecfg/toml.go
@@ -104,6 +104,18 @@ format = "{{ .Logging.Format }}"
 time_format = "{{ .Logging.TimeEncoding }}"
 
 #######################################################################
+###                    Migration Config Options                     ###
+#######################################################################
+
+[migration]
+# Setting this to true will enable the migration process on the node.
+enable = {{ .MigrationConfig.Enable }}
+
+# The listening address of the node to migrate the app state from.
+# mandatory if the start_height and end_height are provided in the genesis file.
+migrate_from = "{{ .MigrationConfig.MigrateFrom }}"
+
+#######################################################################
 ###                      App Config Options                         ###
 #######################################################################
 
@@ -189,9 +201,6 @@ hostname = "{{ .AppConfig.Hostname }}"
 # Might be either absolute path or path related to the kwild root directory.
 genesis_state = "{{ .AppConfig.GenesisState }}"
 
-# The listening address of the node to migrate the app state from.
-# mandatory if the start_height and end_height are provided in the genesis file.
-migrate_from = "{{ .AppConfig.MigrateFrom }}"
 
 #######################################################################
 ###                     Extension Configuration                     ###
--- a/cmd/kwil-admin/nodecfg/toml.go
+++ b/cmd/kwil-admin/nodecfg/toml.go
@@ -108,7 +108,7 @@ time_format = "{{ .Logging.TimeEncoding }}"
 #######################################################################
 
 [migration]
-# Setting this to true will enable the migration process on the node.
+# Setting this to true will enable the zero-downtime migration process on the node.
 enable = {{ .MigrationConfig.Enable }}
 
 # The listening address of the node to migrate the app state from.
--- a/cmd/kwild/root/root.go
+++ b/cmd/kwild/root/root.go
@@ -89,12 +89,15 @@ func RootCmd() *cobra.Command {
 				return fmt.Errorf("failed to initialize private key and genesis: %w", err)
 			}
 
-			if err := genesisConfig.SanityChecks(); err != nil {
-				return fmt.Errorf("genesis configuration failed sanity checks: %w", err)
+			if kwildCfg.MigrationConfig.Enable {
+				kwildCfg, genesisConfig, err = server.PrepareForMigration(cmd.Context(), kwildCfg, genesisConfig)
+				if err != nil {
+					return fmt.Errorf("failed to prepare for migration: %w", err)
+				}
 			}
 
-			if err := kwildCfg.ConfigureExtensions(genesisConfig); err != nil {
-				return err
+			if err := genesisConfig.SanityChecks(); err != nil {
+				return fmt.Errorf("genesis configuration failed sanity checks: %w", err)
 			}
 
 			// Set the chain package's active forks variable. This provides easy
--- a/cmd/kwild/root/root.go
+++ b/cmd/kwild/root/root.go
@@ -18,6 +18,7 @@ import (
 	"github.com/kwilteam/kwil-db/cmd/kwil-admin/nodecfg"
 	kwildcfg "github.com/kwilteam/kwil-db/cmd/kwild/config"
 	"github.com/kwilteam/kwil-db/cmd/kwild/server"
+	"github.com/kwilteam/kwil-db/core/log"
 
 	// kwild's "root" command package assumes the responsibility of initializing
 	// certain packages, including the extensions and the chain config package.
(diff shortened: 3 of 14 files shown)


cmd/kwil-admin/nodecfg/toml.go
```diff
@@ -103,6 +103,18 @@ format = "{{ .Logging.Format }}"
 # Time format: "epochfloat" (default), "epochmilli", or "rfc3339milli"
 time_format = "{{ .Logging.TimeEncoding }}"
 
+#######################################################################
+###                    Migration Config Options                     ###
+#######################################################################
+
+[migration]
```
charithabandi, line 110: Do we need a whole migration section here ? or just migrate_from as the top level config? any opinions?

brennanjl: This seems fine to me

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,161 @@
...
+// pollForGenesisState polls for the genesis state from the old chain
+// and downloads the genesis state to the snapshot file abnd updates the genesis config
+// and the kwild config with the configuration required for migration
+func (m *migrationClient) pollForGenesisState(ctx context.Context) error {
+	// Poll for the genesis state from the old chain
+	delay := defaultPollFrequency
+	ticker := time.NewTicker(delay)
+	defer ticker.Stop()
+
+	fmt.Println("Polling for genesis state from the old chain")
```
brennanjl, line 78: Use logging instead of fmt.Println

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,161 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
```
brennanjl, line 83: Why ticker instead of time.After? I dont know the benefits of ticker, but I expected to see `time.After` here.

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,161 @@
...
+	// Listener extension
+	extensions := m.kwildCfg.AppConfig.Extensions
+	extensions["migrations"] = map[string]string{
+		"listen_address": m.listenAddress,
+		"start_height":   fmt.Sprintf("%d", metadata.MigrationState.StartHeight),
+		"end_height":     fmt.Sprintf("%d", metadata.MigrationState.EndHeight),
+	}
+	m.kwildCfg.AppConfig.Extensions = extensions
+
+	fmt.Println("Genesis state downloaded successfully", m.snapshotFileName)
```
brennanjl, line 159: use logging here

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,161 @@
...
+	// Update the genesis config
+	m.genesisCfg.DataAppHash = genCfg.DataAppHash
+	m.genesisCfg.Validators = genCfg.Validators
+	m.genesisCfg.ConsensusParams.Migration = genCfg.ConsensusParams.Migration
+
+	// Update the kwild config
+	m.kwildCfg.AppConfig.GenesisState = m.snapshotFileName
+	// Listener extension
+	extensions := m.kwildCfg.AppConfig.Extensions
+	extensions["migrations"] = map[string]string{
```
brennanjl, line 152: Why do we need to configure this here? Two concerns:

1. Isn't the start and end height only used in the genesis.json?
2. Why do we need to have `extensions["migrations"].listen_address` if we already have `cfg.migration.migrate_from`? Seems totally redundant

brennanjl: Using control F, it looks like we literally do not reference these values anywhere.

brennanjl (commented):
Looks right but a few pieces of feedback

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		if n != len(chunk) {
+			return err
+		}
+	}
+
+	// Update the genesis config
+	m.genesisCfg.DataAppHash = genCfg.DataAppHash
+	m.genesisCfg.Validators = genCfg.Validators
+	m.genesisCfg.ConsensusParams.Migration = genCfg.ConsensusParams.Migration
```
brennanjl, line 146: We should probably make sure to save the updated genesis file right? If the node restarts it will then have an invalid genesis.json.

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,161 @@
...
+	return m.kwildCfg, m.genesisCfg, nil
+}
+
+// pollForGenesisState polls for the genesis state from the old chain
+// and downloads the genesis state to the snapshot file abnd updates the genesis config
+// and the kwild config with the configuration required for migration
+func (m *migrationClient) pollForGenesisState(ctx context.Context) error {
+	// Poll for the genesis state from the old chain
+	delay := defaultPollFrequency
+	ticker := time.NewTicker(delay)
```
brennanjl, line 75: nbd but what is the point of copying `defaultPollFrequency` into `delay` just to use it in the next line?

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,183 @@
...
+		if n != len(chunk) {
+			return fmt.Errorf("incomplete write to genesis snapshot chunk. expected: %d, written: %d", len(chunk), n)
+		}
+	}
+
+	// Update the genesis config
+	m.genesisCfg.DataAppHash = genCfg.DataAppHash
+	m.genesisCfg.Validators = genCfg.Validators
+	m.genesisCfg.ConsensusParams.Migration = genCfg.ConsensusParams.Migration
+
```
brennanjl, line 167: We need to set the initial height too

charithabandi: umm, isn't it a new chain?

brennanjl: No, we should use the same height. CometBFT supports this, and users see it as more of a network upgrade rather than a totally new network

jchappelow: Semantics questions aside, we've never tested this.  We probably need to check out abci carefully for assumptions and hard coded 1 /+1.

But what practical good does it do to start at say 100?  Chain is also different

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,183 @@
...
+	logger log.Logger
+}
+
+func PrepareForMigration(ctx context.Context, kwildCfg *commonCfg.KwildConfig, genesisCfg *chain.GenesisConfig, logger log.Logger) (*commonCfg.KwildConfig, *chain.GenesisConfig, error) {
+	if kwildCfg.MigrationConfig.MigrateFrom == "" {
+		return nil, nil, errors.New("migrate_from is mandatory for migration")
+	}
+
+	// old chain client
+	clt, err := client.NewClient(ctx, kwildCfg.MigrationConfig.MigrateFrom, nil)
```
jchappelow, line 53: Why start the client here if we find out just below that we already have genesis state?

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,183 @@
...
+		clt:              clt,
+		kwildCfg:         kwildCfg,
+		genesisCfg:       genesisCfg,
+		snapshotFileName: config.GenesisStateFileName(kwildCfg.RootDir),
+		logger:           logger,
+	}
+
+	// check if the genesis state is already available, ie. snapshot file is already present
+	fileExists := false
+	if _, err := os.Stat(m.snapshotFileName); err == nil {
```
jchappelow, line 70: Should detect errors other that is.IsNotExist

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,183 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to write genesis snapshot chunk: %d  error: %w", i, err)
+		}
+		if n != len(chunk) {
+			return fmt.Errorf("incomplete write to genesis snapshot chunk. expected: %d, written: %d", len(chunk), n)
+		}
+	}
+
+	// Update the genesis config
+	m.genesisCfg.DataAppHash = genCfg.DataAppHash
```
jchappelow, line 164: Don't we need to compute this from the downloaded state file?

cmd/common.go
```diff
@@ -63,6 +63,9 @@ func defaultBinaryConfig() binaryConfig {
 var DefaultConfig = func() *commonConfig.KwildConfig {
 	return &commonConfig.KwildConfig{
 		RootDir: defaultKwildRoot(),
+		MigrationConfig: &commonConfig.MigrationConfig{
+			Enable: false,
```
jchappelow, line 67: Default enable?

cmd/kwild/root/root.go
```diff
@@ -89,12 +90,26 @@ func RootCmd() *cobra.Command {
...
+				return err
 			}
 
-			if err := kwildCfg.ConfigureExtensions(genesisConfig); err != nil {
-				return err
+			logger, err := log.NewChecked(*logCfg)
+			if err != nil {
+				return fmt.Errorf("invalid logger config: %w", err)
+			}
+			logger = *logger.Named("kwild")
```
jchappelow, line 102: We threw out the version log just added

cmd/kwild/server/migration.go
brennanjl: Kwil supports using `@height`. If a network is using `@height` to track the order of when certain things occur, it will be broken during migration. The new network should be treated more of a "fork" of the old network.

cmd/kwild/server/migration.go
charithabandi: the genesis file downloaded already has this hash computed, avoiding recomputation here.

cmd/kwild/server/migration.go
charithabandi: Tested this out (migration, blocksync and statesync), it looks okay with the initial height > 0/1.

cmd/kwild/server/migration.go
```diff
@@ -0,0 +1,212 @@
...
+	snapshotFileName string
+
+	clt        *client.Client
+	kwildCfg   *commonCfg.KwildConfig
+	genesisCfg *chain.GenesisConfig
+
+	logger log.Logger
+}
+
+func PrepareForMigration(ctx context.Context, kwildCfg *commonCfg.KwildConfig, genesisCfg *chain.GenesisConfig, logger log.Logger) (*commonCfg.KwildConfig, *chain.GenesisConfig, error) {
```
brennanjl, line 51: Please adds docs to all structs and functions in this file. Just something small so we know what is going on

brennanjl:
Some feedback while testing:

- Logging that a node is "entering startup mode" would be nice, just to make it clear why it is acting differently
- Might be a bit late for this, but Im sitting here wondering if the migration is actually active. It seems like we need a way to get info on the network current migration while it is in the activation period and while it is active. This should be a `kwil-admin` command. We can add this after release since it is not a breaking change or anything, but I think it would be really nice to get that confirmation

charithabandi:
> Some feedback while testing:
> 
> * Logging that a node is "entering startup mode" would be nice, just to make it clear why it is acting differently
> * Might be a bit late for this, but Im sitting here wondering if the migration is actually active. It seems like we need a way to get info on the network current migration while it is in the activation period and while it is active. This should be a `kwil-admin` command. We can add this after release since it is not a breaking change or anything, but I think it would be really nice to get that confirmation

`kwil-admin node status` shows the migration status. 
https://github.com/kwilteam/kwil-db/blob/6c94395be97d0175c0cf12722ad07ef532dad9f5/core/types/types.go#L103  This captures where in these 4 states the node is in. 
```
const (
	NoActiveMigration MigrationStatus = iota
	MigrationNotStarted
	MigrationInProgress
	MigrationCompleted
)
```

brennanjl (approved):
Tested it and worked as expected

PR #1028, opened 2024-09-25
adds array_agg

brennanjl:
This PR adds the `array_agg` function.

Sorry for the 11th hour addition, I had promised this to someone and totally forgot. Thankfully it is very easy to add this type of stuff.


jchappelow (approved):
LGTM.  I haven't thought through the various combinations of start/distinct/order by, but I guess it's as is should be in PGFormat.  Then don't mix an match?

brennanjl:
> LGTM. I haven't thought through the various combinations of start/distinct/order by, but I guess it's as is should be in PGFormat. Then don't mix an match?

Yeah, users can use distinct, but nothing else. It's not necessarily hard to allow custom ordering, but since this is an 11th hour addition I'd rather not since it's not that big of a deal to not have IMO

PR #1030, opened 2024-09-26
kwil-admin: setup genesis command

brennanjl:
Adds a `kwil-admin setup genesis` command. This creates a genesis file, allowing users to automatically generate valid genesis.json files that are up to date

### DIFF
--- a/cmd/kwil-admin/cmds/setup/genesis.go
+++ b/cmd/kwil-admin/cmds/setup/genesis.go
@@ -1,0 +1,216 @@
+package setup
+
+import (
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"math/big"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/kwilteam/kwil-db/cmd/common/display"
+	"github.com/kwilteam/kwil-db/common/chain"
+	"github.com/spf13/cobra"
+)
+
+var (
+	genesisLong = `` + "`" + `genesis` + "`" + `creates a new genesis.json file.
+
+This command allows users to create genesis files without having to write them manually.
+
+If the ` + "`" + `--migration` + "`" + ` flag is set, an incomplete genesis file is generated that can be used in a zero downtime migration.
+If generating a migration genesis file, validators and initial state cannot be set.
+
+Validators, balance allocations, and forks should have the format "name:key:power", "address:balance", and "name:height" respectively.`
+
+	genesisExample = `# Create a new genesis.json file in the current directory
+kwil-admin setup genesis --output .
+
+# Create a new genesis.json file in a specific directory with a specific chain ID and a validator with 1 power
+kwil-admin setup genesis --output /path/to/directory --chain-id mychainid --validator 890fe7ae9cb1fa6177555d5651e1b8451b4a9c64021c876236c700bc2690ff1d:1
+
+# Create a new genesis.json file to be used in a network migration
+kwil-admin setup genesis --migration --output /path/to/directory --chain-id mychainid`
+)
+
+func genesisCmd() *cobra.Command {
+	var validators, allocs, forks []string
+	var chainID, output, genesisState string
+	var migration, withGasCosts bool
+	var maxBytesPerBlock, joinExpiry, voteExpiry, maxVotesPerBlock int64
+	cmd := &cobra.Command{
+		Use:     "genesis",
+		Short:   "`genesis` creates a new genesis.json file",
+		Long:    genesisLong,
+		Example: genesisExample,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			out := filepath.Join(output, "genesis.json")
+
+			makeErr := func(e error) error {
+				return display.PrintErr(cmd, fmt.Errorf("failed to create genesis file: %w", e))
+			}
+
+			genesisInfo := chain.DefaultGenesisConfig()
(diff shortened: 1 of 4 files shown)


cmd/kwil-admin/cmds/setup/genesis.go
```diff
@@ -0,0 +1,219 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/kwilteam/kwil-db/cmd/common/display"
+	"github.com/kwilteam/kwil-db/common/chain"
+	"github.com/spf13/cobra"
+)
+
+var (
+	genesisLong = `` + "`" + `genesis` + "`" + `creates a new genesis.json file.
```
jchappelow, line 19: Missing space.  Prints like this:

```
$  ./kwil-admin setup genesis -h
`genesis`creates a new genesis.json file.

This command allows users to create genesis files without having to write them manually.

If the `--migration` flag is set, an incomplete genesis file is generated that can be used in a zero downtime migration.
If generating a migration genesis file, validators and initial state cannot be set.

Validators, balance allocations, and forks should have the format "name:key:power", "address:balance", and "name:height" respectively.

Usage:
  kwil-admin setup genesis [flags]

Examples:
# Create a new genesis.json file in the current directory
kwil-admin setup genesis

```

cmd/kwil-admin/cmds/setup/genesis.go
```diff
@@ -0,0 +1,219 @@
...
+
+This command allows users to create genesis files without having to write them manually.
```
jchappelow, line 21: Maybe without the third party verbage: "This command creates a new genesis file with optionally specified modifications."

cmd/kwil-admin/cmds/setup/genesis.go
```diff
@@ -0,0 +1,219 @@
...
+	cmd.Flags().StringVar(&chainID, chainIDFlag, "", "Chain ID for the genesis.json file")
+	cmd.Flags().StringArrayVar(&validators, validatorsFlag, nil, "Public keys and power of initial validator(s)")
+	cmd.Flags().StringArrayVar(&allocs, allocsFlag, nil, "Address and initial balance allocation(s)")
+	cmd.Flags().StringArrayVar(&forks, forksFlag, nil, "Block height and name of fork(s)")
+	cmd.Flags().BoolVar(&withGasCosts, withGasCostsFlag, false, "Include gas costs in the genesis file")
+	cmd.Flags().Int64Var(&maxBytesPerBlock, maxBytesPerBlockFlag, 0, "Maximum number of bytes per block")
+	cmd.Flags().Int64Var(&joinExpiry, joinExpiryFlag, 0, "Number of blocks before a join proposal expires")
+	cmd.Flags().Int64Var(&voteExpiry, voteExpiryFlag, 0, "Number of blocks before a vote proposal expires")
+	cmd.Flags().Int64Var(&maxVotesPerBlock, maxVotesPerTxFlag, 0, "Maximum number of votes per validator transaction (each validator has 1 validator tx per block)")
+	cmd.Flags().BoolVar(&migration, migrationFlag, false, "Generate an incomplete genesis file for zero downtime migration")
```
jchappelow, line 189: With **no** flags, the output genesis includes:

```json
    "migration": {
      "start_height": 0,
      "end_height": 0
    },
```

That's super confusing. I think we should modify the struct to omit empty unless setting the default empty to nil instead a poniter to zero value struct causes problems downstream somewhere.

cmd/kwil-admin/cmds/setup/genesis.go
```diff
@@ -0,0 +1,219 @@
...
+If the ` + "`" + `--migration` + "`" + ` flag is set, an incomplete genesis file is generated that can be used in a zero downtime migration.
+If generating a migration genesis file, validators and initial state cannot be set.
+
+Validators, balance allocations, and forks should have the format "name:key:power", "address:balance", and "name:height" respectively.`
+
+	genesisExample = `# Create a new genesis.json file in the current directory
+kwil-admin setup genesis
+
+# Create a new genesis.json file in a specific directory with a specific chain ID and a validator with 1 power
+kwil-admin setup genesis --output /path/to/directory --chain-id mychainid --validator 890fe7ae9cb1fa6177555d5651e1b8451b4a9c64021c876236c700bc2690ff1d:1
```
jchappelow, line 32: This validator example looks inaccurate.  Format in the code is "name:pubkeyhex:power"

cmd/kwil-admin/cmds/setup/genesis.go
brennanjl: Ah good point

common/chain/chaincfg.go
```diff
@@ -129,10 +129,10 @@ type VersionParams struct {
 
 type MigrationParams struct {
 	// StartHeight is the height from which the state from the old chain is to be migrated.
-	StartHeight int64 `json:"start_height"`
+	StartHeight int64 `json:"start_height,omitempty"`
```
jchappelow, line 132: Oh yeah this won't mess anything up like a nil struct ptr but is less confusing :+1:

jchappelow (approved):

common/chain/chaincfg.go
brennanjl: Yeah I went to go make a nil struct pointer but I am finding a _ton_ of places where we might have dereference errors

PR #1032, opened 2024-09-26
update security policy supported versions

jchappelow:


brennanjl (dismissed):

brennanjl:
So I guess no lts for v0.8?

brennanjl:
Nbd IMO, but curious why

jchappelow:
I updated it with 0.8 yesterday after chat.  Will you verify again that the table is what we want?

SECURITY.md
```diff
@@ -5,8 +5,9 @@
 | Version | Supported |
 | ------- | --------- |
 | Latest beta or release candidate    | ✅        |
+| v0.9.x  | ✅        |
 | v0.8.x  | ✅        |
-| < v0.8  | ❌        |
+| < v0.9  | ❌        |
```
brennanjl, line 10: Seems incorrect, this says that we do not support anything less than v0.9. It should be `< v0.8`

jchappelow: derp oh yeah sorry

brennanjl (approved):

PR #1033, opened 2024-09-27
fix output

brennanjl:
Fixes a conflicting flag name in the `genesis` command. For some reason, Cobra did not complain, but instead silently failed to add the flag.


charithabandi (approved):
Lgtm!

PR #1036, opened 2024-09-27
Simplify Migration Structs

brennanjl:
This PR simplifies migration structs. There are three key changes:

1. It removes `chain-id` from the migration proposals, since they are effectively a suggestion anyways.
2. It simplifies the struct that is passed back from the client when querying for migrations. It previously used to return serialized genesis.json, but now returns a much slimmed down struct that only contains necessary info.
3. It adds kwild's minor version to the migration metadata when querying a node. This is because we still send the snapshot metadata serialized, and thus need to maintain compatibility between versions. I thought about moving snapshot metadata into `core/types`, but didn't to keep it simple.

### DIFF
--- a/cmd/kwild/server/migration.go
+++ b/cmd/kwild/server/migration.go
@@ -21,6 +21,7 @@ import (
 	"github.com/kwilteam/kwil-db/core/log"
 	"github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/internal/abci/cometbft"
+	"github.com/kwilteam/kwil-db/internal/migrations"
 	"github.com/kwilteam/kwil-db/internal/statesync"
 )
 
@@ -141,6 +142,13 @@ func (m *migrationClient) downloadGenesisState(ctx context.Context) error {
 	metadata, err := m.clt.GenesisState(ctx)
 	if err != nil {
 		return err
+	}
+
+	// this check should change in every version:
+	// For backwards compatibility, we should be able to unmarshal structs from previous versions.
+	// Since v0.9 is our first time supporting migration, we only need to check for v0.9.
+	if metadata.Version != migrations.MigrationVersion {
+		return fmt.Errorf("genesis state download is incompatible. Received version: %d, supported versions: [%d]", metadata.Version, migrations.MigrationVersion)
 	}
 
 	// Check if the genesis state is ready
--- a/cmd/kwild/server/migration.go
+++ b/cmd/kwild/server/migration.go
@@ -60,29 +60,29 @@ func PrepareForMigration(ctx context.Context, kwildCfg *commonCfg.KwildConfig, g
 	logger.Info("Entering migration mode", log.String("migrate_from", kwildCfg.MigrationConfig.MigrateFrom))
 
 	snapshotFileName := config.GenesisStateFileName(kwildCfg.RootDir)
-	// check if genesis hash is set in the genesis config
-	if genesisCfg.DataAppHash != nil &&
-		genesisCfg.ConsensusParams.Migration.StartHeight != 0 &&
-		genesisCfg.ConsensusParams.Migration.EndHeight != 0 &&
-		validateGenesisState(snapshotFileName, genesisCfg.DataAppHash) {
-		// genesis state already downloaded. No need to poll for genesis state
+
+	// if the genesis state is already downloaded, then no need to poll for genesis state
+	_, err := os.Stat(snapshotFileName)
+	if err == nil {
 		logger.Info("Genesis state already downloaded", log.String("genesis snapshot", snapshotFileName))
+
+		if !validateGenesisState(snapshotFileName, genesisCfg.DataAppHash) {
+			return nil, nil, errors.New("app hash does not match the genesis state")
+		}
+
 		return kwildCfg, genesisCfg, nil
+	} else if !os.IsNotExist(err) {
+		return nil, nil, fmt.Errorf("failed to check genesis state file: %w", err)
 	}
 
 	// if we reach here, then we still need to download the genesis state
-	// Therefore, the genesis app hash, validator set, chain id, initial height,
-	// and migration info should not already be set in the genesis config.
-	if genesisCfg.DataAppHash != nil {
+	// Therefore, the genesis app hash, initial height, and migration info
+	// should not already be set in the genesis config.
+	if len(genesisCfg.DataAppHash) != 0 {
 		return nil, nil, errors.New("migration genesis config should not have app hash set")
 	}
-	if genesisCfg.Validators != nil {
-		return nil, nil, errors.New("migration genesis config should not have validators set")
-	}
-	if genesisCfg.ChainID != "" {
-		return nil, nil, errors.New("migration genesis config should not have chain id set")
-	}
 	if genesisCfg.InitialHeight != 0 && genesisCfg.InitialHeight != 1 {
+		// we are forcing users to adopt the height provided by the old chain
 		return nil, nil, errors.New("migration genesis config should not have initial height set")
 	}
 	if genesisCfg.ConsensusParams.Migration.IsMigration() {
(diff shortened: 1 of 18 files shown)


brennanjl:
Have not tested this btw

internal/version/version.go
```diff
@@ -15,19 +15,30 @@ import (
...
-const kwilVersion = "0.9.0-pre"
+const kwilVersion = "0." + MinorVersionV9 + ".0-pre"
+
+// MinorVersion is the minor version of the major version 0.
+type MinorVersion string
+
+// we start with 9 because we only started tracking this in v0.9.0
+const (
+	// MinorVersionV0 is the first minor version of the major version 0.
+	MinorVersionV9 MinorVersion = "9"
```
jchappelow, line 26: The rest of this PR makes great sense, but I don't quite get this part with "first minor version of the major version 0" because I don't see how it evolves or how it's relevant in the context of this package.  This purpose of this `version` package would be better reflected by:

```
const (
	Major = "0"
	Minor = "9"
	Patch = "1"

	// kwilVersionDefault is the default KwilVersion string used when not set by
	// the linker at build time.
	kwilVersionDefault = Major + "." + Minor + "." + Patch + "-pre"
)
```

But for the purposes of the migration struct doesn't this feel like we want a "Migration Protocol Version", an int in the migration package, like there are block, p2p, and app protocol versions?

brennanjl: Sure that makes sense

jchappelow: Like if we didn't change the shape of `MigrationMetadata` between two major release (hopefully it eventually stabilizes), the integer version should stay the same not bump for no reason?

jchappelow: or... "migration metadata version" if that's really the goal of allowing the receiver to interpret it appropriately.  So I guess migration *protocol* might suggest a much broader change than just the metadata structs.

brennanjl: I just reverted all of the changes made in the version package. Figured we can do that later, not really in scope here

brennanjl: > Like if we didn't change the shape of MigrationMetadata between two major release (hopefully it eventually stabilizes), the integer version should stay the same not bump for no reason

Yeah, correct.

brennanjl: At this point, the versioning is only needed for stability of the `Snapshot` struct, since everything else is sent structured.

core/types/types.go
```diff
@@ -126,8 +125,26 @@ type MigrationState struct {
 // for the migration.
 type MigrationMetadata struct {
 	MigrationState   MigrationState `json:"migration_state"`   // MigrationState is the current state of the migration
-	GenesisConfig    []byte         `json:"genesis_config"`    // GenesisConfig is the genesis file data
+	GenesisInfo      *GenesisInfo   `json:"genesis_info"`      // GenesisInfo is the genesis information
 	SnapshotMetadata []byte         `json:"snapshot_metadata"` // SnapshotMetadata is the snapshot metadata
+	Version          int            `json:"kwild_version"`     // Version of the migration metadata
```
jchappelow, line 130: Oh, json tag is still `kwild_version`

jchappelow (dismissed):
Wow this feels super clean.  Have not tested either and we definitely want @charithabandi to review and test, but LGTM.

brennanjl:
I'm going to test this over the weekend since it is such a big change

charithabandi:
Looking into this rn

internal/migrations/sql.go
```diff
@@ -51,16 +51,15 @@ var (
...
 	)`
 
 	// getMigrationSQL is the sql query used to get the current migration.
-	getMigrationSQL = `SELECT start_height, end_height, chain_id FROM ` + migrationsSchemaName + `.migration;`
+	getMigrationSQL = `SELECT start_height, end_height FROM ` + migrationsSchemaName + `.migration;`
 	// migrationIsActiveSQL is the sql query used  to check if a migration is active.
 	migrationIsActiveSQL = `SELECT EXISTS(SELECT 1 FROM ` + migrationsSchemaName + `.migration);`
 	// createMigrationSQL is the sql query used to create a new migration.
-	createMigrationSQL = `INSERT INTO ` + migrationsSchemaName + `.migration (id, start_height, end_height, chain_id) VALUES ($1, $2, $3, $4);`
+	createMigrationSQL = `INSERT INTO ` + migrationsSchemaName + `.migration (id, start_height, end_height) VALUES ($1, $2, $3, $4);`
```
charithabandi, line 62: `createMigrationSQL = `INSERT INTO ` + migrationsSchemaName + `.migration (id, start_height, end_height) VALUES ($1, $2, $3);`

should take only 3 arguments here instead of 4

brennanjl: Ahh thank you

jchappelow:
Regarding CI failure, on the "unhealthy" node1 I see:

```
2024-09-30 10:30:42 kwild version 0.9.0-pre (Go version go1.23.0)
2024-09-30 10:30:42 Root directory "/app/kwil"
2024-09-30 10:30:42 Private key path: /app/kwil/private_key
2024-09-30 10:30:42 2024-09-30T15:30:42.811Z    info    kwild   version 0.9.0-pre commit 
2024-09-30 10:30:42 2024-09-30T15:30:42.811Z    info    kwild   Entering migration mode {"migrate_from": "http://01d8aa26-5e5f-4756-95af-a6c614fb0aa2-node1-1:8484"}
2024-09-30 10:30:42 2024-09-30T15:30:42.818Z    info    kwild   Requesting genesis state from the old chain     {"listen_address": "http://01d8aa26-5e5f-4756-95af-a6c614fb0aa2-node1-1:8484"}
2024-09-30 10:30:42 2024-09-30T15:30:42.822Z    info    kwild   Genesis state not available     {"error": "status NoActiveMigration", "retry after(sec)": 30}
```

So it never gets healthy, at least not in 30 sec

brennanjl:
Yeah we got it sorted, it was really really stupid.

I realized I was testing with a binary that was built from main, not this branch 🤦 . As soon as I tested with this branch I saw the error

brennanjl:
Ok just finished testing this. I feel good about it

cmd/kwild/server/migration.go
```diff
@@ -60,14 +61,33 @@ func PrepareForMigration(ctx context.Context, kwildCfg *commonCfg.KwildConfig, g
...
-		genesisCfg.ConsensusParams.Migration.EndHeight != 0 &&
-		validateGenesisState(snapshotFileName, genesisCfg.DataAppHash) {
-		// genesis state already downloaded. No need to poll for genesis state
+
+	// if the genesis state is already downloaded, then no need to poll for genesis state
+	_, err := os.Stat(snapshotFileName)
+	if err == nil {
 		logger.Info("Genesis state already downloaded", log.String("genesis snapshot", snapshotFileName))
+
+		if !validateGenesisState(snapshotFileName, genesisCfg.DataAppHash) {
```
jchappelow, line 70: `validateGenesisState` could use an update.  I don't think it should itself call `failBuild` (it's not in build.go and `PrepareForMigration` can return errors without having to panic).  Also, a bunch of logic there is obsoleted now that we're ensuring it's existence here.  The empty apphash logic in `validateGenesisState` is also a little questionable given the current changes, plus the `bytes.Equal` in the final return covers the length an nilness checks implicitly.

brennanjl: Oh yeah. Im just going to change this whole function since it is a bit convoluted in its usage

test/driver/operator/cli_driver.go
```diff
@@ -183,9 +183,9 @@ type respTxQuery struct {
 	} `json:"tx_result"`
 }
 
-func (o *OperatorCLIDriver) SubmitMigrationProposal(ctx context.Context, activationHeight, migrationDuration *big.Int, chainID string) ([]byte, error) {
+func (o *OperatorCLIDriver) SubmitMigrationProposal(ctx context.Context, activationHeight, migrationDuration *big.Int) ([]byte, error) {
 	var res display.TxHashResponse
-	err := o.runCommand(ctx, &res, "migrate", "propose", activationHeight.String(), migrationDuration.String(), chainID)
+	err := o.runCommand(ctx, &res, "migrate", "propose", "activation-period", activationHeight.String(), "duration", migrationDuration.String())
```
jchappelow, line 188: This doesn't need `--` on `activation-period`?

brennanjl: Oh whoops. Currently we do not test the cli for this, but yes you are correct

brennanjl:
Is this good to go? @charithabandi @jchappelow

jchappelow:
I am doing staging e2e this evening, but can merge.  Any issues can be resolved after

brennanjl:
Cool, thanks. Trying to get it merged b/c I know there will be merge conflicts with https://github.com/kwilteam/kwil-db/pull/1040

jchappelow:
> Cool, thanks. Trying to get it merged b/c I know there will be merge conflicts with #1040

They're minimal (I've resolved locally), but that's true

PR #1037, opened 2024-09-27
Always update the genesis config based on the state from the old netw…

charithabandi:


jchappelow:
I'm sure this is right, but what are the "correctness reasons"?  It's hard to tell from the description.  What was incorrect?

cmd/kwild/server/migration.go
```diff
@@ -139,33 +130,40 @@ func (m *migrationClient) downloadGenesisState(ctx context.Context) error {
 		return fmt.Errorf("failed to unmarshal genesis config: %w", err)
 	}
 
-	// Save the genesis state
-	var snapshotMetadata statesync.Snapshot
-	if err := json.Unmarshal(metadata.SnapshotMetadata, &snapshotMetadata); err != nil {
-		return fmt.Errorf("failed to unmarshal snapshot metadata: %w", err)
-	}
+	genesisStateExists := bytes.Equal(genCfg.DataAppHash, m.genesisCfg.DataAppHash) &&
+		validateGenesisState(m.snapshotFileName, genCfg.DataAppHash)
```
jchappelow, line 134: In 

```go
func validateGenesisState(filename string, appHash []byte) bool {
	// check if the genesis state file exists
	if _, err := os.Stat(filename); os.IsNotExist(err) {
		return false
	}

	genesisStateFile, err := os.Open(filename)
	if err != nil {
		return false
	}
```

Shouldn't that error with `os.Open` cause `failBuild`?

charithabandi: Why, if we can regenerate the file ?

jchappelow: Maybe I just misunderstand the intent of `validateGenesisState`.  If the file the file does not exist, OK.  If it exists, but `os.Open` fails, OK?  If open succeeds but gunzip or io.Copy errors (which would be the if the file is not readable too), NOT OK? I don't see why this last case should be OK but it's ok to ignore an error form `os.Open`.  Either case indicates something is wrong with the file.

jchappelow: > Why, if we can regenerate the file with correct permissions?

There's only a narrow chance of that working.  If it's not owned by you, you can't just recreated it with correct permissions.  Errors other than permissions will also prevent os.Create from working.

Fundamentally, if a the file exists, but it cannot be read, that's clearly not what is expected and should be treated as a corrupt file just as if gunzip or io.Copy were to fail.

charithabandi:
Had a conversation with Brennan, Instead of updating the configuration when it's wrong, we instead gonna pull the config just once and if we detect the genesis-state location set, we would skip the whole polling process and fail if the config is incorrect.

jchappelow:
> Had a conversation with Brennan, Instead of updating the configuration when it's wrong, we instead gonna pull the config just once and if we detect the genesis-state location set, we would skip the whole polling process and fail fast if the config is incorrect.

So when doing polling for genesis_state, will we still be able to use settings from our own genesis, or do we have to accept the downloaded genesis?  I'm still not clear on how modifications to genesis can be done in such an automated migration.  e.g. we want change a parameter for the new network, or the new network has a genesis document with new fields.

brennanjl:
> So when doing polling for genesis_state, will we still be able to use settings from our own genesis, or do we have to accept the downloaded genesis? I'm still not clear on how modifications to genesis can be done in such an automated migration. e.g. we want change a parameter for the new network, or the new network has a genesis document with new fields that the one pulled from the old network will omit.

The "correct" setup for a migration is that the user does not have genesis state nor do they have app_hash, validators, and chain_id set in their genesis. There are three cases I can see for when a user starts with `migrations.enable=true`:

1. A user starts without app_hash, chain_id, and validators in their genesis height. This is a proper configuration, and `kwild` should poll for migrations and expect to download this info.
2. A user starts with a valid genesis.json (app_hash, chain_id, validators, etc.), but _no_ genesis state. We should error out here, because the user stated that they wish to enter migration mode, but we cannot do that without overwriting settings of their genesis.json. I have a pr for this here https://github.com/kwilteam/kwil-db/pull/1036
3. A user starts with a valid genesis.json and with genesis state. We assume that they downloaded this previously and that there is no need to re-poll the configured node to download it again, and simply use these.

We currently have case 2 working; my PR makes case 1 work; this PR should make case 3 work (but currently does not, as Charitha was trying to do something different).

Do you guys agree with this?

jchappelow:
> Do you guys agree with this?

Yes.  Case 1 was my concern: respect all settings, just patch those settings.  I still feel like changing validators would be allowable however.

jchappelow:
> > I still feel like changing validators would be allowable however.
> 
> Why?

App hash is the only one that really *must* be patched.

For validators, maybe you want to start with a new validator set:
- some validators decide not to join the new network
- new operators take over in their place or just in addition
- existing network validators want need to change their private key rather than copy over

For Chain ID, that part of the proposal really doesn't need to be binding either.  A migration is proposed, it's approved primarily based on the migration heights, then as activation approaches it is decided on a different chain ID.

jchappelow:
On that subject, keeping the chain ID the same would be quite risky.  There's a good likelihood that old and new networks would be unintentionally bridged at the p2p layer, and validators would be unwittingly submitting duplicate consensus (precommit/prevote) votes getting them all penalized.

brennanjl:
> Constraints on anything other than app hash seem to be unnecessary constraints, although I admit forcing them does create less opportunity for mistakes.

I see.  Ok, I'm convinced of this. I will make these changes on https://github.com/kwilteam/kwil-db/pull/1036.

Do we still want to support chain id as part of proposal then? This makes it essentially just a suggestion.

jchappelow:
> Do we still want to support chain id as part of proposal then? This makes it essentially just a suggestion. I think the less options a user has here, the better.

I can imagine a world where literally every piece of genesis is in the proposal (as an opaque blob since only the new network understands it).  But I have been a little confused by the chain ID in the proposal in the current design exactly because it was a suggestion.  But I dunno, maybe it's still a good-to-have so that someone could manually sanity check the served genesis.json with their knowledge of the proposal... like a migration tag.

charithabandi:
closing this PR. Incoming changes from Brennan here: https://github.com/kwilteam/kwil-db/pull/1036/files

PR #1038, opened 2024-09-27
chain: bump appVersion to 1

jchappelow:
Consider this diff as **for discussion**.

The application "protocol version" is set in genesis as `ConsensusParams.Version.App`.  This is a cometbft consensus parameter, and we have always given it 0.

Major Kwil versions effectively change this version, and although we may keep it the same, the result of running incompatible Kwil versions on the same network is generally a hard to debug app hash mismatch as a result of any logic that affects its computation such as tx execution changes, the commit ID computed in pg logical replication, etc.

By updating the app version, we instead fail fast at startup when attempting to perform blocksync with a peer running an incompatible app version:

`{"level":"error","ts":1727467648.7178495,"logger":"kwild.cometbft","msg":"Error in validation","module":"blocksync","err":"wrong Block.Header.Version. Expected {11 1}, got {11 0}","goversion":"go1.23.1","mod_version":"(devel)","revision":"cb6b0a42de632c34e3f2eba206bdf3c0571fc594","build_time":1727463320,"dirty":true}`

Bumping the app version in the future can also be used to perform coordinated upgrades (in cometbft terminology) by bumping the app version at some planned height, at which point it forces all validators to update the software version *at the same time*.

The other place this version has an effect seems to be in statesync where it ensures the state from the peer has the expected app version in `state.Version.Consensus.App`.

I started looking at this is because I wanted to see if there are parameters in the genesis that could or should be mutable in a migration scenario.  This one jumped out as an obvious one.  So, two questions:

1. should we bump this default?
2. what are the implications in migrations and the work in progress today on automatic retrieval of genesis config?

### DIFF
--- a/cmd/kwild/server/build.go
+++ b/cmd/kwild/server/build.go
@@ -52,6 +52,24 @@ import (
 	"github.com/kwilteam/kwil-db/internal/voting/broadcast"
 )
 
+var (
+	// AppVersion is the application's "protocol version", signaling changes in
+	// the application's component of the state machine that could impact
+	// consensus. This version should be bumped whenever changes in the
+	// application's logic might cause nodes with different versions to come to
+	// different conclusions about the validity of transactions, blocks, votes,
+	// or the resulting state.
+	//
+	// It should NOT be bumped for changes that do not affect consensus, such as
+	// performance optimizations, logging updates, or bug fixes that do not
+	// alter how transactions are validated or blocks are processed.
+	//
+	// Custom applications, especially those use extensions to extend or modify
+	// the state machine's logic, should set this variable appropriately prior
+	// to starting the Server or executing the root command.
+	AppVersion uint64 = 1
+)
+
 // initStores prepares the datastores with an atomic DB transaction. These
 // actions are performed outside of ABCI's DB sessions. The stores should not
 // keep the db after initialization. Their functions accept a DB connection.
@@ -489,7 +507,7 @@ func buildAbci(d *coreDependencies, db *pg.DB, txApp abci.TxApp, snapshotter *st
 	cfg := &abci.AbciConfig{
 		GenesisAppHash:     d.genesisCfg.ComputeGenesisHash(),
 		ChainID:            d.genesisCfg.ChainID,
-		ApplicationVersion: d.genesisCfg.ConsensusParams.Version.App,
+		ApplicationVersion: AppVersion,
 		GenesisAllocs:      d.genesisCfg.Alloc,
 		GasEnabled:         !d.genesisCfg.ConsensusParams.WithoutGasCosts,
 		ForkHeights:        d.genesisCfg.ForkHeights,
--- a/cmd/kwild/server/cometbft.go
+++ b/cmd/kwild/server/cometbft.go
@@ -141,10 +141,12 @@ func newCometConfig(cfg *config.KwildConfig) *cmtCfg.Config {
 
 // extractGenesisDoc is used by cometbft while initializing the node to extract
 // the genesis configuration. Note that cometbft's GenesisDoc is a subset of
-// kwild's genesis file.
+// kwild's genesis file. The app version set in AppVersion is used to supply
+// cometbft with the application protocol version, which is determined by the
+// app code rather than a configurable value in our genesis config.
 func extractGenesisDoc(g *chain.GenesisConfig) (*cmttypes.GenesisDoc, error) {
 	// BaseConsensusParms => cometbft's ConsensusParms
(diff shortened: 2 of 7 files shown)


brennanjl:
Just to make sure I understand correctly, this is essentially a way to signal a breaking consensus logic change so that incompatible versions immediately fail?

brennanjl:
> what are the implications in migrations and the work in progress today on automatic retrieval of genesis config?

None. Previously we ignored this value anyways when reading the old network's consensus params, but with https://github.com/kwilteam/kwil-db/pull/1036 we don't even send this info to the new nodes anymore

jchappelow:
> Just to make sure I understand correctly, this is essentially a way to signal a breaking consensus logic change so that incompatible versions immediately fail?

That is my understanding, correct.

If there are any other implications or gotchas, I am not aware of them.  Would love for @charithabandi's input as well, but we have talked about this in the past and haven't identified much beyond that AFAIK.

brennanjl:
Gotcha. Yeah this makes sense to me, I'm very open to adding this

brennanjl:
lgtm but will wait for @charithabandi's feedback

charithabandi (approved):

charithabandi:
LGTM! I would like to test this out in terms of migrations and statesync.

jchappelow:
> LGTM! I would like to test this out in terms of migrations and statesync.

Note the following check in cometbft: https://github.com/cometbft/cometbft/blob/5a753df1df350ef0207baba4068d73f5a499dde7/statesync/syncer.go#L487-L491
I believe this means if the app reports an app version that does not match the state in the received snapshot, then it is an error.  Does that cover the statesync concern?

charithabandi:
> > LGTM! I would like to test this out in terms of migrations and statesync.
> 
> Note the following check in cometbft: https://github.com/cometbft/cometbft/blob/5a753df1df350ef0207baba4068d73f5a499dde7/statesync/syncer.go#L487-L491 I believe this means if the app reports an app version that does not match the state in the received snapshot, then it is an error. Does that cover the statesync concern?

Yes, we should track the app versions in the snapshot metadata and this is validated here by the cometbft

jchappelow:
> > > LGTM! I would like to test this out in terms of migrations and statesync.
> > 
> > 
> > Note the following check in cometbft: https://github.com/cometbft/cometbft/blob/5a753df1df350ef0207baba4068d73f5a499dde7/statesync/syncer.go#L487-L491 I believe this means if the app reports an app version that does not match the state in the received snapshot, then it is an error. Does that cover the statesync concern?
> 
> Yes, we should track the app versions in the snapshot metadata and this is validated here by the cometbft

I don't think we need to.  Version.App is a consensus parameter that cometbft manages.  It's part of the snapshot and *its* restored state.

brennanjl:
> I don't think we need to. Version.App is a consensus parameter that cometbft manages. It's part of the snapshot and its restored state.

Hmm, if app version is set in genesis, then this feels a bit weird that we are hard-coding it. I would personally be in favor of removing `version.app` from our genesis, and having Kwil insert its own based on our own hard-coded values. I doubt users will use this value (considering we are only now starting to use it).

jchappelow:
> > I don't think we need to. Version.App is a consensus parameter that cometbft manages. It's part of the snapshot and its restored state.
> 
> Hmm, if app version is set in genesis, then this feels a bit weird that we are hard-coding it while also leaving it configurable. I would personally be in favor of removing `version.app` from our genesis, and having Kwil insert its own based on our own hard-coded values. I doubt users will use this value (considering we are only now starting to use it).

That is a VERY good point.  Only use case is extensions effectively alter the state machine, then it should bump app version.

jchappelow:
OK, app version is now hard coded.

Best spot I could think for it was in `cmd/kwild/server.AppVersion`.  A custom app that might further alter application logic with extensions like changing hooks can set it before creating the server or executing the root command.  In most cases I think they'd start with kwild's base app version and increment/mask for their own changes.

In any case, it's removed from our genesis.json as it made no real sense to set it from the document since it signifies the application logic, which is our code.  Only cometbft needs to get it as a genesis value, not us.

jchappelow:
@charithabandi Feel ok about this?  Or do you still think that our application's statesync code needs to adapt somehow?  AFAICT, snapshots obtained in statesync get checked, but maybe you had a different concern?

charithabandi:
looking into this

charithabandi:
This looks good to me. 

CometBFT verifies the ABCI version by making an Info call and compares it with the app version in the genesis config. So these changes should work.

charithabandi (approved):

PR #1041, opened 2024-09-30
update our pg docker image to 16.4, and use in test

jchappelow:
This updates our customized `postgres` Docker image in `build/package/docker/postgres.dockerfile` to extend the official `postgres:16.4` image, bumped from 16.2.  I have been using 16.4 since it's release it's release about 2 months ago, and I think it's a good idea to update our official image before 0.9 release.

This PR also starts using *our* postgres image in all of our tests.  It was used for the pg-enabled unit tests, but not in acceptance or integration tests.  We should be testing our image more consistently.  What this means is not of the special startup flags or the injected pginit.sql script need to be part of the docker compose files since they are already part of our extended postgres image.

NOTE: I have already pushed `kwildb/postgres:16.4-1` to Docker Hub, but I have not yet tagged this as `latest`, which still points to `16.2-1`.  If all is well with this PR and subsequent CI runs, we should update the `latest` tag.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,7 +44,7 @@ jobs:
 
     services:
       postgres:
-        image: kwildb/postgres:16.2-1
+        image: kwildb/postgres:16.4-1
         env:
           POSTGRES_PORT: 5432
           POSTGRES_HOST_AUTH_METHOD: trust
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -145,7 +145,7 @@ tasks:
       - |
         docker run -p 5432:5432 -v kwil-pg-demo:/var/lib/postgresql/data \
             --shm-size 256m -e "POSTGRES_HOST_AUTH_METHOD=trust" \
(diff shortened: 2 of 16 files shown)


brennanjl (commented):
LGTM. Will wait to approve once you publish a latest Kwil Postgres image and update all references of `kwildb/postgres:16.4-1` to `kwildb/postgres:latest`

jchappelow:
> Will wait to approve once you publish a latest Kwil Postgres image

`latest` tag is pushed now.

> and update all references of `kwildb/postgres:16.4-1` to `kwildb/postgres:latest`

I'm a little uncertain about this step.  We previously pinned tests to a specific version, which seems best to me.  This will stabilize the tests on, say, a release branch even if we later push a new `latest`.  The other thing was that `latest` might force both CI and our workstation machines to keep asking docker hub for the current latest, whereas if it's a version it will just go with what it has.

brennanjl:
Ok good point. lgtm then

brennanjl (approved):

PR #1044, opened 2024-10-01
abci: workaround for snapshots at initial height having old time

jchappelow:
This is a kinda dopey way to avoid making a snapshot at `InitialHeight`, which may or may not be 1, where that block's timestamp is the genesis time set in genesis.json that is often quite a bit older than the next block.  This results in an expired header error on statesync observed:

`"msg":"Can't verify","module":"statesync","module":"light","err":"verify from #1 to #2 failed: old header has expired at 2024-09-20 19:15:21.961570453 +0000 UTC (now: 2024-10-01 16:17:07.80072020`

I wasn't eager to store block time in our meta store to do this the proper way, and we can't simply do this on block 2 since InitialHeight is not always 1.  Did I miss a much simpler way in my pre-lunch fog?  Should I have just injected the initial height at Abci construction?

### DIFF
--- a/cmd/kwild/server/build.go
+++ b/cmd/kwild/server/build.go
@@ -490,6 +490,7 @@ func buildAbci(d *coreDependencies, db *pg.DB, txApp abci.TxApp, snapshotter *st
 		GenesisAllocs:      d.genesisCfg.Alloc,
 		GasEnabled:         !d.genesisCfg.ConsensusParams.WithoutGasCosts,
 		ForkHeights:        d.genesisCfg.ForkHeights,
+		InitialHeight:      d.genesisCfg.InitialHeight,
 		ABCIDir:            abciDir,
 	}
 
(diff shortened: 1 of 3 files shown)


brennanjl:
Hmm, how old can a timestamp be before Comet freaks out?

jchappelow:
> Hmm, how old can a timestamp be before Comet freaks out?

10 hours, apparently

jchappelow:
> Should I have just injected the initial height at Abci construction?

Yeah, just changed to do this.  It's not as wonky, even though it's awkward that we're supposed to get it from cometbft in InitChain.  We know it's the same because we first gave it to cometbft when constructing things, so whatever.

charithabandi (approved):

jchappelow:
@charithabandi I almost think we could disable these quick snapshot creations for statsync enabled nodes, but this at least solves the issue Alexandr hit.  Maybe we reconsider after release.

PR #1046, opened 2024-10-01
rpc,txapp,voting: disable delete resolution route

jchappelow:
The delete resolution route was not registered with the tx app's router, nor was it tested, so this removes the client methods and the RPC service handler until we've properly tested the tx route (and motivated it).

### DIFF
--- a/core/rpc/client/admin/admin.go
+++ b/core/rpc/client/admin/admin.go
@@ -31,6 +31,6 @@ type AdminClient interface {
 	// Resolutions
 	CreateResolution(ctx context.Context, resolution []byte, resolutionType string) ([]byte, error)
 	ApproveResolution(ctx context.Context, resolutionID *types.UUID) ([]byte, error)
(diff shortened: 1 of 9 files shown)


charithabandi:
[core/rpc/client/admin/jsonrpc/client.go](https://github.com/kwilteam/kwil-db/blob/13f669cd339ffb5645b29234702e69055cf1c492/core/rpc/client/admin/jsonrpc/client.go#L260)  has DeleteResolution

jchappelow:
@kwilteam/dev  After this we're going to tag `core v0.3.0` unless anyone has anything else in mind to change in `core`.

brennanjl (approved):

brennanjl:
> @kwilteam/dev After this we're going to tag core/v0.3.0 unless anyone has anything else in mind to change in core.

sounds good to me

charithabandi (approved):
lgtm! no other changes

PR #1053, opened 2024-10-02
rpc: fix cors because of blocked OPTIONS

jchappelow:
Will add an OPTIONS unit test, but please see if this works with browser clients.

### DIFF
--- a/internal/services/jsonrpc/server.go
+++ b/internal/services/jsonrpc/server.go
@@ -275,7 +275,7 @@ func NewServer(addr string, log log.Logger, opts ...Opt) (*Server, error) {
 		s.authSHA = slices.Clone(authSHA[:])
 	} // otherwise no basic auth check
 
-	// JSON-RPC handler (POST)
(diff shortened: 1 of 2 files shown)


internal/services/jsonrpc/server.go
```diff
@@ -290,7 +290,7 @@ func NewServer(addr string, log log.Logger, opts ...Opt) (*Server, error) {
 	h = realIPHandler(h, cfg.proxyCount) // for effective rate limiting
 	h = recoverer(h, log)                // first, wrap with defer and call next ^
 
-	mux.Handle("POST "+pathRPCV1, h)
```
jchappelow, line 293: @Yaiba Turns out the new router magic was blocking `OPTIONS`. :(

Yaiba: ah, you're right, this `POST /xx` will only accept POST..

Yaiba (approved):
LGTM

PR #1054, opened 2024-10-03
rpc: test cors, and respect option on rest endpoints

jchappelow:
This adds unit tests to the `jsonrpc` server to ensure the http mux and middleware stack handle CORS correctly, accepting `OPTIONS` requests if the server has cors enabled.  The test also cover some basic json-rpc method handling behavior.

A second commit updates `internal/version/version.go` to report `v0.9.2-pre`.  We have tagged `v0.9.1`, and using the branch on a commit past that tag would ideally report a build from a "prerelease" revision of the next sequential version, which would be `v0.9.2`.  I normally don't bother to bump batch version ahead of time like this, but I think it makes sense.


Yaiba (approved):
lgtm

Issue #1057, opened 2024-10-03
setup init command ignores provided private key


jchappelow:
When using `kwil-admin setup init` with the `--app.private-key-path` flag, it should use the specified private key file instead of a newly generated private key when generating the new root directory.

However, it seems to ignore the flag altogether:

```sh
$  ./kwil-admin setup init --genesis prod-chain/genesis.json --chain.moniker "node0" \
     --app.private-key-path ./private_key0 --root-dir prod-chain/node0
Initialized node in /home/jon/kwil/git/kwil-db/cmd/kwil-admin/prod-chain/node0
```

but

```sh
$  cat ./private_key0
a33488a470321da703c024cd0af729552a7f9c5fe9bf8b86b4c481455d1093db8e11018570091d79b6105d5954cbbc4f1570cd3a02df2214b362e7c3f22c0459

$  cat prod-chain/node0/private_key 
f786b944c8e2bd5dced708ca5fdbe0de0109f1a770f4d0e72fafcba6cae7a812b842f365bbb495b286d65f634a3dac7e9a405f9060e1ee779bccf8c721f11933
```

So then I thought it must have (unfortunately) just put the specified path into the config.toml, but it didn't do that either:

```
$  grep private_key_path prod-chain/node0/config.toml 
private_key_path = "private_key"
```

To be clear, I would like `setup init` to use the contents of the provided private key, not just update the `private_key_path` setting in the generated config.toml.

PR #1058, opened 2024-10-03
admin: use specified private key with setup init

jchappelow:
Requires and based on https://github.com/kwilteam/kwil-db/pull/1059

Resolves https://github.com/kwilteam/kwil-db/issues/1057

This change alters how `setup init` handles the `--app.private_key_path` flag.
Recently it was modified to ignore the flag entirely, with the apparent intent of ensuring the generated `config.toml` in the initialized root directory would always be `private_key_path = "private_key"` with the actual private key always being written into the root directory. This behavior is good, however, when the flag is set it is desirable to import the private key file pointed to by this flag, which is what this change does.

This new behavior streamlines the production deployment guide where the workflow for the genesis validator set involves preparing the private keys so that then can be used in the genesis file created with `kwil-admin setup genesis` first.

### DIFF
--- a/cmd/kwil-admin/cmds/setup/init.go
+++ b/cmd/kwil-admin/cmds/setup/init.go
@@ -3,7 +3,6 @@ package setup
 import (
 	"errors"
 	"fmt"
-	"io"
 	"math/big"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


jchappelow:
Not urgent @brennanjl, but I wanted to ensure this behavior isn't out of line with your intent for this command, particularly since you recently updated it.  I think this is a minor change and should not cause any problems, but it does help some workflows where you want to initialize your node's root directory with an existing private key and existing genesis.json.

charithabandi (approved):
LGTM!

jchappelow:
@charithabandi Would you please re-approve?  I added docs and de-duped some other instances of copyfile

brennanjl (approved):

PR #1059, opened 2024-10-03
use go workspace mode everywhere

### LINKED ISSUE #1048: get CI and all build scripts using Go workspace mode so we never need replaces in go.mod
Go workspace mode supposedly makes it so we don't ever need to do this `replace` dance in `go.mod` files.  However we still have CI and build scripts that bounce in and out of `GOWORK` mode.  Ideally we should get everything working with just a correct `go.work` and never need to add in `replace` statements to `go.mod` during development.

See https://go.dev/blog/get-familiar-with-workspaces and https://go.dev/doc/tutorial/workspaces.  These are somewhat outdated, but the motivation is there.

jchappelow:
Use Go workspace mode everywhere (no more need for `replace` statements in go.mod files).

Resolves #1048


jchappelow:
Whatever workspace issues we had before seem to be gone now @Yaiba

Yaiba (approved):
LGTM

PR #1062, opened 2024-10-07
[preview] Support TrustPeriod config in config.toml

charithabandi:
Forgot to add the config.toml support in the earlier pr, adding it here.


jchappelow (approved):
OK, but I have not tested if `trust_period = "36000h"` reads and parses as intended.  I suggest adding a test to `config_test.go` and that one test toml in `test_data` with an altered value, but also in hours.

jchappelow (approved):
Looks like CI will not run tests for `preview` PRs, but this still looks good.

PR #1063, opened 2024-10-07
Expose TrustPeriod flag for snapshots

charithabandi:


cmd/kwild/server/build.go
```diff
@@ -768,20 +768,30 @@ func buildStatesyncer(d *coreDependencies, db sql.ReadTxMaker) *statesync.StateS
 
 	providers := strings.Split(d.cfg.ChainConfig.StateSync.RPCServers, ",")
 
-	if len(providers) == 0 {
+	if len(providers) == 0 || d.cfg.ChainConfig.StateSync.RPCServers == "" {
```
jchappelow, line 771: Oh good catch.  I'm not sure how it would ever be `len(providers) == 0`, but might as well keep that.

jchappelow (dismissed):

brennanjl:
@charithabandi seems to be a lint issue

jchappelow (approved):

PR #1075, opened 2024-10-25
Feature/v0.10 sql ddl

Yaiba:
This pr adds new DDLs:
- create table 
- alter table
- create index
- drop index


parse/grammar/KuneiformParser.g4
```diff
@@ -147,6 +156,29 @@ constraint:
...
+    ON (UPDATE|DELETE)
+    (SET NULL
+    | SET DEFAULT
+    | RESTRICT
+    | NO ACTION
+    | CASCADE)
+;
+
+fk_constraint:
+    REFERENCES table=identifier (LPAREN column=identifier RPAREN) (fk_action (fk_action)*)?
```
brennanjl, line 179: Probably better to do:

```
REFERENCES table=identifier (LPAREN column=identifier RPAREN) (fk_action (fk_action)?)?
```

so that only a max of 2 can be specified

parse/grammar/KuneiformParser.g4
```diff
@@ -184,13 +216,58 @@ sql:
 
 sql_statement:
     (WITH RECURSIVE? common_table_expression (COMMA common_table_expression)*)?
-    (select_statement | update_statement | insert_statement | delete_statement)
+    (create_table_statement | alter_table_statement| create_index_statement | drop_index_statement // ddl
```
brennanjl, line 219: This seems incorrect. DDL shouldn't have CTEs. For example, I cant have:

```sql
WITH my_cte AS (
-- ...
)
CREATE TABLE ...
```

These should instead be separate clauses

parse/grammar/KuneiformParser.g4
```diff
@@ -184,13 +216,58 @@ sql:
...
+    | ADD COLUMN column=identifier type                                                # add_column
+    | DROP COLUMN column=identifier                                                    # drop_column
+    | RENAME COLUMN old_column=identifier TO new_column=identifier                     # rename_column
+    | RENAME TO new_table=identifier                                                   # rename_table
+    | ADD constraint_def                                                               # add_table_constraint
+    | DROP CONSTRAINT identifier                                                       # drop_table_constraint
+;
+
+create_index_statement:
+    CREATE UNIQUE? INDEX name=identifier? ON table=identifier LPAREN  columns=identifier_list RPAREN
```
brennanjl, line 264: Can you add `if not exists`?

parse/grammar/KuneiformParser.g4
```diff
@@ -184,13 +216,58 @@ sql:
...
+    (create_table_statement | alter_table_statement| create_index_statement | drop_index_statement // ddl
+    | select_statement | update_statement | insert_statement | delete_statement) // dml
 ;
 
 common_table_expression:
     identifier (LPAREN (identifier (COMMA identifier)*)? RPAREN)? AS LPAREN select_statement RPAREN
 ;
 
+create_table_statement:
+    CREATE TABLE name=identifier
```
brennanjl, line 228: Can you add `if not exists`?

parse/grammar/KuneiformParser.g4
```diff
@@ -184,13 +216,58 @@ sql:
...
+constraint_def: // TODO: rename to table_constraint
+    (CONSTRAINT name=identifier)?
+    unnamed_constraint
+;
+
+unnamed_constraint:
+    PRIMARY KEY LPAREN identifier_list RPAREN
+    | UNIQUE LPAREN identifier_list RPAREN
+    | CHECK LPAREN sql_expr RPAREN
+    | FOREIGN KEY LPAREN identifier RPAREN fk_constraint
```
brennanjl, line 244: not a huge deal but it might be easier if these are named? e.g.

```
unnamed_constraint:
    PRIMARY KEY LPAREN identifier_list RPAREN #unnamed_constraint_pk
    | UNIQUE LPAREN identifier_list RPAREN #unnamed_contract_unique
```

brennanjl (commented):
A few comments but mostly looked great

parse/grammar/KuneiformParser.g4
Yaiba: because we only support `on DELETE` and `on UPDATE`? make sense

parse/grammar/KuneiformParser.g4
Yaiba: sure, Do we need this for other statements, like `alter table add column if not exist xxx`?  
I only create the syntax you put in the docs

parse/grammar/KuneiformParser.g4
Yaiba: oh  I wanted to make a quick pass first then create another statement, totally forget about it.  A new `ddl_statement` sounds good?

parse/grammar/KuneiformParser.g4
brennanjl: Yeah that'd be great.

parse/grammar/KuneiformParser.g4
brennanjl: I think that is all postgres supports too.

parse/grammar/KuneiformParser.g4
brennanjl: Actually I don't think this is important. As long as we have it for `create` that is ok for now. We can add it in a new release

Yaiba:
To deal with new DDL, I create `ParseDDL` and `WriteDDL` for now so I don't break the downstream analyzer/planner, also allow me to write tests.

parse/grammar/KuneiformParser.g4
Yaiba: the syntax of the constraint is simple enough, for now it's done just in one switch, I don't think it's necessary to create different tree nodes on constraints.

parse/grammar/KuneiformParser.g4
```diff
@@ -177,12 +203,19 @@ procedure_return:
     The following section includes parser rules for SQL.
 */
 
-// sql is a top-level SQL statement.
-sql:
-    sql_statement SCOL
+// sql is a top-level SQL statement, it maps to SQLStmt interface in AST.
+sql_stmt:
```
Yaiba, line 207: To change `sql_statement` to `sql_stmt` in action/procedure block will break the parser/analyzer/planner, better do it together when we make the changes

brennanjl: Agreed

parse/grammar/KuneiformParser.g4
```diff
@@ -191,6 +224,47 @@ common_table_expression:
     identifier (LPAREN (identifier (COMMA identifier)*)? RPAREN)? AS LPAREN select_statement RPAREN
 ;
 
+create_table_statement:
+    CREATE TABLE (IF NOT EXISTS)? name=identifier
+    LPAREN
+    (table_column_def | table_constraint_def | table_index_def)
+    (COMMA  (table_column_def | table_constraint_def | table_index_def))*
+    RPAREN
+;
```
brennanjl, line 233: drop table?

parse/parse.go
```diff
@@ -275,6 +275,41 @@ type SQLParseResult struct {
...
+
+	// Mutative is true if the statement mutates state.
+	Mutative bool
+}
+
+// ParseDDL parses an SQL DDL statement.
+// NOTE: THIS is only temporary so that I can write tests. After we shift from
+// *SQLStatement to SQLStmt we can delete this.
+func ParseDDL(sql string, schema *types.Schema, skipValidation bool) (res *DDLParseResult, err error) {
+	parser, errLis, visitor, deferFn, err := setupParser2(sql)
```
brennanjl, line 295: Why do you need the schema parameter? It is unused.

parse/parse.go
```diff
@@ -275,6 +275,41 @@ type SQLParseResult struct {
...
+// NOTE: THIS is only temporary.
+type DDLParseResult struct {
+	// AST is the abstract syntax tree of the SQL statement.
+	AST SQLStmt
+	// Errs are the errors that occurred during parsing and analysis.
+	// These include syntax errors, type errors, etc.
+	ParseErrs ParseErrs
+
+	// Mutative is true if the statement mutates state.
+	Mutative bool
```
brennanjl, line 288: All DDL is mutative, so this is redundant

parse/grammar/KuneiformParser.g4
Yaiba: ah shoot...  Will add it now

parse/parse.go
brennanjl: skipValidation is unused as well

Yaiba: ah didn't event look the function signature, just copied from `ParseSQL`, will change.

parse/grammar/KuneiformParser.g4
Yaiba: hmm, do you want to add `drop table` in our DDL docs? Or do I just use syntax from https://www.postgresql.org/docs/current/sql-droptable.html ?

brennanjl: Yeah we should add drop ddl.

brennanjl: Just base it on postgres's tho

Yaiba:
OK Just add `Drop table` statement, also address feedback on `ParseDDL` function.

brennanjl (approved):

PR #1088, opened 2024-11-21
Preview block timing

brennanjl:
Sets the current block and block timestamp to the most recently seen block if in a `call`.


brennanjl:
One this is good to go, I will cherry-pick to main and v0.9. I did it here first because it's at the request of @sapience

jchappelow:
Only alternative that comes to mind is that `usersvc.(*Service).Call` could do:

```go
stat, _ := svc.chainClient.Status(ctx)
height, stamp := stat.Sync.BestBlockHeight, stat.Sync.BestBlockTime
```

Does that work?   I haven't digested the need for the committed vs uncommitted in the current approach.

brennanjl:
Closed in favor of https://github.com/kwilteam/kwil-db/pull/1094

PR #1101, opened 2024-11-28
[v0.9] add all-scalar flag

brennanjl:
Adds https://github.com/kwilteam/kwil-db/pull/1100 to release

PR #1102, opened 2024-11-28
All scalar

brennanjl:
adds https://github.com/kwilteam/kwil-db/pull/1100 to main


jchappelow:
Only need https://github.com/kwilteam/kwil-db/pull/1101 for 0.9 now, as per discussion yesterday.

PR #1107, opened 2024-12-04
fixes resolution status query in adminsvc

charithabandi:


jchappelow (approved):
TY. I thought it might need SQL changes, so this is a relief

PR #1113, opened 2024-12-04
accountsHash determinism, and global types.Hash use

jchappelow:
I came across what looks like some non-determinism in the computation of `accountsHash` while converting some uses of `sha256.Sum256` to our `types.Hash`/`types.Hasher`/`types.HashBytes`.

@charithabandi pls correct me if I'm wrong, but `ce.accounts.Updates()` code seems to range over a map, but even if it sorted internally I think it's right to have CE do a sort its way, as the other hashes do.

### DIFF
--- a/node/consensus/block_executor.go
+++ b/node/consensus/block_executor.go
@@ -1,11 +1,12 @@
 package consensus
 
 import (
+	"bytes"
 	"context"
-	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
 	"fmt"
+	"slices"
 	"sort"
 
 	"github.com/kwilteam/kwil-db/common"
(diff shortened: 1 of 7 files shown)


node/consensus/block_executor.go
```diff
@@ -204,29 +205,32 @@ func (ce *ConsensusEngine) nextAppHash(prevAppHash, changesetHash, valUpdatesHas
...
-	return types.Hash(hasher.Sum(nil))
+	return hasher.Sum(nil)
 }
 
 func (ce *ConsensusEngine) accountsHash() types.Hash {
 	accounts := ce.accounts.Updates()
-	hasher := sha256.New()
+	slices.SortFunc(accounts, func(a, b *ktypes.Account) int {
+		return bytes.Compare(a.Identifier, b.Identifier)
+	})
```
jchappelow, line 225: Very minor thing, just didn't want to forget.

charithabandi (approved):
THis makes sense. Thanks for the change.

PR #1114, opened 2024-12-04
[v0.9] version: update self-report tag for vanilla builds

jchappelow:
This updates the hard coded version string that is used for vanilla `go build`s to self report the version.  Official builds set it with an env var.


charithabandi (approved):
LGTM

PR #1124, opened 2024-12-06
Initialize the DB with the GenesisState if the node is bootstrapping for the first time

charithabandi:


config/config.go
```diff
@@ -80,6 +82,28 @@ func LoadGenesisConfig(filename string) (*GenesisConfig, error) {
...
+	for _, v := range gc.Validators {
+		hasher.Write(v.PubKey)
+		binary.Write(hasher, binary.BigEndian, v.Power)
+	}
+
+	binary.Write(hasher, binary.BigEndian, gc.MaxBlockSize)
+	binary.Write(hasher, binary.BigEndian, gc.JoinExpiry)
+	binary.Write(hasher, binary.BigEndian, gc.VoteExpiry)
+	binary.Write(hasher, binary.BigEndian, gc.DisabledGasCosts)
+	binary.Write(hasher, binary.BigEndian, gc.MaxVotesPerTx)
```
jchappelow, line 102: One point about these, but not necessarily needing change: these are almost too convenient in that we can casually change the type of these and this code will not break, but the result will change, which is a consensus breaker.  It's a lot more cumbersome to use the functions like binary.BigEndian.Append/PutUint64 etc, so let's go ahead with this now since we will certainly be changing `gc` a lot before release, but at some point I think we should harden this code.

core/types/types.go
```diff
@@ -44,6 +44,13 @@ type Validator struct {
 	Power  int64    `json:"power"`
 }
 
+// ByPubKey implements sort.Interface based on the PubKey field.
+type ByPubKey []Validator
+
+func (a ByPubKey) Len() int           { return len(a) }
+func (a ByPubKey) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a ByPubKey) Less(i, j int) bool { return a[i].PubKey.String() < a[j].PubKey.String() }
```
jchappelow, line 52: Sorting in Go has evolved a lot.  Back in the day you'd see a ton of `sort.Interface` implementations, but they realized the Len and Swap were almost always identical boilerplate.  Enter `sort.Slice` where you specify just the less function.  But now we would use the modern `slices.SortFunc` where `bytes.Compare` satisfies the new comparison (returning an `int` rather than a less bool).

node/consensus.go
```diff
@@ -415,10 +415,10 @@ func (n *Node) startDiscoveryRequestGossip(ctx context.Context, ps *pubsub.PubSu
...
 			}
 
-			// We're only interested if we are the validator.
-			if n.ce.Role() != types.RoleValidator {
-				continue // discard, we are just relaying to leader
-			}
+			// // We're only interested if we are the validator.
+			// if n.ce.Role() != types.RoleValidator {
+			// 	continue // discard, we are just relaying to leader
+			// }
```
jchappelow, line 421: Remove or uncomment?  Can't tell if this was a test.

node/consensus/blocksync.go
```diff
@@ -13,10 +13,12 @@ import (
 // caught up with the network, before it can start proposing blocks.
 // Else it can propose a block at a height that is already finalized
 // leading to a fork.
-
 func (ce *ConsensusEngine) doBlockSync(ctx context.Context) error {
 	if ce.role.Load() == types.RoleLeader {
-		if len(ce.validators.GetValidators()) == 1 {
+		// TODO: which validator set we should use here? whatever we have in the state?
+		// what if the current validators are not the same as the ones in the state?
```
jchappelow, line 19: I don't think I understand the difference between "current validators" and "the ones in the state".

node/consensus/engine.go
```diff
@@ -313,6 +316,8 @@ func New(cfg *Config) *ConsensusEngine {
 	return ce
 }
 
+var initialHeight int64 = 0 // TODO: get it from genesis?
```
jchappelow, line 319: Ah, good point.  Yeah we should add this to the genesis config struct.  Can do later though.

node/consensus/engine.go
```diff
@@ -337,6 +342,52 @@ func (ce *ConsensusEngine) Start(ctx context.Context, proposerBroadcaster Propos
 	return ce.runConsensusEventLoop(ctx)
 }
 
+// GenesisInit initializes the node with the genesis state. This included initializing the
+// votestore with the genesis validators, accounts with the genesis allocations and the
+// chain meta store with the genesis network parameters.
+// This is called only once when the node is bootstrapping for the first time.
+func (ce *ConsensusEngine) GenesisInit(ctx context.Context) error {
+	genesisTx, err := ce.db.BeginPreparedTx(ctx)
```
jchappelow, line 350: Not a prepared tx here, so just `BeginTx`

jchappelow (changes requested):
Only one thing I think should change, otherwise just some thought.

config/config.go
```diff
@@ -80,6 +82,28 @@ func LoadGenesisConfig(filename string) (*GenesisConfig, error) {
...
+
+	return hasher.Sum(nil), nil
```
jchappelow, line 104: Oh, I think we can have no error return? Unless we think there's a need to check for errors from binary.Write and I don't see one.

jchappelow (approved):

PR #1126, opened 2024-12-06
pg: in repl there is no fail, we get the stream or not

jchappelow:
With prepared transactions, where we are hashing for commit ID and collecting changesets during migration period, the sequence of queries and resulting logical replication messages is like this:

```go
		// - BEGIN;
		// - mods: UPDATE / INSERT / DELETE
		// - PREPARE TRANSACTION 'uid';
		//	* msgs: Begin Prepared -> [update messages] -> Prepare (ready to commit)
		// - COMMIT PREPARED 'uid';
		//  * msgs: Commit Prepared (NO regular "Commit" message)
```

If we `ROLLBACK PREPARED` (after `precommit`), we are discarding the prepared transaction that `postgres` has written to disk.  This is what emits the `RollbackPreparedMessageV3` in the logical replication stream.

But since this is after we have done `PREPARE TRANSACTION`, we have already hashed the changes and collected the changeset and, importantly, already closed the channel and set it to `nil`.  So, doing this again in response to the `RollbackPreparedMessageV3` has no effect, except to cause a data race on the field containing the channel if the next transaction starts immediately.

### DIFF
--- a/node/pg/repl_changeset.go
+++ b/node/pg/repl_changeset.go
@@ -557,25 +557,7 @@ func (c *changesetIoWriter) decodeDelete(delete *pglogrepl.DeleteMessageV2, rela
 // It exports the metadata to the writer.
 // It zeroes the metadata, so that the changeset can be reused,
 // and send a finish signal to the writer.
-func (c *changesetIoWriter) commit() error {
-	if c.csChan == nil {
-		return nil
-	}
-	// clear the relation index list for the next block
-	c.metadata = &changesetMetadata{
-		relationIdx: map[[2]string]int{},
-	}
-
-	// close the changes chan to signal the end of the changeset
-	close(c.csChan)
-	c.csChan = nil
-
-	return nil
-}
-
-// fail is called when the changeset is incomplete.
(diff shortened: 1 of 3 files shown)


node/consensus/engine_test.go
```diff
@@ -175,6 +175,11 @@ func verifyStatus(_ *testing.T, val *ConsensusEngine, status Status, height int6
 	return nil
 }
 
+func TestMain(m *testing.M) {
```
charithabandi, line 178: duplicated in the other PR

node/pg/repl_changeset.go
```diff
@@ -557,25 +557,7 @@ func (c *changesetIoWriter) decodeDelete(delete *pglogrepl.DeleteMessageV2, rela
 // It exports the metadata to the writer.
 // It zeroes the metadata, so that the changeset can be reused,
 // and send a finish signal to the writer.
-func (c *changesetIoWriter) commit() error {
```
charithabandi, line 560: 👌

charithabandi (approved):

PR #1130, opened 2024-12-06
CE clean shutdown fixes

charithabandi:


node/consensus/engine.go
```diff
@@ -403,7 +423,14 @@ func (ce *ConsensusEngine) GenesisInit(ctx context.Context) error {
 // Apart from the above events, the node also periodically checks if it needs to
 // catchup with the network and reannounce the messages.
 func (ce *ConsensusEngine) runConsensusEventLoop(ctx context.Context) error {
+	ce.wg.Add(1)
```
jchappelow, line 426: These have to be in the caller.

charithabandi: ok

node/consensus/engine.go
```diff
@@ -339,7 +342,25 @@ func (ce *ConsensusEngine) Start(ctx context.Context, proposerBroadcaster Propos
...
-	return ce.runConsensusEventLoop(ctx)
+	ce.wg.Add(1)
+	go ce.runConsensusEventLoop(ctx)
+
+	ce.wg.Wait()
+	ce.log.Info("Consensus engine stopped")
+	return nil
+}
+
+func (ce *ConsensusEngine) Close() {
```
jchappelow, line 353: I think this should be done after the `ce.wg.Wait` in `Start`.  (1) Exported like this it looks like this is how caller stops CE, but it's actually by cancelling Start's context, (2) if `runConsensusEventLoop` returns early, it can/should cancel an context internal to `Start` to get it to pass Wait and thus hit close to clean up any tx.

node/consensus/leader.go
```diff
@@ -183,6 +187,7 @@ func (ce *ConsensusEngine) processVotes(ctx context.Context) error {
 		// start the next round
 		ce.nextState()
 
+		ce.wg.Add(1)
```
jchappelow, line 190: This block below is one reason why I like to put the `wg.Done` always in an anonymous func.  The done is in `startNewRound`, but it's also needed in the `<-ctx.Done` return case.

charithabandi: Yeah, I removed startNewRounds from a go rountine, now we only have 1 go routine to manage which is the event loop

jchappelow (approved):
Oh, that was a nice iteration.  I think CI will like it too.

jchappelow:
Merge when ready!

PR #1131, opened 2024-12-06
fix node shutdown issues

jchappelow:
More shutdown fixes following https://github.com/kwilteam/kwil-db/pull/1130, but in Node and server.


jchappelow:
ffs

jchappelow:
k

charithabandi (approved):

PR #1141, opened 2024-12-10
Block processor for executing blocks and managing the consensus state

charithabandi:


node/nogossip.go
```diff
@@ -66,11 +66,17 @@ func (n *Node) txAnnStreamHandler(s network.Stream) {
...
+	} else {
+		n.mp.Store(txHash, rawTx)
+
+		// re-announce
+		go n.announceTx(context.Background(), txHash, rawTx, s.Conn().RemotePeer())
+	}
 
-	// re-announce
-	go n.announceTx(context.Background(), txHash, rawTx, s.Conn().RemotePeer())
+	fetched = true
```
jchappelow, line 79: Interesting problem here.  If checkTx fails and we don't store it, then it remains in the prefetch status forever.  We need this `fetched = true` only in the case where we store it.  If it reannounces we'll just have to refetch and recheck, but we don't want the prefetch map to fill or reject txns that might be valid later.

charithabandi: umm, I was thinking, if we should store rejected txs (basically all the seen transactions) in a cache (FIFO)

jchappelow: It might not be rejected forever though.  If a node somehow gets a tx with a none that is too high, it might be valid again later.  For example, a node starts at an arbitrary time that causes it to miss a predecesor txn, or if user broadcasts man quickly and one tx takes a longer gossip route than another.

charithabandi:
```
	ctx := context.Background()
	if err := n.ce.CheckTx(ctx, rawTx); err != nil {
		n.log.Warnf("tx %v failed check: %v", txHash, err)
	} else {
		n.mp.Store(txHash, rawTx)
		fetched = true

		// re-announce
		go n.announceTx(context.Background(), txHash, rawTx, s.Conn().RemotePeer())
	}
``` 
How about this?

charithabandi: umm, but this won't fetch other transactions :/

jchappelow: wdym by "other"?

jchappelow (approved):

PR #1146, opened 2024-12-11
use structured transaction in store and mempool

jchappelow:
Both the block store and mempool types were still using opaque transctions (`[]byte`) from the early version of the p2p demo.  This updates both to use `*types.Transaction`.

This change stops short of updating the `Block` type to have a `Txns []*Transaction`  field instead of `Txns [][]byte`, but that change should also be made.  This can wait until the block data / chain RPCs are complete.  It will also require updating the block serialization code, although the effective binary serialization will be identical.

This also updates the mempool implementation and interface:
- internally there is a FIFO queue so that `PeekN` will return those at the front of the queue.  The maps still remain there for fast indexing and existence checks.
- add a `Remove` method that is the same as `Store(txid, nil)`

### DIFF
--- a/core/crypto/auth/signer.go
+++ b/core/crypto/auth/signer.go
@@ -7,14 +7,14 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/kwilteam/kwil-db/core/crypto"
 	"golang.org/x/crypto/sha3"
+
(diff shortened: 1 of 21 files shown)


charithabandi (approved):
LGTM! Merge if you have no other changes.

PR #1147, opened 2024-12-11
feat(rpc): chain RPCs

### LINKED ISSUE #1121: Feature Request: Support RPC endpoints for Chain stats
### Describe the Feature

Bring back the endpoints that the Cometbft RPC server used to support that powers our indexer. 

- [ ] `/block` : Get block at a specified height
- [ ] `/block_by_hash`: Get block by hash 
- [ ] `/block_results`:  Get block results at a specified height
- [ ] `/unconfirmed_txs`: Get list of unconfirmed transactions
- [ ] `/num_unconfirmed_txs`: Get data about unconfirmed transactions
- [ ] `/tx_search`
- [ ] `/block_search`
- [ ] `/tx`: Get tx by hash
- [ ] `/check_tx`
- [ ] `/commit`: (not very sure about this, @jchappelow ) /block should satisfy this.
- [ ] `/genesis

- [ ] `/validators`: Get validator set at a specified height, this is a bit hard to satisfy and would require few more updates to the blockstore [Not very important for Beta]
- [ ] `/consensus_params`

Most of this can be satisfied with blockstore and node and votestore without getting deep into the consensus or node operations. 

## Indexer Requirements:
- block_search
- tx_search
- validators set at a specified height
- validators that actively participated in a block round <currently this info is not stored anywhere, something to support in future>
- number of accounts, nodes in the network

CometBFT Reference: https://docs.cometbft.com/v0.34/rpc/#/Info/genesis

Yaiba:
This resolves #1121

All APIs listed in the issue are implemented under `chain` rpc namespace.

These APIs are not implemented:
- `/num_unconfirmed_txs`,  `chain.unconfirmed_txs` returned the total txs
- `/tx_search`, it's more useful for tx subscription, we don't have the needs yet
- `/block_search`, it's more useful for block subscription
- `/check_tx`
- `/commit`, I don't think we have this kind of info

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,8 +6,7 @@ import (
 	"time"
 
 	"github.com/kwilteam/kwil-db/core/log"
-	ktypes "github.com/kwilteam/kwil-db/core/types"
-	"github.com/kwilteam/kwil-db/node/types"
+	"github.com/kwilteam/kwil-db/core/types"
 
 	"github.com/pelletier/go-toml/v2"
 )
@@ -41,7 +40,7 @@ type GenesisConfig struct {
 	// Leader is the leader's public key.
 	Leader types.HexBytes `json:"leader"`
 	// Validators is the list of genesis validators (including the leader).
-	Validators []*ktypes.Validator `json:"validators"`
+	Validators []*types.Validator `json:"validators"`
 
 	// MaxBlockSize is the maximum size of a block in bytes.
 	MaxBlockSize int64 `json:"max_block_size"`
@@ -89,7 +88,7 @@ func DefaultGenesisConfig() *GenesisConfig {
 		ChainID:          "kwil-test-chain",
 		InitialHeight:    0,
 		Leader:           types.HexBytes{},
-		Validators:       []*ktypes.Validator{},
+		Validators:       []*types.Validator{},
 		DisabledGasCosts: true,
 		JoinExpiry:       14400,
 		VoteExpiry:       108000,
--- a/node/consensus/interfaces.go
+++ b/node/consensus/interfaces.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	ktypes "github.com/kwilteam/kwil-db/core/types"
-	"github.com/kwilteam/kwil-db/node/mempool"
 	"github.com/kwilteam/kwil-db/node/types"
 	"github.com/kwilteam/kwil-db/node/types/sql"
 )
@@ -22,7 +21,6 @@ type DB interface {
 type Mempool interface {
 	PeekN(maxSize int) []types.NamedTx
 	Remove(txid types.Hash)
-	RecheckTxs(ctx context.Context, checkFn mempool.CheckFn)
 }
 
 // BlockStore includes both txns and blocks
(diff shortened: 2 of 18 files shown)


config/config.go
```diff
@@ -6,8 +6,7 @@ import (
 	"time"
 
 	"github.com/kwilteam/kwil-db/core/log"
-	ktypes "github.com/kwilteam/kwil-db/core/types"
-	"github.com/kwilteam/kwil-db/node/types"
```
Yaiba, line 10: can be simplify to just use core types

node/services/jsonrpc/funcsvc/service.go
```diff
@@ -68,14 +68,6 @@ func (svc *Service) Methods() map[jsonrpc.Method]rpcserver.MethodDef {
 	}
 }
 
-func (svc *Service) Handlers() map[jsonrpc.Method]rpcserver.MethodHandler {
```
Yaiba, line 71: I believe this is not used

node/interfaces.go
```diff
@@ -30,6 +30,8 @@ type ConsensusEngine interface {
 		blkRequester consensus.BlkRequester, stateResetter consensus.ResetStateBroadcaster, discoveryBroadcaster consensus.DiscoveryReqBroadcaster) error
 
 	CheckTx(ctx context.Context, tx *ktypes.Transaction) error
+
+	ConsensusParams() *ktypes.ConsensusParams
```
Yaiba, line 34: This is kind awkward, I have to define this on both ConsensusEngine and Node to pass over to chain rpc

Yaiba:
need to implement the client as well, don't review/merge yet

Yaiba:
OK, I think it's ready to be reviewed

charithabandi (dismissed):
Glanced through it, LGTM!

jchappelow:
@charithabandi resolution for the RecheckTxs issue?

jchappelow:
Just update the interface I suppose?

charithabandi:
yeah, merge conflict issue, we fixed it, @Yaiba pushing the updates

node/consensus/interfaces.go
```diff
@@ -22,7 +21,6 @@ type DB interface {
 type Mempool interface {
 	PeekN(maxSize int) []types.NamedTx
 	Remove(txid types.Hash)
-	RecheckTxs(ctx context.Context, checkFn mempool.CheckFn)
```
jchappelow, line 25: probably a rebase oopsie?

node/interfaces.go
jchappelow: On this point, I think we are likely to have block executor provide.  OK for now

node/consensus/interfaces.go
Yaiba: yeah seems so

jchappelow (approved):

node/interfaces.go
charithabandi: Yeah, we could have one more interface using  BlockProcessor to provide these functions.

charithabandi (approved):

PR #1150, opened 2024-12-11
Rollback support to preemptively cancel and rollback current block ex…

charithabandi:
…ecution
yet to add `kwil-admin` command to issue cancellation request and block execution status request. But rest is good.  

Added support for leader to manually trigger block cancellation and also provide a list of long running transactions to be purged from the mempool and re-propose a block for the same height. 
Leader upon receiving this cancellation request will broadcast the resetState messages to all the validators, who upon receiving the resetMsg will stop the block execution and rollback the state. 

If leader has already executed the transactions by the time we receive the cancellation request, should we still reject the current block and propose a new block or should it just ignore the request? Right now I am ignoring, but can be changed.

### DIFF
--- a/core/types/messages.go
+++ b/core/types/messages.go
@@ -1,5 +1,7 @@
 package types
 
+import "time"
+
 // This file contains the messages exchanged between the consensus engine and the block processor.
 
 type BlockExecRequest struct {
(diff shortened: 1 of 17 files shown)


jchappelow (approved):
LGTM with `resetCount` removed.  Will have to look at nested locking carefully, but I don't see any issues presently.

PR #1154, opened 2024-12-12
config: add comments, remove koanf tag, fix toml Duration

jchappelow:
This change adds "comment" tags, which are used in two places:
 - TOML comments for each field or section
 - cli flag descriptions

This also removes the unneeded "koanf" tags and now specify "toml" as the attribute tag used for config merging.

Replace a few time.Duration fields with config.Duration so these durations are set correctly as duration strings in TOML rather than giant integers.

Finally, remove the replaces from go.mod and go mod tidy.

`kwild print-config` example output:

```toml
# log level
# possible values are 'debug', 'info', 'warn', and 'error'
log_level = 'info'
# log format
# possible values are 'json', 'text' (kv), and 'plain' (fmt-style)
log_format = 'plain'
# private key to use for node
privkey = 'a4071a2debdd8c31f19cee9642586b9196af6cc236a11f39445d5123d82079eb'

# P2P related configuration
[p2p]
# IP address to listen on for P2P connections
ip = '127.0.0.1'
# port to listen on for P2P connections
port = 6601
# enable peer exchange
pex = true
# bootnodes to connect to on startup
bootnodes = ['0226b3ff29216dac187cea393f8af685ad419ac9644e55dce83d145c8b1af213bd#0@127.0.0.1:6600']

# Consensus related configuration
[consensus]
# timeout for proposing a block (applies to leader)
propose_timeout = '1s'
# max size of a block in bytes
max_block_size = 50000000
# max number of transactions per block
max_txs_per_block = 20000

# DB (PostgreSQL) related configuration
[db]
# postgres host name (IP or UNIX socket path)
host = '127.0.0.1'
# postgres TCP port (leave empty for UNIX socket)
port = '5433'
# postgres role/user name
user = 'kwild'
# postgres password if required for the user and host
pass = ''
# postgres database name
dbname = 'kwild'
# timeout on read transactions from user RPC calls and queries
read_timeout = '45s'
# maximum number of DB connections to permit
max_connections = 60

# User RPC service configuration
[rpc]
# address in host:port format on which the RPC server will listen
listen = '0.0.0.0:8485'
# user request duration limit after which it is cancelled
timeout = '20s'
# largest permissible user request size
max_req_size = 6000000
# enable private mode that requires challenge authentication for each call
private = false
# lifetime of a server-generated challenge
challenge_expiry = '30s'
# maximum number of challenges per second that a user can request
challenge_rate_limit = 10.0

# Admin RPC service configuration
[admin]
# enable the admin RPC service
enable = true
# address in host:port format or UNIX socket path on which the admin RPC server will listen
listen = '127.0.0.1:8585'
# optional password for the admin service
pass = ''
# disable TLS when the listen address is not a loopback IP or UNIX socket
notls = false
# TLS certificate for use with a non-loopback listen address when notls is not true
cert = 'admin.cert'
# TLS key for use with a non-loopback listen address when notls is not true
key = 'admin.key'

# Snapshot creation and provider configuration
[snapshots]
# enable creating and providing snapshots for peers using statesync
enable = false
# snapshot creation period in blocks
recurring_height = 14400
# number of snapshots to keep, after the oldest is removed when creating a new one
max_snapshots = 3

# Statesync configuration (vs block sync)
[state_sync]
# enable using statesync rather than blocksync
enable = false
# trusted snapshot providers in node ID format (see bootnodes)
trusted_providers = []
# how long to discover snapshots before selecting one to use
discovery_time = '30s'
# how many times to try after failing to apply a snapshot before switching to blocksync
max_retries = 3
```


config/config_test.go
```diff
@@ -87,7 +87,7 @@ func TestMarshalConfig(t *testing.T) {
 	k.Print()
 
 	var inCfg Config
-	k.UnmarshalWithConf("", &inCfg, koanf.UnmarshalConf{Tag: "koanf"})
+	k.UnmarshalWithConf("", &inCfg, koanf.UnmarshalConf{Tag: "toml"})
```
jchappelow, line 90: eh, this entire test shoudl be removed...

charithabandi (approved):

PR #1156, opened 2024-12-13
Stable sorts transactions in the block in nonce order, enforce consen…

charithabandi:
…sus limits, induce voteBody and voteID transactions

Support for the following things have been added in this PR:
- Ability to configure and run extensions
- Order Block transactions in nonce order and enforce consensus limits such as maxBlockSz, maxVotesPerTx
- Leader to induce ValidatorVoteBodies Tx if there are any events without resolutions
- Validators can broadcast their vote for the existing resolutions
- ethDeposits extension


extensions/listeners/eth_deposits/deposits.go
```diff
@@ -0,0 +1,338 @@
+// package ethdeposits implements an listener that listens to Ethereum events
+// and triggers the creation of deposit events in Kwil.
+package ethdeposits
```
jchappelow, line 3: Can we git mv this package instead?

charithabandi: oh sure!

node/txapp/txapp.go
```diff
@@ -101,18 +100,6 @@ func (r *TxApp) GenesisInit(ctx context.Context, db sql.DB, validators []*types.
...
-	// for _, hook := range hooks.ListGenesisHooks() {
-	// 	err := hook.Hook(ctx, &types.App{
-	// 		Service: r.service.NamedLogger(hook.Name),
-	// 		DB:      db,
-	// 		Engine:  r.Engine,
-	// 	}, chain)
-	// 	if err != nil {
-	// 		return fmt.Errorf("error running genesis hook: %w", err)
-	// 	}
-	// }
```
jchappelow, line 114: oh whoops, conflict here

core/types/payloads.go
```diff
@@ -606,15 +607,67 @@ type ValidatorVoteBodies struct {
 var _ Payload = (*ValidatorVoteBodies)(nil)
 
 func (v *ValidatorVoteBodies) MarshalBinary() ([]byte, error) {
-	return serialize.Encode(v)
+	buf := new(bytes.Buffer)
```
jchappelow, line 610: ty very much for doing this :pray:

node/txapp/txapp.go
charithabandi: hahah, rebased.

core/rpc/client/admin/jsonrpc/client.go
```diff
@@ -147,7 +147,6 @@ func (cl *Client) Status(ctx context.Context) (*adminTypes.Status, error) {
 			Syncing:         res.Sync.Syncing,
 		},
 		Validator: &adminTypes.ValidatorInfo{
-			Role:   res.Validator.Role,
```
jchappelow, line 150: Is this not still helpful to distinguish leader vs regular validator/follower?

go.mod
```diff
@@ -90,7 +108,7 @@ require (
 	github.com/docker/go-units v0.5.0 // indirect
 	github.com/dustin/go-humanize v1.0.1 // indirect
 	github.com/elastic/gosigar v0.14.3 // indirect
-	github.com/ethereum/go-ethereum v1.14.11 // indirect
+	github.com/ethereum/go-ethereum v1.14.11
```
jchappelow, line 111: Sorry, very silly nit, but can you move this into the direct imports since it's apparently directly used by deposits (I think) now.

jchappelow (approved):
LGTM.  Please squash or not as desired and merge when ready.  Thanks for addressing the in-place tx sort issue.

core/rpc/client/admin/jsonrpc/client.go
jchappelow: Oh, wait, we cannot deduce leader though, can we?

jchappelow: Also, what is confusing about it?

PR #1158, opened 2024-12-13
support offline migration and admin tool to create network snapshots …

charithabandi:
…for offline migration

### DIFF
--- a/app/node/build.go
+++ b/app/node/build.go
@@ -1,12 +1,14 @@
 package node
 
 import (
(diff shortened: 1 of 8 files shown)


jchappelow (approved):
OK.
I sorta wanted `app/shared/bind` to be more generic helpers for config merge, but it fits.  Better than exporting `GetPostgresFlags` from `app/reset` and using in `app/snapshot` I suppose.

PR #1163, opened 2024-12-16
Zero-Downtime migrations

charithabandi:


node/block_processor/processor.go
```diff
@@ -664,3 +737,11 @@ func (bp *BlockProcessor) SetNetworkParameters(params *common.NetworkParameters)
 		MigrationStatus:  params.MigrationStatus,
 	}
 }
+
+func (bp *BlockProcessor) GetMigrationMetadata(ctx context.Context) (*ktypes.MigrationMetadata, error) {
+	bp.mtx.RLock()
+	defer bp.mtx.RUnlock()
+
+	status := bp.chainCtx.NetworkParameters.MigrationStatus
+	return bp.migrator.GetMigrationMetadata(ctx, status)
```
jchappelow, line 746: Mutex use is a little use here and inside migrator.  `GetMigrationMetadata` will do a lot of work, potentially blocking for seconds on an RPC, correct?  Holding a mutex, even for read only, for longer than computer-perceivable durations should be avoided, especially if the mutex is broad or system level.

In this BP method, I think `bp.mtx` is only for network params as per `mtx sync.RWMutex // mutex to protect the consensus params` (:+1: ), so I would unlock it before calling into `bp.migrator.GetMigrationMetadata`.

I also noticed inside the `migrator.GetMigrationMetadata` impl, it's locking an RWMutex twice, which is a mostly harmless bug, but even in migrator, we should look at what kind of long blocking actions happen under mutex lock.

jchappelow (dismissed):
Nice migrator refactor.  Looks much prettier now.

Issue #1115, opened 2024-12-05
move private key out of config toml and back to private_key file


jchappelow:

jchappelow:
I think this one should be for the beta.  There can be a `type NodeKeyFile struct` that marshals to json with a "privkey" and "type" field to future proof

Issue #1116, opened 2024-12-05
update key gen and key info cmds to work with multiple key types


jchappelow:
Right now it just works with secp, and not a key with a key type suffix (e.g. `#0` for secp256k1 vs `#1` for ed25519)

PR #1164, opened 2024-12-16
app,config: define node key file and update cmds

jchappelow:
This resolves https://github.com/kwilteam/kwil-db/issues/1115 and https://github.com/kwilteam/kwil-db/issues/1116.


charithabandi (approved):
lgtm!

PR #1165, opened 2024-12-16
add go pprof capabilities

### LINKED ISSUE #1152: reenable profiling options
```
	// ProfileMode string `toml:"profile_mode"`
	// ProfileFile string `toml:"profile_file"`
```

and the code to start the profilers is gone, so we need to grab the `startProfilers` function from here: https://github.com/kwilteam/kwil-db/blob/d09e5bc74f8ad37d4e6b4ff122f0f6670471f5cc/cmd/kwild/root/root.go#L225 (and start it as before)

jchappelow:
This adds the Go pprof system back.  Resolves #1152

### DIFF
--- a/app/node/prof.go
+++ b/app/node/prof.go
@@ -1,0 +1,93 @@
+package node
+
+import (
+	"fmt"
+	"net/http"
+	_ "net/http/pprof"
+	"os"
+	"runtime"
+	"runtime/pprof"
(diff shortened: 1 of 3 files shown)


charithabandi (approved):
Awesome, would love to try this out today

charithabandi (approved):

PR #1167, opened 2024-12-17
more unit test coverage

jchappelow:
This adds more unit test coverage and fixes a number of minor bugs.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,8 @@
 run:
   timeout: 10m
   go: '1.22'
+  build-tags:
+    - pglive
 
 issues:
   exclude-dirs:
--- a/node/migrations/changesets_test.go
+++ b/node/migrations/changesets_test.go
@@ -132,6 +132,7 @@ func TestChangesetMigration(t *testing.T) {
 		Engine:     nil,
 	}
 	err = applyChangeset(ctx, app, tx, int64(height), int64(totalChunks))
+	require.NoError(t, err)
 
 	res, err := tx.Execute(ctx, "select * from ds_test.test", pg.QueryModeExec)
 	require.NoError(t, err)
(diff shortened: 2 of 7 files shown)


core/types/block_test.go
```diff
@@ -0,0 +1,370 @@
...
+	"encoding/binary"
+	"testing"
+	"time"
+
+	"github.com/kwilteam/kwil-db/core/crypto"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetRawBlockTx(t *testing.T) {
```
jchappelow, line 14: I actually did a git mv, but added enough new tests that git decided it wasn't similar enough to link.  :shrug:

core/types/decimal/decimal.go
```diff
@@ -127,7 +127,7 @@ func inferPrecisionAndScale(s string) (precision, scale uint16) {
 
 	intPart := uint16(len(parts[0]))
 	if len(parts) == 1 {
-		return intPart, 0
+		return max(1, intPart), 0 // at least 1 for precision with 0 scale
```
jchappelow, line 130: In `utils/conv/conv_test.go` I discovered that `"0"` could not be used as a valid Decimal.  I don't think that is intended, and this case catches if `strings.TrimLeft(parts[0], "0")` reduced it to an empty string.  @brennanjl lmk if a zero decimal is actually not supposed to be allowed.  Based on code in other places, it looks like it should have been.

node/discover.go
```diff
@@ -16,25 +15,11 @@ import (
...
-	buf := make([]byte, len(discoverPeersMsg))
-	nr, err := s.Read(buf)
-	if err != nil && !errors.Is(err, io.EOF) {
-		n.log.Warn("failed to read peer discovery request", "error", err)
-		return
-	}
-	if nr == 0 { // they hung up
-		return
-	}
-	if string(buf) != discoverPeersMsg {
```
jchappelow, line 33: This protocol was an early text based version.  The initiator need not send some magic text message to start it off.  The stream is specific to the protocol ID and ready to go.

charithabandi (approved):

PR #1168, opened 2024-12-17
rollback block admin commands

charithabandi:


node/services/jsonrpc/adminsvc/service.go
```diff
@@ -561,3 +571,45 @@ func (svc *Service) ResolutionStatus(ctx context.Context, req *adminjson.Resolut
...
+		txInfo[i] = &types.TxInfo{
+			ID:     txID,
+			Status: status.TxStatus[txID.String()],
+		}
+	}
+	resp.TxInfo = txInfo
+	return resp, nil
+}
+
+func (svc *Service) RollbackBlock(ctx context.Context, req *adminjson.RollbackBlockRequest) (*adminjson.RollbackBlockResponse, *jsonrpc.Error) {
```
jchappelow, line 598: I feel like this RPC is a big misleading.  You can't rollback any arbitrary block.  We were referring to it as "reset" or "abort", referring to an uncommitted consensus round?  Thoughts?

node/consensus/engine.go
```diff
@@ -332,7 +341,7 @@ func (ce *ConsensusEngine) runConsensusEventLoop(ctx context.Context) error {
 	ce.log.Info("Starting the consensus event loop...")
 	catchUpTicker := time.NewTicker(5 * time.Second)
 	reannounceTicker := time.NewTicker(3 * time.Second)
-	blkPropTicker := time.NewTicker(1 * time.Second)
+	blkPropTicker := time.NewTicker(10 * time.Second)
```
jchappelow, line 344: What's the significance of this duration value?  Should this be the configured propose timeout, or is this something else?

node/services/jsonrpc/adminsvc/service.go
charithabandi: That makes sense

node/consensus/engine.go
charithabandi: blkPropTimer is what determines how quickly the node can catchup to the current consensus round (as this broadcasts the current block proposals), if this is pretty high, I observed that the block times are almost proportional to this.

charithabandi: I will probably make this into a config

jchappelow (approved):
utACK.

jchappelow (approved):

PR #1169, opened 2024-12-17
fix cpu spikes in the eventloop

charithabandi:


jchappelow (approved):
OK.  Even if it were an empty block in `default:` it would still eat all CPU on one core just iterating as fast as possible. Profile shows `goselect` hogging it all.

charithabandi:
that makes sense. as it keeps executing by default in every loop :/

PR #1172, opened 2024-12-18
structured tx in block

jchappelow:
```
types: use structured tx in block, no error Tx srlzn

This updates the block structure with a `Txns []*Transaction` field
insead of `[][]byte`.

This also updates transaction serialization so that it is guaranteed
not to error (except if a writer fails). This means that it supports
a nil Body even if that is invalid in a block. The Bytes() method may
be used instead of MarshalBinary for this.

Finally, two bytes for version are added to transaction and transaction
results serializations.
```

### DIFF
--- a/node/block_processor/processor.go
+++ b/node/block_processor/processor.go
@@ -221,11 +221,7 @@ func (bp *BlockProcessor) Rollback(ctx context.Context, height int64, appHash kt
 }
 
 func (bp *BlockProcessor) CheckTx(ctx context.Context, tx *ktypes.Transaction, recheck bool) error {
-	rawTx, err := tx.MarshalBinary()
-	if err != nil {
-		return fmt.Errorf("invalid transaction: %v", err) // e.g. missing fields
-	}
-	txHash := types.HashBytes(rawTx)
+	txHash := tx.Hash()
 
 	// If the network is halted for migration, we reject all transactions.
 	if bp.chainCtx.NetworkParameters.MigrationStatus == ktypes.MigrationCompleted {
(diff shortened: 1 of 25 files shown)


jchappelow:
Ah heck, I think we should make marshaling and Hash() error-free.

node/block_processor/processor.go
```diff
@@ -367,31 +367,29 @@ func (bp *BlockProcessor) ExecuteBlock(ctx context.Context, req *ktypes.BlockExe
 
 	txResults := make([]ktypes.TxResult, len(req.Block.Txns))
 
-	bp.initBlockExecutionStatus(req.Block)
+	txHashes, err := bp.initBlockExecutionStatus(req.Block)
```
jchappelow, line 370: Whoops, I can revert this since there are no longer errors from Hash

node/block_processor/status.go
```diff
@@ -11,7 +11,7 @@ type blockExecStatus struct {
 	startTime, endTime time.Time
 	height             int64
 	txIDs              []ktypes.Hash
-	txStatus           map[string]bool
+	txStatus           map[ktypes.Hash]bool
```
jchappelow, line 14: One of the nice things about defining a `Hash` type as a fixed length array `[32]byte` is that it is `comparable` and can be a map key.  It's also nice to ensure we always encode to the same length hex chars, which we didn't do before (see old `display` tests where we had "hashes" displayed of just a few chars :exploding_head: )

charithabandi (approved):
This looks way better :)

PR #1176, opened 2024-12-20
pex / peer mgr reusability and seeder cmd

jchappelow:


jchappelow:
Ready for review, but I need to circle back on the `kwild seed` command and the `Seeder` types since a great deal had changed since I initially created and tested that.

node/types/consensus.go
```diff
@@ -60,10 +60,11 @@ func (cr *ConsensusReset) UnmarshalBinary(data []byte) error {
...
 type AckRes struct {
-	Height  int64
-	ACK     bool
-	BlkHash Hash
-	AppHash *Hash
+	Height    int64
+	ACK       bool
+	BlkHash   Hash
+	AppHash   *Hash
+	Signature []byte // signature of (n)ack || blkHash || appHash
```
charithabandi, line 67: This change would go away. (Implemented in the CommitSigs PR)

jchappelow: Whoops that wasn't supposed to be saved.  I think that was from when we were chatting.

charithabandi (approved):
I didn't understand the changes completely, but whatever I could, looks good.

jchappelow:
I'm not super happy with the `SecureTransport` and how it intermittently works with crawler although it appears flawless with regular nodes.  I'm not sure what the right way to do that is, so the nodes also support regular Noise security, with a chain ID check in the Notifiee.

PR #1177, opened 2024-12-20
Int tests

charithabandi:
Got the docker based env up and running with the dns resolved hostnames. 

 task dev:testnet:up should work

Started working on tests to deploy namespaces, tables and actions. Will get to reviving the rest of the tests in other PR, pushing these incase someone wants to start building tests for their use case.

### DIFF
--- a/test/integration/helper.go
+++ b/test/integration/helper.go
@@ -17,6 +17,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+
 	"slices"
 	"sort"
 	"strconv"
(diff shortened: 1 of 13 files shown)


test/integration/helper.go
```diff
@@ -566,7 +583,7 @@ func (r *IntHelper) RunDockerComposeWithServices(ctx context.Context, services [
 		if r.t.Failed() {
 			r.t.Logf("Stopping but keeping containers for inspection after failed test: %v", dc.Services())
 			cancel() // Stop, not Down, which would remove the containers too --- this doesn't work, dang
-			time.Sleep(5 * time.Second)
+			time.Sleep(5 * time.Minute)
```
jchappelow, line 586: I don't think we want CI to hang for 5 min on failure, which is what will happen now, I believe.

jchappelow (changes requested):
Looks good. Thanks for updating the integ composes.  `task dev:testnet:up` works as expected.  However in the the `r.t.Failed()` condition, I do not think we want a 5 minute sleep.  The containers can just be stopped because with a `Stop` of each container rather than `dc.Down` they won't be deleted, and we can still see the container logs, or even start them again manually.

test/integration/helper.go
charithabandi: Ah, I changed it for testing, forgot to change it back. Oops!

jchappelow (approved):

PR #1179, opened 2024-12-21
kwil-cli: fixed nil pointer dereference

brennanjl:
Fixes a nil pointer dereference bug I introduced while making changes to `kwil-cli`


brennanjl:
cc @KwilLuke this should fix your issue

KwilLuke (approved):

PR #1180, opened 2025-01-02
kwil-cli: minor improvements and bug fixes

brennanjl:
Fixes a small bug where not selecting a namespace does not default to the default namespace `main`.


jchappelow (approved):
lgtm

PR #1196, opened 2025-01-03
CommitSignatures support for verifying the correctness of the committ…

charithabandi:
…ed block

### DIFF
--- a/node/block.go
+++ b/node/block.go
@@ -263,7 +263,8 @@ func (n *Node) getBlk(ctx context.Context, blkHash types.Hash) (int64, []byte, *
 		n.log.Debug("Obtained content for block", "block", blkHash, "elapsed", time.Since(t0))
 
 		rd := bytes.NewReader(resp)
-		var height, ciLen, blkLen int64
+		var height int64
+		var ciLen, blkLen int32
 		binary.Read(rd, binary.LittleEndian, &height)
 		binary.Read(rd, binary.LittleEndian, &ciLen)
 		ciBts := make([]byte, ciLen)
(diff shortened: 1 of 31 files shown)


node/block.go
```diff
@@ -254,17 +262,30 @@ func (n *Node) getBlk(ctx context.Context, blkHash types.Hash) (int64, []byte, t
...
 
-		height := binary.LittleEndian.Uint64(resp[:8])
-		var appHash types.Hash
-		copy(appHash[:], resp[8:8+types.HashLen])
-		rawBlk := resp[8+types.HashLen:]
+		rd := bytes.NewReader(resp)
+		var height int64
+		var ciLen, blkLen int32
+		binary.Read(rd, binary.LittleEndian, &height)
+		binary.Read(rd, binary.LittleEndian, &ciLen)
```
jchappelow, line 269: The `if len(resp) < 8` check above this is only good enough for height.   We should give these `binary.Read`s an error return and check, probably beef up sanity checks before based on `len(resp)`

node/block.go
```diff
@@ -254,17 +262,30 @@ func (n *Node) getBlk(ctx context.Context, blkHash types.Hash) (int64, []byte, t
...
+		ciBts := make([]byte, ciLen)
+		io.ReadFull(rd, ciBts)
```
jchappelow, line 271: Also err check if there's not enough bytes.

node/types/interfaces.go
```diff
@@ -17,6 +17,7 @@ type BlockStore interface {
 	TxGetter
 	BlockResultsStorer
 
+	// TODO: Should this return commitInfo or AppHash?
```
jchappelow, line 20: Probably not.  If needed the caller can get by hash.

jchappelow (commented):
Nice!  Looks great.  Just probably need some length and error checks in the stream response readers.

jchappelow (approved):

PR #1202, opened 2025-01-07
setup init and autogen command support

charithabandi:
- Introduce GenesisAllocs back into the genesis config
- Added support for` kwild setup init` and `kwild --autogen` and `kwild setup genesis` commands
- Bunch of other concurrency fixes
- add the key type to the core/types Validator
- p2p config will now take listen_address rather than port and ip


node/node.go
```diff
@@ -763,7 +767,7 @@ func NewKey(r io.Reader) *crypto.Secp256k1PrivateKey {
 	return pk
 }
 
-func newHost(ip string, port uint64, chainID string, privKey crypto.PrivateKey, wl connmgr.ConnectionGater, logger log.Logger) (host.Host, error) {
+func newHost(ip string, port string, chainID string, privKey crypto.PrivateKey, wl connmgr.ConnectionGater, logger log.Logger) (host.Host, error) {
```
jchappelow, line 770: Really this is best as `port uint16`.  I think we should do `strconv.ParseUint()` after `net.SplitHostPort`

jchappelow (commented):
LGTM but I think we should not allow creating a multiaddr string with an invalid port.  See comment.

app/setup/genesis-hash.go
```diff
@@ -0,0 +1,269 @@
...
+kwil-admin setup genesis-hash --dbname kwild --host "127.0.0.1" --port "5432" --user kwild --genesis "~/.kwild/abci/config/genesis.json"
+
+# Compute the genesis hash from an existing PostgreSQL database using the root directory
+kwil-admin setup genesis-hash --genesis "~/.kwild/abci/config/genesis.json" --root-dir "~/.kwild"
+
+# Compute the genesis hash from a snapshot file
+kwil-admin setup genesis-hash --snapshot "/path/to/snapshot.sql.gz" --genesis "~/.kwild/abci/config/genesis.json"`
+)
+
+func GenesisHashCmd() *cobra.Command { //nolint:unused
```
jchappelow, line 44: No longer need `//nolint:unused`.

Also, please either git mv from _previous or git rm the stuff that's now migrated.

node/node.go
charithabandi: umm, makes sense

node/voting/voting.go
```diff
@@ -680,5 +762,55 @@ func (v *VoteStore) Rollback() {
...
+
+// DecodePubKey decodes a byte slice into a public key and key type.
+func DecodePubKey(b []byte) ([]byte, crypto.KeyType, error) {
+	if len(b) <= 4 {
+		return nil, 0, errors.New("insufficient data for public key")
+	}
+
+	keyType := crypto.KeyType(binary.LittleEndian.Uint32(b))
+	return b[4:], keyType, nil
+}
```
jchappelow, line 783: I think these functions should be unexported so this remains an implementation detail of storage serialization.

jchappelow (approved):

jchappelow:
Thx.  Please squash with a revised squashed commit msg, and merge when ready.

PR #1204, opened 2025-01-09
fixed gavins feedback

brennanjl:
This PR has three bug fixes:

- It fixes an issue where composite constraints cause our info queries to fail
- It fixes an issue where extensions don't have access to account and validator stores when they should (they were simply nil)
- It fixes an issue in the CLI when executing an action

It also includes one new addition to extensions: `OnStart`, which is like `Initialize`, but has access to the engine.


brennanjl:
@Yaiba

Yaiba (approved):

PR #1211, opened 2025-01-09
fixes the votestore init issue

charithabandi:
@Yaiba


Yaiba (approved):
just tested locally, works well

PR #1212, opened 2025-01-09
fix: engine incorrectly gives foreign keys to planner

brennanjl:
This fixes a bug with foreign keys that @Yaiba found


Yaiba (approved):
LGTM

PR #1213, opened 2025-01-10
mutable consensus parameters

jchappelow:
Based on top of https://github.com/kwilteam/kwil-db/pull/1222

This aims to make consensus parameters mutable and add the update and proposal code to support changing them.  The RPCs and CLI commands may be removed for product, but I needed them to test the internals e2e.

This needs more work for sure, but I'm putting it up as draft anyway since I'm not progressing quickly today.

Part of this branch includes changes related to key type, signers, authenticators, etc.

```
Usage:
  kwild consensus [command]

Available Commands:
  approve        Approve a consensus update proposal.
  list-proposals List all the pending consensus update proposals.
  params         Show active consensus parameters.
  propose        Submit a migration proposal.
  update-status  Pending consensus update proposal status.
```

```
$  ./kwild consensus params
Network Parameters:
	Leader: 0226b3ff29216dac187cea393f8af685ad419ac9644e55dce83d145c8b1af213bd [secp256k1]
	DB Owner: bababa
	Max Block Size: 6291456
	Join Expiry: 14400
	Vote Expiry: 108000
	Disabled Gas Costs: true
	Max Votes Per Tx: 200
	Migration Status: NoActiveMigration
```

```
$  ./kwild consensus propose -d "testchange" -u '{"max_block_size": 99999}'
Please review the following proposal:

Description: testchange

Updates: {"max_block_size":99999}

Do you want to submit this proposal? (y/n): y
TxHash: 6f56d7b67f8a4261e08a47b7a60092c90f47ce8665b4b6e810af120f1d284dac
```

```
$  ./kwild consensus params
Network Parameters:
	Leader: 0226b3ff29216dac187cea393f8af685ad419ac9644e55dce83d145c8b1af213bd [secp256k1]
	DB Owner: bababa
	Max Block Size: 99999
	Join Expiry: 14400
	Vote Expiry: 108000
	Disabled Gas Costs: true
	Max Votes Per Tx: 200
	Migration Status: NoActiveMigration
```

Also tested with db owner.  Just with a single leader and no validators, more to check...

We'll probably need update hooks as simple changing the network parameters struct is unlikely to be sufficient for every change.  e.g. I imagine db owner changing will involve DB query action.  We probably also do not want to permit migration status to be arbitrarily changed since we have other mechanisms for that.


app/node/build.go
```diff
@@ -139,12 +143,9 @@ func buildServer(ctx context.Context, d *coreDependencies) *server {
 		jsonRPCAdminServer.RegisterSvc(jsonAdminSvc)
 		jsonRPCAdminServer.RegisterSvc(jsonRPCTxSvc)
 		jsonRPCAdminServer.RegisterSvc(&funcsvc.Service{})
+		jsonRPCAdminServer.RegisterSvc(jsonChainSvc)
```
jchappelow, line 146: NOTE: @charithabandi user and admin servers now register all the "chain" service methods.
The `client` types now also include chain service client method.

app/params/cmd.go
```diff
@@ -0,0 +1,26 @@
...
+	Long:  "The `consensus` command provides functions for dealing with consensus update proposals.",
+}
+
+func NewConsensusCmd() *cobra.Command {
+	consensusCmd.AddCommand(
+		proposeUpdatesCmd(),
+		listUpdateProposalsCmd(),
+		approveUpdateProposalCmd(),
+		showUpdateProposalCmd(),
+		showConsensusParams(),
```
jchappelow, line 21: All of these `kwild consensus` commands are mainly for my testing.  We can hide or change entirely for beta release.

core/crypto/auth/auth.go
```diff
@@ -28,18 +28,18 @@ import "fmt"
...
+	// the signature is invalid, the method should return an error. If the
+	// signature is valid, the method should return nil.
+	Verify(compactID, msg, signature []byte) error
 
 	// Identifier returns a string identifier for a given sender.
 	// This string identifier is used to identify the sender when
 	// interacting with the Kuneiform engine, and will be used as
 	// the `@caller` variable in the engine.
-	Identifier(sender []byte) (string, error)
+	Identifier(compactID []byte) (string, error)
```
jchappelow, line 42: This method has the narrow documented purpose of getting `@caller` for engine based on a compact ID and key type.   It's a different type of authentication inside the engine, not signature based auth, so I think we should consider moving this method out of the `Authenticator` interface, somewhere closer to engine as something like `CallerFromIdentity(id []byte, authType AuthType)`.

core/crypto/auth/signer.go
```diff
@@ -129,15 +129,23 @@ type Signer interface {
...
+	// this Signer. This is used to set the Sender field of a Transaction.
+	// If you need the public key, use the PubKey method.
+	CompactID() []byte
+
+	// PubKey returns the public key associated with this signer. This may or
+	// may not correspond directly to the Identity, which may be an address. The
+	// purpose of Identity is to be used with the Verify method of the
+	// Authenticator for the AuthType of this Signer, which may not take the
+	// bytes of this pubkey.
+	PubKey() crypto.PublicKey
```
jchappelow, line 144: Can take this out of `Signer interface` as discussed, or make a more narrow Signer interface for client focused packages that won't be doing this.  Personally I think this being here helps underscore the possible distinction between key and "ID", which could be an eth address.

core/rpc/json/user/responses.go
```diff
@@ -85,6 +85,15 @@ type MigrationStatusResponse struct {
 	Status *types.MigrationState `json:"status"`
 }
 
+type ListPendingConsensusUpdatesResponse struct {
+	Proposals []*types.ConsensusParamUpdateProposal `json:"proposals"`
+}
+
+type ConsensusUpdateStatusResponse struct{}
+
+// 	Status *types.ConsensusUpdateProposalStatus `json:"status"`
```
jchappelow, line 94: TODO for me.  I didn't get around to implementing the proposal update resolution status query.

core/types/block.go
```diff
@@ -46,8 +46,9 @@ type Block struct {
 // CommitInfo includes the information about the commit of the block.
 // Such as the signatures of the validators aggreeing to the block.
 type CommitInfo struct {
-	AppHash Hash
-	Votes   []*VoteInfo
+	AppHash      Hash
+	Votes        []*VoteInfo
+	ParamUpdates ParamUpdates
```
jchappelow, line 51: Also as discussed, this does not need to be explicitly a field of `CommitInfo` since the update contents are absorbed into `AppHash`.  I think there's little wasted by repeating it here though, and it helps debug.

core/types/hex.go
```diff
@@ -32,6 +36,9 @@ func (hb *HexBytes) UnmarshalJSON(b []byte) error {
 
 // MarshalJSON satisfies the json.Marshaler interface.
 func (hb HexBytes) MarshalJSON() ([]byte, error) {
+	if hb == nil {
+		return []byte("null"), nil
+	}
```
jchappelow, line 41: NOTE: `nil` support for `HexBytes`.  This is not 100% required, but it gives faithful round trip (un)marshal if a field was `nil` rather than simply empty but non-nil.  However, in the Transaction marshalling, I ensure it is always marshaled one way for consistency.

core/types/messages.go
```diff
@@ -19,6 +19,7 @@ type BlockExecResult struct {
 	TxResults        []TxResult
 	AppHash          Hash
 	ValidatorUpdates []*Validator
+	ParamUpdates     ParamUpdates
```
jchappelow, line 22: Whoops, I'm not doing this.  Need to remove.

core/types/params.go
```diff
@@ -0,0 +1,566 @@
...
+		panic("ValidateUpdateTypes: incorrect parameter types defined")
+	}
+	if err := MergeUpdates(&np, paramMap); err != nil {
+		panic("MergeUpdates: incorrect parameter types defined")
+	}
+}
+
+type PublicKey struct {
+	crypto.PublicKey
+}
```
jchappelow, line 36: This adds consistent JSON marshal/unmarshal around a `crypto.PublicKey`

core/types/params.go
```diff
@@ -0,0 +1,566 @@
...
+	*pk = PublicKey{pubkey}
+	return nil
+}
+
+// NetworkParameters are network level configurations that can be
+// evolved over the lifetime of a network.
+type NetworkParameters struct {
+	// Leader is the leader's public key. The leader must be in the current
+	// validator set.
+	Leader PublicKey `json:"leader"`
```
jchappelow, line 91: Notable change here is that `Leader` type in genesis doc is a pubkey (which includes type in marshaling as above).

The other notable decision here is that I've made `null` invalid.   If there's a use case for creating a genesis file with no leader set, this support can be added (it's commented out above).

core/types/params.go
```diff
@@ -0,0 +1,566 @@
...
+// use a map to make updates easy to specify while keeping the NetworkParameters
+// struct simple and easy to use without nil checks.
+//
+// This approach also makes update serialization more compact, only encoding
+// data for updated fields.
+//
+// This approach however requires the definition of parameter names (ParamName)
+// and code to assert type of the values. The parameter names are enumerated
+// below, and their values are set using reflection during package initialization.
+type ParamUpdates map[ParamName]any
```
jchappelow, line 128: Please note the docs on this type.  There were some big decision points, and decided to go this route.  In someways it feels more complex under the hood, but is far less fuss in use of network parameters and merging updates.

core/types/payloads.go
```diff
@@ -216,6 +216,69 @@ type EncodedValue struct {
 	Data [][]byte `rlp:"optional"`
 }
 
+const evVersion = 0
+
+func (e EncodedValue) MarshalBinary() ([]byte, error) {
```
jchappelow, line 221: This is the bit I said I need to revert.  I do want to eliminate RLP as much as possible, but this will create some work on kwil-js if this is changed as done here.

core/types/types.go
```diff
@@ -94,9 +132,33 @@ type Migration struct {
...
+	ID          UUID         `json:"id"`
+	Description string       `json:"description"`
+	Updates     ParamUpdates `json:"updates"`
+}
+
 // MigrationStatus represents the status of the nodes in the zero downtime migration process.
 type MigrationStatus string
 
+func (ms MigrationStatus) NoneActive() bool {
+	return ms == NoActiveMigration || ms == ""
```
jchappelow, line 145: There is was some funny checking for empty string in the code.  Really this should be a panic or something.  We should audit wherever this could be left empty and add validation code.

node/block_processor/interfaces.go
```diff
@@ -36,7 +36,7 @@ type ValidatorModule interface {
 type TxApp interface {
 	Begin(ctx context.Context, height int64) error
 	Execute(ctx *common.TxContext, db sql.DB, tx *ktypes.Transaction) *txapp.TxResponse
-	Finalize(ctx context.Context, db sql.DB, block *common.BlockContext) (finalValidators []*ktypes.Validator, err error)
+	Finalize(ctx context.Context, db sql.DB, block *common.BlockContext) error
```
jchappelow, line 39: This removed return was a relic.  In a previous change @charithabandi got the validator from a method in later consumer code.

node/block_processor/processor.go
```diff
@@ -352,17 +351,8 @@ func (bp *BlockProcessor) ExecuteBlock(ctx context.Context, req *ktypes.BlockExe
...
-		MaxBlockSize:     bp.chainCtx.NetworkParameters.MaxBlockSize,
-		JoinExpiry:       bp.chainCtx.NetworkParameters.JoinExpiry,
-		VoteExpiry:       bp.chainCtx.NetworkParameters.VoteExpiry,
-		DisabledGasCosts: bp.chainCtx.NetworkParameters.DisabledGasCosts,
-		MaxVotesPerTx:    bp.chainCtx.NetworkParameters.MaxVotesPerTx,
-		MigrationStatus:  bp.chainCtx.NetworkParameters.MigrationStatus,
-	}
-	oldNetworkParams := *networkParams
+	inMigration := bp.chainCtx.NetworkParameters.MigrationStatus == ktypes.MigrationInProgress
+	haltNetwork := bp.chainCtx.NetworkParameters.MigrationStatus == ktypes.MigrationCompleted
```
jchappelow, line 355: This is not new code, but it is notable in that it modifies the `MigrationStatus` during block execution rather than after in commit.  I think this is known, but wanted to call it out in case it should be handled in the `ParamUpdates` instead.

node/block_processor/transactions.go
```diff
@@ -144,7 +144,7 @@ func (bp *BlockProcessor) prepareBlockTransactions(ctx context.Context, txs []*t
 			}
 		}
 
-		if bytes.Equal(tx.Sender, bp.signer.Identity()) {
+		if bytes.Equal(tx.Sender, bp.signer.CompactID()) && tx.Signature.Type == bp.signer.AuthType() {
```
jchappelow, line 147: This is actually incorrect.  I just discussed a related change with @charithabandi , which is a function that converts from authType (what is in `tx.Signature.Type`) to `crypto.KeyType`.  Recall that there are three authenticators presently, and two of them use the same key type (secp256k1) but the auth type is different to signal different message/tx sig message serialization (direct, or text message for personal sign)

node/consensus/engine.go
```diff
@@ -812,9 +796,14 @@ func (ce *ConsensusEngine) Role() types.Role {
 }
 
 func (ce *ConsensusEngine) updateNetworkHeight(height int64) {
+	// maybe: updateNetworkHeight(from, to int64) bool // require "from" is current or reject update?
```
jchappelow, line 799: After discussion with @charithabandi , we are removing this `ce.networkHeight`, which had a different intent that is now not used.

node/consensus/updates.go
```diff
@@ -0,0 +1,108 @@
...
+		app.Service.Logger.Info("Applying param updates", "description", pud.Description, "paramUpdates", pud.ParamUpdates)
+
+		// block.ChainContext.NetworkUpdates <= pud.ParamUpdates
+		if block.ChainContext.NetworkUpdates == nil {
+			block.ChainContext.NetworkUpdates = make(types.ParamUpdates, len(pud.ParamUpdates))
+		}
+		block.ChainContext.NetworkUpdates.Merge(pud.ParamUpdates)
+		return nil
+	},
+}
```
jchappelow, line 57: A couple important decisions / questions here

1. What is the "right" way presently to handle error returns from as `ResolveFunc`?  Should it never return errors?  Does the caller just log and not crash the node?  Can/should the `ResolutionConfig` include a `BodyValidation` function?  In the caller we have `if the resolveFunc fails, we should still continue on, since it simply means` documented, but I want to confirm and consider documenting this in the types.
2. While this code does the merge of updates into the active parameters, it's really not so simple for every parameter.  e.g. changing DB owner needs some other actions almost certainly.  MigrationStatus, while mutable and grouped with the other mutable params, should not be changed via this mechanism, right?  Leader can be changed in this updates mechanism, but like DB owner, there would have to be other actions or at least the change detected by CE.

node/meta/meta.go
```diff
@@ -5,61 +5,62 @@ package meta
...
 	getChainState = `SELECT height, app_hash, dirty FROM ` + chainSchemaName + `.chain;`
 
-	upsertParam = `INSERT INTO ` + chainSchemaName + `.consensus_params ` +
-		`VALUES ($1, $2) ` +
-		`ON CONFLICT (param_name) DO UPDATE SET param_value = $2;`
+	// network parameters table (TODO: combine with chain table)
+
+	initParamsTable = `CREATE TABLE IF NOT EXISTS ` + chainSchemaName + `.params (
+		params BYTEA
+	);` // no primary key, only one row
```
jchappelow, line 42: We can re-fancy this if needed.  It just very straightforward to have it store the binary marshaled params.  It is also extremely cheap just to write them at the end of each block and not try to compute "diffs", which is error prone and requires maintenance.

node/node.go
```diff
@@ -469,8 +469,10 @@ func (n *Node) Start(ctx context.Context, bootpeers ...string) error {
 		defer cancel()
 
 		broadcastFns := consensus.BroadcastFns{
-			ProposalBroadcaster:     n.announceBlkProp,
-			TxAnnouncer:             n.announceTx,
+			ProposalBroadcaster: n.announceBlkProp,
+			TxAnnouncer: func(ctx context.Context, txHash types.Hash, rawTx []byte) {
+				n.announceTx(ctx, txHash, rawTx, n.host.ID())
```
jchappelow, line 474: NOTE: @charithabandi taking an approach with a closure that captures the host's p2p ID rather than having CE/BP be aware.

jchappelow (commented):
Notes and comments mainly for @charithabandi and @brennanjl , but interested in thoughts from all.

jchappelow:
Ready for review from @charithabandi , although after a day of testing and working out bugs, there are a few TODOs on this PR I intend to resolve.

node/consensus/updates.go
charithabandi: 1. As Resolutions can be defined by the clients and can have impact on the consensus, we didn't want these to stop the node unless there is some non-determinism. So this approach of not returning any errors back to TxApp. 

2. Huh, I think we can have all required modules subscribe to the consensus parameters and let them know the final params at the end of every commit and the modules can take appropriate actions based on the updated state.

cmd/kwil-cli/cmds/account/balance.go
```diff
@@ -35,7 +35,7 @@ func balanceCmd() *cobra.Command {
 						return display.PrintErr(cmd, errors.New("no account ID provided and no signer set"))
 					}
 
-					ident, err := auth.Secp25k1Authenticator{}.Identifier(cl.Signer().Identity())
+					ident, err := auth.Secp25k1Authenticator{}.Identifier(cl.Signer().CompactID())
```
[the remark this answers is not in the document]
charithabandi: I am gonna make the accounts changes today, will probably take the similar approach as Votestore.

core/types/types.go
charithabandi: Ah Geez, agreed.

node/consensus/updates.go
jchappelow: > 2\. Huh, I think we can have all required modules subscribe to the consensus parameters and let them know the final params at the end of every commit and the modules can take appropriate actions based on the updated state.

This feels right to me too.  The full params can go to subscribers and they can decide what to do with it.

Making this comment thread into an issue so we can get this PR merged.

app/params/cmd.go
charithabandi: These are pretty useful, lets keep them

charithabandi (approved):

PR #1214, opened 2025-01-10
Engine testing

brennanjl:
This PR starts some initial work on better testing of the info schema. It also fixes a bug that @Yaiba is blocked on, where numeric precision and scale are not being handled properly.

I will finish the tests in a subsequent PR


Yaiba (approved):
LGTM

PR #1217, opened 2025-01-13
fix(extension): missing export method ReturnColumns

Yaiba:
fix missing `ReturnColumns` on exported method


brennanjl:
This is resolved in https://github.com/kwilteam/kwil-db/pull/1221

PR #1226, opened 2025-01-14
Consensus Engine should have upto-date validator set upon restart

charithabandi:


jchappelow (approved):
Thanks for pulling out these fixes

PR #1228, opened 2025-01-14
planner: fix pointer dereference

brennanjl:
This fixes a pointer de-reference bug that was caused when referencing an invalid column within an insert.

It also fixes a separate minor issue where a typecast for a subquery was not being handled correctly.


jchappelow (approved):
LGTM, but maybe @Yaiba knows the issue better.

Yaiba (approved):
LGTM

PR #1230, opened 2025-01-15
Make account store and the VoteStore aware of the KeyType

charithabandi:


jchappelow:
The main consideration when deciding on a key type vs. auth type in a struct is -- is it necessary to generate the formatted string for `@caller`.

The ID + *key* type should be sufficient for both user uniqueness purposes, and thus the account store (we wouldn't really want different balances for the same actual typed key depending on the signing method (direct vs personal sign), right?

It seems that the main places where auth type (rather than key type) are important: (1) transaction sig verification where it says how to serialize the signed message, and (2) generating a string identifier such as the checksumed eth address that is nice for user-facing things.

charithabandi:
@brennanjl @jchappelow PR is ready for review.

app/migration/proposal_status.go
```diff
@@ -53,9 +53,9 @@ func proposalStatusCmd() *cobra.Command {
 
 type MigrationStatus struct {
 	ProposalID *types.UUID
-	ExpiresAt  int64            `json:"expires_at"` // ExpiresAt is the block height at which the migration proposal expires
-	Board      []types.HexBytes `json:"board"`      // Board is the list of validators who are eligible to vote on the migration proposal
-	Approved   []bool           `json:"approved"`   // Approved is the list of bools indicating if the corresponding validator approved the migration proposal
+	ExpiresAt  int64           `json:"expires_at"` // ExpiresAt is the block height at which the migration proposal expires
+	Board      []types.NodeKey `json:"board"`      // Board is the list of validators who are eligible to vote on the migration proposal
```
brennanjl, line 57: idek if/why this matters, but should these be pointers?

core/types/types.go
```diff
@@ -3,18 +3,90 @@ package types
...
 
 // TODO: doc it all
 
+// AccountID uniquely identifies kwil accounts.
+// It is a combination of the Identifier and the key type.
+// Identifier can be a public key or an address.
+type AccountID struct {
+	Identifier HexBytes       `json:"identifier"`
+	KeyType    crypto.KeyType `json:"key_type"`
+}
```
brennanjl, line 24: Im still confused why we need an AccountID and NodeKey. They are literally the same thing, and can totally be used interchangeably.

brennanjl: Additionally, I noticed that KeyType is an int32. Do you think we should make them `string`? Normally I wouldn't care, but there are two reasons why I thought it might matter:

1. This data will be sent across JSON RPC, and might be read by curl / from the raw JSON body. I know this has been a concern before with how we send data over RPC.
2. 
3. Since users will be creating their own authenticators, we run a risk of colliding with theirs if we decide to add more. For example, we have:

```go
const (
	KeyTypeSecp256k1 KeyType = 0
	KeyTypeEd25519   KeyType = 1
	// KeyTypeRSA       KeyType = 2
)
```

IDOS presumably would have KeyTypeNEP413 as KeyType 2. If we decide to add RSA in the future, they will have to break all of their transactions, and bump their KeyType to 3.

CC @jchappelow

brennanjl (commented):
Still reviewing more but wanted to get a convo started with these

app/migration/proposal_status.go
charithabandi: Ah my bad, these should be pointers. will change it

core/types/payloads.go
```diff
@@ -549,10 +549,13 @@ func EncodeValue(v any) (*EncodedValue, error) {
...
 
+// transfer payload version
+const tVersion = 0
+
 // Transfer transfers an amount of tokens from the sender to the receiver.
 type Transfer struct {
-	To     string `json:"to"`     // to be string as user identifier
-	Amount string `json:"amount"` // big.Int
+	To     *AccountID `json:"to"`     // to be string as user identifier
+	Amount string     `json:"amount"` // big.Int
```
jchappelow, line 558: I was considering making this a `*big.Int` since we have `ReadBigInt` and `WriteBigInt` to make that simpler.  We'd also get validation on unmarshal.
Would changing this field create inconvenient conversion elsewhere in the code?

core/types/types.go
jchappelow: Regarding the JSON concerns, that's valid, and I would address that by implementing `MarshalJSON` and `UnmarshalJSON` for `KeyType`, or using a plain `string` for the field in a struct that needs to be marshalled.  However, Since we unmarshal into target struct rather than a `map[string]any`, I don't think we'd hit the headache with float64/string/json.Number.

We talked about custom registered key types, and @charithabandi has pushed the discussed changes.  I think this'll work out.  But as for the specific case of NEP413, that would still be `KeyTypeEd25519`.  Still your point about RSA and Schorr signatures is valid, and should be covered with what is now in this PR.

node/block_processor/processor.go
```diff
@@ -733,12 +747,16 @@ func txResultsHash(results []ktypes.TxResult) types.Hash {
 
 func (bp *BlockProcessor) accountsHash() types.Hash {
 	accounts := bp.accounts.Updates()
+
 	slices.SortFunc(accounts, func(a, b *ktypes.Account) int {
-		return strings.Compare(a.Identifier, b.Identifier)
+		id1 := fmt.Sprintf("%s:%s", hex.EncodeToString(a.ID.Identifier), a.ID.KeyType.String())
+		id2 := fmt.Sprintf("%s:%s", hex.EncodeToString(b.ID.Identifier), b.ID.KeyType.String())
+		return strings.Compare(id1, id2)
```
jchappelow, line 754: This is fine, but the compare func can be written with more logic like:

```go
	slices.SortFunc(accounts, func(a, b *ktypes.Account) int {
		if idCmp := bytes.Compare(a.ID.Identifier, b.ID.Identifier); idCmp != 0 {
			return idCmp
		}
		return cmp.Compare(a.ID.KeyType, b.ID.KeyType)
	})
```

the `fmt.Sprintf` is a relatively computationally expensive approach.

node/block_processor/processor.go
```diff
@@ -765,14 +783,14 @@ func validatorUpdatesHash(updates map[string]*ktypes.Validator) (types.Hash, []*
 		updatesList = append(updatesList, v)
 	}
 	slices.SortFunc(updatesList, func(a, b *ktypes.Validator) int {
-		return bytes.Compare(a.PubKey, b.PubKey)
+		return bytes.Compare(a.Identifier, b.Identifier)
```
jchappelow, line 786: Just the bytes is good enough for this compare?

core/types/payloads.go
charithabandi: probably not. now that we are not using rlp encoding, we can switch to big.Int

jchappelow (approved):

PR #1231, opened 2025-01-15
remove rlp from core tx payloads

jchappelow:
This change removes RLP from the `core` transaction payloads and the entire module.  The affected types are: `ActionCall` (not actually a payload though), `ActionExecution`, `RawStatement`, `EncodedValue`, and `Transfer`.  The serializations are versioned and documented clearly in `core/types/payloads.go`.  For instance:

```go
// RawStatement serialization is as follows (using SerializationByteOrder in all
// cases):
//
//   - Two bytes for version (uint16), which is presently 0 (rsVersion).
//   - The statement string is written according to WriteString, which has a
//	   4 byte (uint32) length prefix followed by the bytes of the utf8 string.
//   - The number of parameters is written as a uint16.
//   - For each parameter:
//     - The parameter name is written according to WriteString.
//     - The EncodedValue is serialized according to its MarshalBinary,
//       written according to WriteBytes.
```

This also modifies the `core/types/serialize` so that the default `Codec` is one that detects if the type implements `encoding.BinaryMarshaler` and `encoding.BinaryUnmarshal` (from the standard library).  Parts of node internals including `extensions/resolutions/credit/credit.go` and `node/migrations` still use the RLP codec. The RLP codec is moved into `node/rlp`, where it may be used via the `EncodingTypeRLP` encoding type, for which a codec is registered in the `init()` function of the `node/rlp` package.  **This should be temporary.**

This should be enough to start updating kwil-js to learn the binary serializations for the user payloads at the top, but we should get to updating the internal node types too.

Finally, one inefficiency of the serialization approach we have applied in many places is using fixed size integer encodings (e.g. 4 little endian bytes for a uint32, etc.), but we can utilize `varint` encoding to save space when it is more common to have values less than 255, which is very often the case.


charithabandi (approved):

core/crypto/auth/eth_personal_sign_test.go
```diff
@@ -41,3 +47,106 @@ func Test_eip55ChecksumAddr(t *testing.T) {
...
+	authnr = GetAuthenticator(sig.Type)
+	if _, is := authnr.(EthSecp256k1Authenticator); !is {
+		t.Errorf("expected EthSecp256k1Authenticator, got %T", authnr)
+	}
+
+	err = authnr.Verify(signer.CompactID(), msg, sig.Data)
+	require.NoError(t, err)
+
+	// now pretend to be metamask
+	sig.Data[crypto.RecoveryIDOffset] += 27
```
jchappelow, line 82: @KwilMachine what I'd really like is a signature from metamask, if you have one (with the address or pubkey)

jchappelow: Nevermind, I got it with etherscan: https://etherscan.io/verifySig/265446

jchappelow: `TestEthSecp256k1AuthenticatorVerifyMetaMask` added with real signature from a web3 wallet.  It hits this recovery byte check condition, so I think the fix is the right one.

brennanjl (approved):
This seems good to me, @charithabandi any other considerations?

charithabandi (approved):
Looks good to me!

PR #1232, opened 2025-01-16
fixes to app, block store, ce

jchappelow:
Tons of fixes and improvements all around.  This is based on https://github.com/kwilteam/kwil-db/pull/1230

- cleanup and consistency of all kwild commands, including descriptions (e.g. consistent capitalization and sentence structure, `kwil-admin` => `kwild admin` etc.
- make the `block abort` command's `--block_height` flag into a required argument, as it is in fact required.  The default of 0 would never be applicable.
- the `app/rpc.AdminSvcClient` function that construct the admin RPC client now tries to find any config from kwild's loaded config (including config.toml) if `--rpcserver` has not been set.  The rationale here is that if it finds *local* kwild config then you are likely to want to talk to that unless the flag is set to override that.
- the `admin gen-auth-key` command will now automatically write the generated client certificate to the node's clients.pem file, which is essentially the list of authorized client TLS certificates if operating in mTLS mode.
- consistent toml config file name (config.toml rather than kwil.toml)
- Fix the node status RPC and the Node method that handles it.  Several fields were unset, which had the effect of leaving block time stamp set to the zero Time in a call's BlockContext.
- The block store in `node/store` has the `Best` method now returning the timestamp.  The implementation of `Best` is now less ridiculously inefficient by tracking the best block height and hash as fields rather than looping through the block index map.
- In CE (`node/consensus`), the `stateInfo` struct that supports external methods such as RPC calls now includes the `lastCommit`, which gives access the last `CommitInfo` as well as the `BlockHeader` of the latest committed block.
- The rpc server now calls service methods with a context that contains the `*Server` instance in under the `ServerCtx` context key.  The `MethodHandler` signature is designed with a `*Server` input parameter for this purpose, but it had never been used previously.

With these changes, it will serve as the bases for docs generated with `cmd/kwild/generate` and `cmd/kwil-cli/generate`.

### DIFF
--- a/node/block_processor/processor.go
+++ b/node/block_processor/processor.go
@@ -13,6 +13,7 @@ import (
 	"slices"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/kwilteam/kwil-db/common"
 	"github.com/kwilteam/kwil-db/config"
@@ -270,7 +271,7 @@ func (bp *BlockProcessor) Rollback(ctx context.Context, height int64, appHash kt
 	return nil
 }
 
-func (bp *BlockProcessor) CheckTx(ctx context.Context, tx *ktypes.Transaction, recheck bool) error {
+func (bp *BlockProcessor) CheckTx(ctx context.Context, tx *ktypes.Transaction, height int64, blockTime time.Time, recheck bool) error {
 	txHash := tx.Hash()
 
 	// If the network is halted for migration, we reject all transactions.
@@ -315,7 +316,8 @@ func (bp *BlockProcessor) CheckTx(ctx context.Context, tx *ktypes.Transaction, r
 		Ctx: ctx,
 		BlockContext: &common.BlockContext{
 			ChainContext: bp.chainCtx,
-			Height:       bp.height.Load() + 1,
+			Height:       height + 1,
+			Timestamp:    blockTime.Unix() + 1, // eh, one second later... ?
 			Proposer:     bp.chainCtx.NetworkParameters.Leader,
 		},
 		TxID:          hex.EncodeToString(txHash[:]),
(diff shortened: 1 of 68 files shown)


jchappelow:
This is based on https://github.com/kwilteam/kwil-db/pull/1230 (which should merge first).

jchappelow:
PR description and commit message updated.  This will provide a basis for the generated docs and the node operator docs.  Ready for review, but it is based on https://github.com/kwilteam/kwil-db/pull/1230, so that goes first and this will likely change a little.

node/block_processor/processor.go
```diff
@@ -315,9 +316,9 @@ func (bp *BlockProcessor) CheckTx(ctx context.Context, tx *ktypes.Transaction, r
 		Ctx: ctx,
 		BlockContext: &common.BlockContext{
 			ChainContext: bp.chainCtx,
-			Height:       bp.height.Load() + 1,
-			// Timestamp:    bp.blockTime,
-			Proposer: bp.chainCtx.NetworkParameters.Leader,
+			Height:       height + 1,
+			Timestamp:    blockTime.Unix() + 1, // eh, one second later... ?
```
jchappelow, line 320: I don't know what we're supposed to claim block timestamp is here.  I think it's relatively unimportant for the engine... perhaps just `time.Now()` thinking height+1 block could be mined at any time

charithabandi: Yeah, checkTx is totally not concerned with the Timestamp, we can either keep it lastCommittedBlockTime + 1 or even just time.Now(). No preference

charithabandi (approved):
okay, LGTM!

jchappelow:
Resolved conflicts and re-pushed @charithabandi .  Will you please re-approve?

charithabandi (approved):

PR #1233, opened 2025-01-16
engine: made precompile methods register as actions

brennanjl:
This PR makes a few interface changes for extensions requested by @yaiba. It also registers extension methods as actions.


Yaiba (approved):
LGTM

PR #1236, opened 2025-01-16
set dbOwner as the leader when run with --autogen

charithabandi:
`init`, `testnet` and `autogen` will take `--db-owner` flag to configure the Owner, if not set, it sets it to the leader.


jchappelow (dismissed):
My only comment was using the authenticator to get the dbowner string in the fallback case.  Everything else looks good.

brennanjl:
We still need some sort of flag that allows us to set the db owner during autogen:

```
kwild --autogen --db-owner 0xabc
```

```
$kwild --db-owner 0xabc
error: cannot set a db owner if not using the --autogen flag
```

PR #1245, opened 2025-01-17
fixes network params initialization and sanity checks the blockproces…

charithabandi:


test/integration/kwild_test.go
```diff
@@ -607,7 +612,7 @@ func TestKwildEthDepositOracleValidatorUpdates(t *testing.T) {
 	require.NoError(t, err)
 
 	var validators []*crypto.Secp256k1PrivateKey
-	for i := 0; i < 3; i++ {
+	for _ = range [3]int{} {
```
jchappelow, line 615: `for range 3`.  Range-over-int is available since 1.22 (supported by Kwil)

node/block_processor/processor.go
```diff
@@ -861,13 +850,22 @@ func (bp *BlockProcessor) ConsensusParams() *ktypes.NetworkParameters {
...
 	bp.mtx.RLock()
-	status := bp.chainCtx.NetworkParameters.MigrationStatus
+	networkParams := bp.chainCtx.NetworkParameters
 	bp.mtx.RUnlock()
 
+	if networkParams == nil {
+		return nil, fmt.Errorf("network parameters not found, node is still booting up, try again later")
+	}
+
+	status := networkParams.MigrationStatus
```
jchappelow, line 869: I get we would want to use `bp.ConsensusParams`, which clones it to avoid a race here if another goroutine modifies `MigrationStatus` concurrently.

jchappelow (commented):
Just a couple small nits, but LGTM

jchappelow (approved):

PR #1246, opened 2025-01-17
core: added scanning for results

brennanjl:
Sorry, I should've split this up into separate PRs. It does 3 things:
1. It adds the ability to scan into values when using the Go client.
2. It moves the `decimal` package into `types`. It also renames some of the exported functions related to decimals to make it clear that they are for decimals (e.g. `NewExplicit` -> `NewDecimalExplicit`). Previously, it was clear that these were for decimals because they were in the decimals package, but now that is not the case.
3. It renames references to the Kuneiform's `decimal` type to `numeric`, to make it match Postgres more closely. This was actually a change I made in the new engine a while ago., there were just a few straggling things I hadn't updated. `decimal` still works as an alias to `numeric`.

### DIFF
--- a/core/types/data_types.go
+++ b/core/types/data_types.go
@@ -7,8 +7,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-
-	"github.com/kwilteam/kwil-db/core/types/decimal"
 )
 
 // DataType is a data type.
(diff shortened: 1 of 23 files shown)


Yaiba (dismissed):
LGTM

Yaiba (approved):

PR #1250, opened 2025-01-20
added type aliases for encode

brennanjl:
In v0.10, we rename `blob` and `decimal` to be `bytea` and `numeric`. We still allow `blob` and `decimal`, they are simply aliased to `bytea` and `numeric`. We were not representing this aliasing in value decoding. This PR fixes that.


jchappelow (approved):
LGTM

PR #1251, opened 2025-01-20
restrict empty block production to 1 empty block per emptyBlockTimeout

charithabandi:
This PR changes the behavior of the block production by the leader:

Configurations:
- `propose_timeout`: used to specify the minimum duration to wait before proposing a new block with transactions. If set to 0, the behavior will simulate the `--automine` feature in the hardhat/ganache where the leader produces a block the moment the transactions become available. 
- `empty_block_timeout`: The duration to wait before proposing an empty block. 
Thus with this change the block intervals can range from propose_timeout to empty_block_timeout depending on the availability of the transactions in the mempool. 

Updates the resolution infra to now use expiry in terms of timestamp rather than number of blocks. 

Also makes some performance improvements to the engine to improve the blocksync performance.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -260,6 +260,7 @@ func DefaultConfig() *Config {
 		},
 		Consensus: ConsensusConfig{
 			ProposeTimeout:        Duration(1000 * time.Millisecond),
+			EmptyBlockTimeout:     Duration(1 * time.Minute),
 			BlockProposalInterval: Duration(1 * time.Second),
 			BlockAnnInterval:      Duration(3 * time.Second),
 		},
@@ -356,7 +357,9 @@ type DBConfig struct {
 }
 
 type ConsensusConfig struct {
-	ProposeTimeout Duration `toml:"propose_timeout" comment:"timeout for proposing a block (applies to leader)"`
+	ProposeTimeout Duration `toml:"propose_timeout" comment:"minimum time to wait before proposing a block with transactions (applies to leader). If set to 0, the leader will propose a block as soon as it has transactions"`
+
+	EmptyBlockTimeout Duration `toml:"empty_block_timeout" comment:"timeout for proposing an empty block"`
 	// reannounce intervals
 
 	// BlockProposalInterval is the interval between block proposal reannouncements by the leader.
--- a/config/config.go
+++ b/config/config.go
@@ -28,24 +28,6 @@ const (
 	AdminCertName       = "admin.cert"
 )
 
-// Duration is a wrapper around time.Duration that implements text
-// (un)marshalling for the go-toml package to work with Go duration strings
-// instead of integers.
-type Duration time.Duration
-
-func (d Duration) MarshalText() ([]byte, error) {
-	return []byte(time.Duration(d).String()), nil
-}
-
-func (d *Duration) UnmarshalText(text []byte) error {
-	duration, err := time.ParseDuration(string(text))
-	if err != nil {
-		return err
-	}
-	*d = Duration(duration)
-	return nil
-}
-
 type GenesisAlloc struct {
 	ID      KeyHexBytes `json:"id"`
 	KeyType string      `json:"key_type"`
@@ -239,7 +221,7 @@ func DefaultGenesisConfig() *GenesisConfig {
 			Leader:           types.PublicKey{},
 			DBOwner:          "",
 			MaxBlockSize:     6 * 1024 * 1024,
-			JoinExpiry:       14400,
+			JoinExpiry:       types.Duration(86400 * time.Second),
 			DisabledGasCosts: true,
 			MaxVotesPerTx:    200,
 			MigrationStatus:  types.NoActiveMigration,
@@ -259,10 +241,10 @@ func DefaultConfig() *Config {
 			BootNodes:     []string{},
 		},
 		Consensus: ConsensusConfig{
-			ProposeTimeout:        Duration(1000 * time.Millisecond),
-			EmptyBlockTimeout:     Duration(1 * time.Minute),
-			BlockProposalInterval: Duration(1 * time.Second),
-			BlockAnnInterval:      Duration(3 * time.Second),
+			ProposeTimeout:        types.Duration(1000 * time.Millisecond),
+			EmptyBlockTimeout:     types.Duration(1 * time.Minute),
+			BlockProposalInterval: types.Duration(1 * time.Second),
+			BlockAnnInterval:      types.Duration(3 * time.Second),
 		},
 		DB: DBConfig{
 			Host:          "127.0.0.1",
@@ -270,16 +252,16 @@ func DefaultConfig() *Config {
 			User:          "kwild",
 			Pass:          "",
 			DBName:        "kwild",
-			ReadTxTimeout: Duration(45 * time.Second),
+			ReadTxTimeout: types.Duration(45 * time.Second),
 			MaxConns:      60,
 		},
 		RPC: RPCConfig{
 			ListenAddress:      "0.0.0.0:8484",
-			BroadcastTxTimeout: Duration(15 * time.Second),
-			Timeout:            Duration(20 * time.Second),
+			BroadcastTxTimeout: types.Duration(15 * time.Second),
+			Timeout:            types.Duration(20 * time.Second),
 			MaxReqSize:         6_000_000,
 			Private:            false,
-			ChallengeExpiry:    Duration(30 * time.Second),
+			ChallengeExpiry:    types.Duration(30 * time.Second),
 			ChallengeRateLimit: 10,
 		},
 		Admin: AdminConfig{
@@ -297,7 +279,7 @@ func DefaultConfig() *Config {
 		},
 		StateSync: StateSyncConfig{
 			Enable:           false,
-			DiscoveryTimeout: Duration(30 * time.Second),
+			DiscoveryTimeout: types.Duration(30 * time.Second),
 			MaxRetries:       3,
 		},
 		Extensions: make(map[string]map[string]string),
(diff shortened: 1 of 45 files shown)


node/consensus/leader.go
```diff
@@ -32,8 +32,53 @@ import (
...
+			ce.newBlockProposal <- struct{}{}
+			return nil
+		case <-ticker.C:
+			// check for the availability of transactions in the mempool
+			if ce.mempool.TxsAvailable() {
+				ce.newBlockProposal <- struct{}{}
+				return nil
+			} else {
+				// check if the leader has any new events to broadcast a voteID transaction
+				hasEvents, err := ce.blockProcessor.HasEvents(ctx)
```
jchappelow, line 64: :/  a read DB transaction twice per second to check for events.  Does this apply to only leader?  Regardless shouldn't it be more straightforward to know if there are unbroadcasted events if we've been running?

charithabandi: Yeah this applies only to leader, but nonetheless, it shouldn't access db that much. I am looking for alternatives.

jchappelow: Twice a second is a bit much IMO.  We have designed kwild like a combination of microservices communicating through a DB, but it's monolithic and we are kinda not taking advantage of the benefits that this design can bring.  All of this is known without accessing postgres.

charithabandi: @jchappelow, haven't fixed that part yet. recent commit just involves resolution changes. 

I am thinking if we can track the numEvents in-memory somehow

jchappelow:
I accidentally commented on a specific commit hash and it's not showing in the conversation view.  I was confused about the `types.Duration` thing: https://github.com/kwilteam/kwil-db/commit/c66f76b3ed5818582ceddd5221a596c24b347a49#r151594760
I think it only applies to toml.

charithabandi:
Yeah, my bad. I wanted the json to marshal in "24h" format rather than 86400, but kinda not needed.

jchappelow:
> Yeah, my bad. I wanted the json to marshal in "24h" format rather than 86400, but kinda not needed.

Oh, but for json you can use `time.Duration`.  The extra MarshalText / UnmarshalText methods were just for go-toml.

jchappelow:
> > Yeah, my bad. I wanted the json to marshal in "24h" format rather than 86400, but kinda not needed.
> 
> Oh, but for json you can use `time.Duration`. The extra MarshalText / UnmarshalText methods of `config.Duration` were just for go-toml.

OMG, I'm wrong.  https://github.com/golang/go/issues/65501

node/block_processor/processor.go
```diff
@@ -418,6 +418,8 @@ func (bp *BlockProcessor) ExecuteBlock(ctx context.Context, req *ktypes.BlockExe
 	inMigration := bp.chainCtx.NetworkParameters.MigrationStatus == ktypes.MigrationInProgress
 	haltNetwork := bp.chainCtx.NetworkParameters.MigrationStatus == ktypes.MigrationCompleted
 
+	isLeader := bytes.Equal(bp.signer.CompactID(), req.Proposer.Bytes())
```
jchappelow, line 421: I think we should use `bp.signer.PubKey().Equals(req.Proposer)`

config/config.go
```diff
@@ -298,37 +281,39 @@ type DBConfig struct {
...
+	DBName        string         `toml:"dbname" comment:"postgres database name"`
+	ReadTxTimeout types.Duration `toml:"read_timeout" comment:"timeout on read transactions from user RPC calls and queries"`
+	MaxConns      uint32         `toml:"max_connections" comment:"maximum number of DB connections to permit"`
 }
 
 type ConsensusConfig struct {
-	ProposeTimeout Duration `toml:"propose_timeout" comment:"timeout for proposing a block (applies to leader)"`
+	ProposeTimeout types.Duration `toml:"propose_timeout" comment:"minimum time to wait before proposing a block with transactions (applies to leader). If set to 0, the leader will propose a block as soon as it has transactions"`
+
+	EmptyBlockTimeout types.Duration `toml:"empty_block_timeout" comment:"timeout for proposing an empty block"`
```
jchappelow, line 296: I think we also need to explain the interactions.  There are various combinations, some of which may be invalid or special cases

| propose timeout zero | empty block timeoute zero  | behavior Description                 |
|-----------------------|-----------------------|---------------------------------------|
| :heavy_check_mark:                      | :heavy_check_mark:                      | Automine, no empty blocks ever. |
| :heavy_check_mark:                      | :x:                      | Automine, with empty blocks after empty timeout.  |
| :x:                      | :heavy_check_mark:                      | No automine, no empty blocks.  |
| :x:                      | :x:                      | No automine, empty blocks after empty timeout. |

Conceptually, does the above seems like an intuitive system?  Does this PR accomplish all those cases?  Can we concisely document?

CometBFT reference, which is somewhat difficult to understand: https://docs.cometbft.com/v1.0/references/config/config.toml#consensuscreate_empty_blocks_interval

jchappelow (changes requested):

config/config.go
charithabandi: umm, I did not add support to "no empty blocks(emptyBlockTimeout = 0)" scenario, is this something we want?

node/consensus/leader.go
```diff
@@ -32,8 +32,38 @@ import (
 // the current block, revert any state changes made, and remove the problematic transactions from the mempool before
 // reproposing the block.
 
-// startNewRound initiates a new round of the consensus process (Prepare Phase).
-func (ce *ConsensusEngine) startNewRound(ctx context.Context) error {
+func (ce *ConsensusEngine) newBlockRound(ctx context.Context) error {
+	ticker := time.NewTicker(ce.proposeTimeout)
+	emptyBlockTicker := time.NewTicker(ce.emptyBlockTimeout)
```
jchappelow, line 37: If either of these are zero, the loop below is going to fire like crazy.  This is the same pathological case we saw before with a loop with nothing slowing down iterations.  CPU will be 100% just looping and doing nothing.

charithabandi: Yea, just realized while thinking about emptyBlockTimeout = 0 scenario. fixing it

config/config.go
jchappelow: Probably not, I'm just saying that from a config perspective, if it is left as zero, then we need to set it or ignore it to get sensible behavior.  The entire matrix in that table needs to be considered.

node/consensus/leader.go
jchappelow: Right, related to this: https://github.com/kwilteam/kwil-db/pull/1251#discussion_r1924452421
Need to consider all combinations from a config perspective, and from a CE implementation perspective.

jchappelow: I'm wondering if a single ticker that is the minimum of the two non-zero timeouts, and the `case` code could do a `time.Since(lastProposeTime)` to compare with the empty block timeout when there are no transactions and the empty timeout is non-zero.  Whatever works cleanly though.

node/consensus/blocksync.go
```diff
@@ -106,67 +108,92 @@ func (ce *ConsensusEngine) discoverBestHeight(ctx context.Context) (int64, error
...
+	_, rawblk, ci, err := ce.getBlock(ctx, height)
+	if err != nil { // all kinds of errors?
+		ce.log.Info("Error requesting block from network", "height", height, "error", err)
+		return fmt.Errorf("error requesting block from network: height : %d, error: %w", height, err)
+	}
+
+	ce.state.mtx.Lock()
+	defer ce.state.mtx.Unlock()
+
+	if ce.state.lc.height != 0 && ci != nil && ci.AppHash.IsZero() {
```
jchappelow, line 185: What's the situation is this condition is describing?

node/consensus/blocksync.go
```diff
@@ -106,67 +108,92 @@ func (ce *ConsensusEngine) discoverBestHeight(ctx context.Context) (int64, error
...
+	return nil
+}
+
+// syncBlockWithRetry requests the network for the specified block and retries until the block
+// is received and applied to the db.
+func (ce *ConsensusEngine) syncBlockWithRetry(ctx context.Context, height int64) error {
+	_, rawblk, ci, err := ce.getBlock(ctx, height)
+	if err != nil { // all kinds of errors?
+		ce.log.Info("Error requesting block from network", "height", height, "error", err)
+		return fmt.Errorf("error requesting block from network: height : %d, error: %w", height, err)
```
jchappelow, line 179: The caller that receives this error will presumably log it too?

jchappelow (dismissed):

jchappelow:
Thanks for addressing that long mutex lock.

node/consensus/blocksync.go
jchappelow: In other words "only genesis should have a zero apphash in commit info"?  For some reason I find it hard to tease out what this is testing for.  Is it a normal case, or some strange state that "should not happen"?

charithabandi: some strange cases, I added this for crash protection due to incorrect initialization of the node, but that issue is fixed, so probably not needed. I will run some tests and check this out.

jchappelow (dismissed):

node/consensus/engine.go
```diff
@@ -722,6 +731,9 @@ func (ce *ConsensusEngine) doCatchup(ctx context.Context) error {
 
 	if err := ce.processCurrentBlock(ctx); err != nil {
 		ce.log.Error("error during block processing in catchup", "height", startHeight+1, "error", err)
+		if err == types.ErrBlkNotFound {
```
jchappelow, line 734: We'd want to use `errors.Is(err, types.ErrBlkNotFound)` since the error would generally be wrapped.  Also would you comment on how this directs the flow in the caller?

node/consensus/engine.go
```diff
@@ -451,6 +458,8 @@ func (ce *ConsensusEngine) runConsensusEventLoop(ctx context.Context) error {
 
 		case m := <-ce.msgChan:
 			ce.handleConsensusMessages(ctx, m)
+			// reset the ticker as we just processed consensus messages
+			catchUpTicker.Reset(catchupDur)
```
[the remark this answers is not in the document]
charithabandi: Yeah, there are only 2 types of messages blkProp and blkAnn, which we would need to reset the ticker after.

node/consensus/engine.go
charithabandi: Yeah, will do that

jchappelow (approved):

jchappelow (approved):

jchappelow:
Good to go.  Please squash merge with a revised commit message.

PR #1260, opened 2025-01-21
Cli act tests

brennanjl:
Still some work to do on this but putting it up for a discussion


config/config.go
```diff
@@ -45,9 +46,36 @@ func (d *Duration) UnmarshalText(text []byte) error {
...
-	ID      string   `json:"id"`
-	KeyType string   `json:"key_type"`
-	Amount  *big.Int `json:"amount"`
+	ID      KeyHexBytes `json:"id"`
+	KeyType string      `json:"key_type"`
+	Amount  *big.Int    `json:"amount"`
+}
+
+// KeyHexBytes wraps hex bytes, and allows it to receive Ethereum 0x addresses
+type KeyHexBytes types.HexBytes
```
brennanjl, line 55: Looking for feedback here

config/config.go
```diff
@@ -45,9 +46,36 @@ func (d *Duration) UnmarshalText(text []byte) error {
...
-	ID      string   `json:"id"`
-	KeyType string   `json:"key_type"`
-	Amount  *big.Int `json:"amount"`
+	ID      KeyHexBytes `json:"id"`
+	KeyType string      `json:"key_type"`
+	Amount  *big.Int    `json:"amount"`
+}
+
+// KeyHexBytes wraps hex bytes, and allows it to receive Ethereum 0x addresses
+type KeyHexBytes struct{ types.HexBytes }
```
brennanjl, line 55: Looking for some feedback here.

brennanjl:
Ok this is ready to go. There are still some tests we need with roundtripping all data types, kgw, etc., but for a beta this should be enough.

config/config.go
charithabandi: Didn't we decide for the genesis allocs to use compactID bytes (hex) rather than the Identifier, as we started identifying accounts with that. 
@jchappelow

jchappelow: I thought so.  These also can certainly go to validators, who didn't have an eth addr

jchappelow: We can probably sort this out in https://github.com/kwilteam/kwil-db/pull/1254 unless these changes are needed for these acceptance tests.

config/config.go
```diff
@@ -45,9 +46,36 @@ func (d *Duration) UnmarshalText(text []byte) error {
...
-	Amount  *big.Int `json:"amount"`
+	ID      KeyHexBytes `json:"id"`
+	KeyType string      `json:"key_type"`
+	Amount  *big.Int    `json:"amount"`
+}
+
+// KeyHexBytes wraps hex bytes, and allows it to receive Ethereum 0x addresses
+type KeyHexBytes types.HexBytes
+
+func (l *KeyHexBytes) UnmarshalJSON(b []byte) error {
```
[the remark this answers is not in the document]
jchappelow: It just occurred to me that one potentially problematic quirk of this approach is that we cannot round trip (write out) a genesis.json file with the 0x or case.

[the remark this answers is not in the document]
brennanjl: Why couldn't we just have a `MarshalJSON` method that adds an 0x if the key type is secp and the id is 20 bytes?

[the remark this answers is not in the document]
brennanjl: Hmm I guess this would need to be a special Marshal (with an Unmarshal) on the `GenesisAlloc` type

config/config.go
jchappelow: Not a bad idea in https://github.com/kwilteam/kwil-db/pull/1260#discussion_r1923935817.  I can resolve any conflicts in the other PR if you wanna give that a shot.

brennanjl: Sure ill try it

brennanjl: ok just took a try

core/types/results.go
```diff
@@ -289,10 +301,14 @@ func convPtrArr[T any](src []any, dst *[]*T) error {
 
 		s := new(T)
 
-		err := convertScalar(val, s)
+		wrote, err := convertScalar(val, s)
 		if err != nil {
 			return err
 		}
+		if !wrote {
+			// if it didnt write a value, do not add the empty pointer to the slice
```
jchappelow, line 309: This means "leave the nil pointer in the slice" or "do not insert a non-nil pointer to the zero value for T"?  I didn't know what an "empty pointer" was describing at first.

jchappelow (dismissed):
LGTM.  @Yaiba probably should look though.

Yaiba:
> LGTM. @Yaiba probably should look though.

will take a quick scan

core/types/results.go
brennanjl: I think thats a bad comment, will fix.

It should say: "if convertScalar did not write to val, we should not add the new pointer to the slice"

cmd/kwil-cli/cmds/exec-action.go
```diff
@@ -0,0 +1,389 @@
...
+	clientType "github.com/kwilteam/kwil-db/core/client/types"
+	"github.com/kwilteam/kwil-db/core/types"
+	"github.com/kwilteam/kwil-db/node/engine/interpreter"
+	"github.com/spf13/cobra"
+)
+
+var (
+	execActionLong = `TODO: fill me out`
+
+	execActionExample = `TODO: fill me out`
```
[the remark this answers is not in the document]
brennanjl: So @Yaiba pointed out that our old version of database execute simply doesn't work with some of the new things supported within the engine. Namely, it requires all named parameters, rather than positional (which doesn't work for a lot of the things Gavin is doing with extension methods).

Furthermore, the old execute got pretty overloaded. Now that we have ad-hoc INSERT/UPDATE/DELETE, it had to also handle that (at least that's what I felt was most clear, since `execute` is definitely what I would expect to use to do that).

So I sort've took this as a chance to fix that. Now, it is clear what to use for actions and SQL. It can still use named params if you want, but positional is the default. The same command that handles executing an action once also handles it with CSV values.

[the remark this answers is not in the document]
brennanjl: I have another commit coming that adds docs for this, as well as a few other things. Since this is already reviewed, we can merge this and then Ill add the other later, since it has a lot of stuff (found a lot of bugs/edge cases)

cmd/kwil-cli/cmds/params.go
```diff
@@ -0,0 +1,355 @@
...
+)
+
+// parseParams parses the arguments into a map of name to value.
+// it expects the arguments to be in the form of name:type=value
+func parseParams(args []string) (map[string]any, error) {
+	params := make(map[string]any)
+
+	for _, arg := range args {
+		// trim off any leading '$'
+		arg = strings.TrimPrefix(arg, "$")
```
Yaiba, line 19: Just to point it out that this is unlikely to happen, if user pass `$xx`(without single quote) to the cli, it'll be expanded by SHELL.

cmd/kwil-cli/cmds/params_test.go
```diff
@@ -0,0 +1,184 @@
...
+		input   string
+		name    string
+		want    any
+		wantErr bool
+	}
+
+	tests := []testcase{
+		// scalars
+		{"username:text=satoshi", "username", "satoshi", false},
+		{`username:text="satoshi"`, "username", "satoshi", false},
```
Yaiba, line 23: I think we should only support single quote, to avoid the confusion if user want to insert a text like `$xxx`, because the SHELL will expand the value

node/engine/interpreter/planner.go
```diff
@@ -75,6 +85,28 @@ func makeActionToExecutable(namespace string, act *Action) *executable {
 			for _, stmt := range stmtFns {
 				err := stmt(exec2, func(row *row) error {
 					row.columns = returnColNames
+
+					// we will ensure that the return values match the expected return types
```
Yaiba, line 89: +1

Yaiba:
lgtm, only issue I have is parsing 'text' type in cli I mentioned at https://github.com/kwilteam/kwil-db/pull/1260#discussion_r1924493308

cmd/kwil-cli/cmds/params.go
brennanjl: They often times will need to wrap it in single quotes though. For example, if they are passing an array, they have to do:

```
'$param:int[]=[1,2]'
```

So IMO we should keep it

cmd/kwil-cli/cmds/params_test.go
brennanjl: This is still the case though. For example:

```
$ ab='hello!'; kwil-cli query 'select $a' -p a:text="$ab"
|        |
+--------+
| hello! |
```

The double quotes will only get parsed if a user wraps the entire param in them:

```
ab='hello!'; kwil-cli query 'select $a' -p 'a:text="$ab"'
|     |
+-----+
| $ab |
```

brennanjl:
I ended up including some other changes in this PR. It's all in the last commit, so if you would like me to pop it off into a separate PR so that it is easier to review, I'm happy to.

It doesn't include much new stuff except tests, bug fixes, and docs for the new CLI commands.

jchappelow (approved):
LGTM.  Send it

PR #1271, opened 2025-01-22
kwil-cli: fix generate-key command

brennanjl:
Fixes a bug in `kwil-cli utils generate-key` causing it to always fail.

### DIFF
--- a/cmd/kwil-cli/cmds/utils/generate_key.go
+++ b/cmd/kwil-cli/cmds/utils/generate_key.go
@@ -38,7 +38,7 @@ func generateKeyCmd() *cobra.Command {
 			pubKeyBts := pubKey.Bytes()
 
 			pubKeyHex := hex.EncodeToString(pubKeyBts)
-			address, err := auth.EthSecp256k1Authenticator{}.Identifier(pubKeyBts)
+			address, err := auth.EthSecp256k1Authenticator{}.Identifier(crypto.EthereumAddressFromPubKey(pubKey.(*crypto.Secp256k1PublicKey)))
 			if err != nil {
 				return display.PrintErr(cmd, err)
 			}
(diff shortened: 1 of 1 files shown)


cmd/kwil-cli/cmds/utils/generate_key.go
```diff
@@ -38,7 +38,7 @@ func generateKeyCmd() *cobra.Command {
 			pubKeyBts := pubKey.Bytes()
 
 			pubKeyHex := hex.EncodeToString(pubKeyBts)
-			address, err := auth.EthSecp256k1Authenticator{}.Identifier(pubKeyBts)
+			address, err := auth.EthSecp256k1Authenticator{}.Identifier(crypto.EthereumAddressFromPubKey(pubKey.(*crypto.Secp256k1PublicKey)))
```
jchappelow, line 41: oh.  This works, but this should be feeding `EthSecp256k1Authenticator{}.Identifier()` the output from a signer's CompactID method.

```
oh.  This works, but this should be feeding `EthSecp256k1Authenticator{}.Identifier()` the output from a signer's CompactID method.

```
// CompactID returns the identity of the signer (ETH address for this signer).
func (e *EthPersonalSigner) CompactID() []byte {
	pubKey := e.Key.Public().(*crypto.Secp256k1PublicKey)
	return crypto.EthereumAddressFromPubKey(pubKey)
}
```
```

jchappelow: `crypto.GetUserSigner(priv)` or just `&EthPersonalSigner{Key: *priv}` would be the consistent way to approach this rather than using the `EthereumAddressFromPubKey` func, although it clear does work.

brennanjl: Ok, I'll do that.

brennanjl:
Replacing this PR with https://github.com/kwilteam/kwil-db/pull/1272, which includes more CLI changes that I'm finding while writing docs.

The requested changes so far are included in the new PR.

Issue #1274, opened 2025-01-22
make blockHash available to common.BlockContext


Yaiba:
Specifically, common.EngineContext.TxContext.BlockContext.

The reward extension needs the block hash when generating the Merkle tree, and it's not available for now.

brennanjl:
See https://github.com/kwilteam/kwil-db/pull/1275

jchappelow:
Resolved by https://github.com/kwilteam/kwil-db/pull/1275

PR #1275, opened 2025-01-22
node: add block hash to BlockContext

brennanjl:
This PR adds block hash to BlockContext, as requested by @Yaiba.

It makes a minor refactor to pass in the genesis config to txapp. I felt this was appropriate because

1. We were pretty much just destructuring `config.GenesisConfig` into 5-6 args.
2. txapp already relied on the `config` package

I'm happy to change it back though, no strong opinion.

### DIFF
--- a/common/context.go
+++ b/common/context.go
@@ -33,6 +33,8 @@ type BlockContext struct {
 	ChainContext *ChainContext
 	// Height gets the height of the current block.
 	Height int64
+	// Hash is the hash of the current block.
+	Hash types.Hash
 	// Timestamp is a timestamp of the current block, in seconds (UNIX epoch).
 	// It is set by the block proposer, and therefore may not be accurate.
 	// It should not be used for time-sensitive operations where incorrect
--- a/common/context.go
+++ b/common/context.go
@@ -34,6 +34,8 @@ type BlockContext struct {
 	// Height gets the height of the current block.
 	Height int64
 	// Hash is the hash of the current block.
+	// It can be empty if the block is the genesis block and
+	// no initial state hash was specified in the genesis configuration.
 	Hash types.Hash
 	// Timestamp is a timestamp of the current block, in seconds (UNIX epoch).
 	// It is set by the block proposer, and therefore may not be accurate.
--- a/node/block_processor/interfaces.go
+++ b/node/block_processor/interfaces.go
@@ -5,6 +5,7 @@ import (
 	"math/big"
 
 	"github.com/kwilteam/kwil-db/common"
+	"github.com/kwilteam/kwil-db/config"
 	ktypes "github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/node/migrations"
 	"github.com/kwilteam/kwil-db/node/snapshotter"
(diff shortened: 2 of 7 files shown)


node/block_processor/interfaces.go
```diff
@@ -40,7 +41,7 @@ type TxApp interface {
 	Finalize(ctx context.Context, db sql.DB, block *common.BlockContext) (approvedJoins, expiredJoins []*types.AccountID, err error)
 	Commit() error
 	Rollback()
-	GenesisInit(ctx context.Context, db sql.DB, validators []*ktypes.Validator, genesisAccounts []*ktypes.Account, initialHeight int64, initialDBOwner string, chain *common.ChainContext) error
+	GenesisInit(ctx context.Context, db sql.DB, genesisConfig *config.GenesisConfig, chain *common.ChainContext) error
```
jchappelow, line 44: Seems fine to me.  The separate args I think were a relic from before we fleshed out the GenesisConfig

jchappelow (dismissed):
LGTM but @charithabandi

node/block_processor/interfaces.go
brennanjl: Right that's what I assumed. We were trying to keep txapp agnostic to abci / anything comet related

Yaiba:
Resolves #1274

node/block_processor/processor.go
```diff
@@ -418,6 +402,7 @@ func (bp *BlockProcessor) ExecuteBlock(ctx context.Context, req *ktypes.BlockExe
 		Timestamp:    req.Block.Header.Timestamp.Unix(),
 		ChainContext: bp.chainCtx,
 		Proposer:     req.Proposer,
+		Hash:         req.BlockID,
```
jchappelow, line 405: We have this in `req.Block.Hash()`, but seems ok to dup it in `req.BlockID` to avoid recomputing it.

jchappelow: Oh, it was already in `BlockExecRequest`.  :confused: well cool!

brennanjl: Something I saw in `go-ethereum` was that they had a `Hash()` method that they used, but they would cache the hash in an unexported field so as to not recompute it.

If there's a chance `Block.ID` might not be set yet, I can do that

jchappelow: I had a cached hash early on just like that, but it opens a can of concurrency safety worm.  I just didn't realize there was already a `BlockID` field in this request struct, so I think we can just go with what we have.

common/context.go
```diff
@@ -31,6 +31,10 @@ type BlockContext struct {
 	ChainContext *ChainContext
 	// Height gets the height of the current block.
 	Height int64
+	// Hash is the hash of the current block.
+	// It can be empty if the block is the genesis block and
+	// no initial state hash was specified in the genesis configuration.
+	Hash types.Hash
```
charithabandi, line 37: Can we call it BlkID or BlkHash to be more explicit and to be consistent with the rest of the system?

charithabandi (approved):
LGTM!

jchappelow (approved):

jchappelow:
Resolved conflicts for you.  Just noticed @charithabandi 's comment though.  Could call it `BlkHash`.  But this is `BlockContext`, so it's fairly well contextualized IMO.

common/context.go
brennanjl: Idk, I feel like `Block.Hash` is pretty self-explanatory.

If it was in `TxContext`, then I would agree that `Hash`, `Height`, `Proposer` are not very descriptive. But within a `Block`, feel like it's pretty descriptive...

Yaiba:
Is this good to merge?

common/context.go
charithabandi: Yeah, its the block context, makes sense. good to merge

PR #1279, opened 2025-01-23
accouts: composite primary key over id and id_type

jchappelow:
This makes no functional or user visible changes, but it does alter how the account store tables work.

With

```json
"alloc": [
    {
      "id": "0xAfFDC06cF34aFD7D5801A13d48C92AD39609901D",
      "key_type": "secp256k1",
      "amount": 10000000000000000000000000000000000000000000
    }
  ],
```

we had

```
# table kwild_accts.accounts ;
┌─[ RECORD 1 ]──────────────────────────────────────────────────────────────────────────────┐
│ identifier │ \x14000000affdc06cf34afd7d5801a13d48c92ad39609901d09000000736563703235366b31 │
│ balance    │ 10000000000000000000000000000000000000000000                                 │
│ nonce      │ 0                                                                            │
└────────────┴──────────────────────────────────────────────────────────────────────────────┘
```

With this change we have

```
 # table kwild_accts.accounts ;
┌─[ RECORD 1 ]──────────────────────────────────────────────┐
│ identifier │ \xaffdc06cf34afd7d5801a13d48c92ad39609901d   │
│ id_type    │ 0                                            │
│ balance    │ 10000000000000000000000000000000000000000000 │
│ nonce      │ 0                                            │
└────────────┴──────────────────────────────────────────────┘
```

So more compact and readable for debugging.

This didn't take very long to do, so open to any suggestions even if it means something completely different.

### DIFF
--- a/core/crypto/keys.go
+++ b/core/crypto/keys.go
@@ -76,6 +76,14 @@ func ParseKeyType(s string) (KeyType, error) {
 	return "", fmt.Errorf("unknown key type: %s", s)
 }
 
+func ParseKeyTypeID(id uint32) (KeyType, error) {
+	kt, ok := encodingIDs[id]
+	if !ok {
+		return "", fmt.Errorf("unknown key type ID: %d", id)
+	}
+	return kt, nil
+}
+
 func UnmarshalPublicKey(b []byte, kt KeyType) (PublicKey, error) {
 	kd, ok := keyTypes[kt]
 	if !ok {
--- a/node/accounts/accounts.go
+++ b/node/accounts/accounts.go
@@ -2,11 +2,13 @@ package accounts
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"math/big"
 	"sync"
 
+	"github.com/kwilteam/kwil-db/core/crypto"
 	"github.com/kwilteam/kwil-db/core/log"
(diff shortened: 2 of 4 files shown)


brennanjl (approved):
lgtm

jchappelow:
Summarizing some good discussion in Slack about this:

- The owner of a given secp256k1 pubkey can (with difficulty) transact with different auth types where one indicates eth address as sender and the other indicates full pubkey, and these are unfortunately considered different accounts.
- If there were a hypothetical third secp256k1 auth type that also uses full pubkey as id/sender (just some different message digest function), that will end up the same account.  This was the reason behind using key type rather than auth type in the accounts store.
- The above effects are inconsistent.
- We cannot get the pubkey (practically) for eth web3 wallets, so we can't change that.
- It's unlikely a user will fall into this pitfall, since it takes real effort to make a transaction from a browser / node app that would use the non-personal-sign / direct auth type and the full pubkey.  They'd need a kwil-js or kwil-cli fork, since these tools are specifically designed to match and use eth addresses.
- We could consider linking/aliasing certain account types, for the purpose of nonce and balance checks via only 0x adddress.  This can be done after the first transaction from a user that contains a valid signature that allows the recovered public key to be recorded in this lookup table.  Balance lookups by eth address or full pubkey will work.  This could be a mess of an implementation.

We decided on no changes for now.

PR #1281, opened 2025-01-23
fix rlp decode

jchappelow:
This fixes RLP decoding.   Decoding with the `seriailize` package gets the codec from a prefix in the `[]byte`.  It is not specified manually.  That is only for encoding.

Example RLP support, e.g. from an extension:

```go
package main

import (
	"bytes"

	"github.com/kwilteam/kwil-db/core/types/serialize"
	"github.com/kwilteam/kwil-db/node/rlp" // register RLP codec with serialize package
)

func main() {
	s := "hi"
	bts, err := serialize.EncodeWithEncodingType(s, rlp.EncodingTypeRLP)
	if err != nil {
		panic(err)
	}

	bts2, err := serialize.EncodeWithCodec(s, *rlp.Codec)
	if err != nil {
		panic(err)
	}

	if !bytes.Equal(bts, bts2) {
		panic("not equal")
	}

	var s2 string
	err = serialize.Decode(bts, &s2)
	if err != nil {
		panic(err)
	}

	println(s2) // "hi"
}
```

### DIFF
--- a/core/types/serialize/encode.go
+++ b/core/types/serialize/encode.go
@@ -156,14 +156,6 @@ func Decode(bts []byte, v any) error {
 	return codec.Decode(val, v)
 }
 
-func DecodeWithEncodingType(bts []byte, v any, encodingType EncodingType) error {
-	codec, ok := encodings[encodingType]
-	if !ok {
-		return fmt.Errorf("unregistered encoding type: %d", encodingType)
-	}
-	return codec.Decode(bts, v)
-}
-
 // DecodeGeneric decodes the given serialized data into the given value. See
 // also Decode for use with an existing instance. This is generally no more
 // useful than Decode; it is syntactic sugar that requires no existing instance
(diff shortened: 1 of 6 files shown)


core/types/serialize/encode.go
```diff
@@ -156,14 +156,6 @@ func Decode(bts []byte, v any) error {
 	return codec.Decode(val, v)
 }
 
-func DecodeWithEncodingType(bts []byte, v any, encodingType EncodingType) error {
-	codec, ok := encodings[encodingType]
-	if !ok {
-		return fmt.Errorf("unregistered encoding type: %d", encodingType)
-	}
-	return codec.Decode(bts, v)
```
jchappelow, line 164: This was just wrong.  The encoding type prefix is the mechanism.  Not sure why I put this here.

charithabandi (approved):
ok, lgtm! Thanks for doing this!

jchappelow:
We can make RLP baked into `core/types/serialize` again, but the system we have presently is with the RLP codec in an extension-like package that registers into `serialize`.  This way an extension author can use it, but the client sdk isn't burdened by go-ethereum's GPL code unless they opt-in.

brennanjl:
Awesome thank you

PR #1282, opened 2025-01-23
Serialization and other minor fixes, updated docs for setup commands

charithabandi:


config/config.go
```diff
@@ -131,6 +131,8 @@ func (gc *GenesisConfig) SanityChecks() error {
 		return errors.New("both start and end height should be set or unset")
 	}
 
+	// ensure that the leader is part of the validator set
```
jchappelow, line 134: TODO?

charithabandi: Yeah, lot more fixes to come. Just pushed serialization fixes for now. In draft

jchappelow: Ah, sorry didnt' see draft status, my bad

app/setup/genesis.go
```diff
@@ -164,29 +164,39 @@ func mergeGenesisFlags(conf *config.GenesisConfig, cmd *cobra.Command, flagCfg *
 				return nil, fmt.Errorf("invalid format for alloc, expected id#keyType:balance, received: %s", a)
 			}
 
+			// <id#keyType>, keyType is optional for 0x addresses
```
jchappelow, line 167: Just being pedantic, but key type is actually disallowed/invalid for 0x addresses, not just optional. (with this current code)

app/snapshot/create.go
```diff
@@ -136,23 +144,30 @@ func PGDump(ctx context.Context, dbName, dbUser, dbPass, dbHost, dbPort string,
...
 	defer gzipWriter.Close()
 
-	pgDumpCmd := exec.CommandContext(ctx,
-		"pg_dump",
+	namespaces, err := namespaces(ctx, dbName, dbUser, dbPass, dbHost, dbPort)
+	if err != nil {
+		return -1, nil, "", nil, fmt.Errorf("failed to get namespaces: %w", err)
+	}
+
+	schemaArgs := make([]string, 0)
```
jchappelow, line 152: `make([]string, 0, len(namespaces))`, otherwise no need to `make` unless it really must be non-nil when empty.  Honestly though this may as well be appending to `args` below.

app/snapshot/create.go
```diff
@@ -218,37 +236,37 @@ func PGDump(ctx context.Context, dbName, dbUser, dbPass, dbHost, dbPort string,
...
+
 			voterID, err := hex.DecodeString(strs[1][3:]) // Remove the leading \\x
 			if err != nil {
-				return -1, nil, nil, fmt.Errorf("failed to decode voter ID: %w", err)
+				return -1, nil, "", nil, fmt.Errorf("failed to decode voter ID: %w", err)
 			}
 
 			// voterID is the encoded public key
-			pubkey, err := crypto.WireDecodePubKey(voterID)
+			keyType, pubkey, err := crypto.WireDecodeKeyType(voterID)
```
jchappelow, line 256: Yikes this would be something I would definitely forget to update if we change the voters table encoding.  :thinking: Tough situation with a pgdump to parse.  Probably OK for now but I almost wish we coupled this encoding to the `voting` package.

app/snapshot/create.go
```diff
@@ -337,3 +362,78 @@ func chainHeight(ctx context.Context, dbName, dbUser, dbPass, dbHost, dbPort str
...
+	}
+	defer pool.Close()
+
+	tx, err := pool.BeginReadTx(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to begin transaction: %w", err)
+	}
+	defer tx.Rollback(ctx)
+
+	res, err := tx.Execute(ctx, "select name from kwild_engine.namespaces where type='USER'")
```
jchappelow, line 415: Some pretty tight implicit couplings with other packages.  Can we at least point to where the components are affected in engine and/or DB code?

jchappelow (dismissed):
Lots of great changes here!  Just a few concerns about the app commands having some implicit coupling and hard to maintain code, but otherwise fine.

brennanjl:
Will wait for @jchappelow to confirm

jchappelow:
Good to merge.  Please make a new squashed commit message with the merge.

PR #1283, opened 2025-01-23
extensions/encoding: locate registerable codecs, add borsch, example

jchappelow:
This is a follow-up on https://github.com/kwilteam/kwil-db/pull/1281.

- make a central location for opt-in / registerable codecs for the `core/types/serialize` package
- add `borsh` (was already a dependency because of an auth extension)
- add an `example` package with a `main` function showing use and comparing RLP and Borsh


extensions/encoding/example/main.go
```diff
@@ -0,0 +1,121 @@
...
+type example2 struct { // this is not RLP-encodable for several reasons
+	Field1 string
+	Field2 int // not RLP-encodable
+	Field3 bool
+	Field4 []byte
+	Field5 exampleInner
+	Field6 map[string]string // not RLP-encodable
+	Field7 int64             // not RLP-encodable
+	Field8 *string
+}
```
jchappelow, line 38: I feel like this is a decent demo of RLP shortcomings vs. Borsch in particular.  I also didn't see any demo or docs of using `serialize.RegisterCodec` or the other codec types, which is probably helpful for extensions.

jchappelow:
This is obviously not a release requirement, but open to thoughts.  It's the next steps after https://github.com/kwilteam/kwil-db/pull/1281 to make encodings more extension-like.

brennanjl:
Will take a look in a bit

brennanjl:
I actually like this a lot. My one concern is that we don't actually need borsch as a dependency anymore (because idOS moved the nep413 authenticator to their own repo), so maybe it doesn't make sense to use as the example, but otherwise I like it.

I'm ok leaving borsch, just wanted to lyk in case you wanted to remove it (I know NEAR isn't always the most stable).

brennanjl (approved):
lgtm, but since it re-adds borsch (which got removed in https://github.com/kwilteam/kwil-db/pull/1292 which I just merged), I'll wait and let you merge in case you care to change it. No strong opinion from me though

jchappelow:
Just tucked away the borsh codec and the example app in a module.  Does not affect the primary Kwil modules this way.

brennanjl (approved):

PR #1284, opened 2025-01-23
improve several kwild cmd issues with config parsing and error handling

jchappelow:
Going through several small things that @KwilLuke mentioned yesterday

- some commands try to read `~/.kwild/config.toml` unnecessarily, leading to confusing errors if the file is invalid since it's really irrelevant to the command
- fix misuses of `display.PrintErr` and document proper use
- use the `conf.ActiveConfig`, which is the merged configuration in several commands instead of manually loading the toml file
- revise the `start` command's docs to reflect that it is no longer the default command
- `setup init` will error if the destination directory already exists.  Previous it ran with unpredictable results.

### DIFF
--- a/app/setup/genesis-hash.go
+++ b/app/setup/genesis-hash.go
@@ -11,12 +11,12 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/kwilteam/kwil-db/app/node/conf"
 	"github.com/kwilteam/kwil-db/app/shared/bind"
 	"github.com/kwilteam/kwil-db/app/shared/display"
 	"github.com/kwilteam/kwil-db/app/snapshot"
 	"github.com/kwilteam/kwil-db/config"
 	"github.com/kwilteam/kwil-db/node"
-	"github.com/kwilteam/kwil-db/node/pg"
 	"github.com/spf13/cobra"
 )
 
@@ -41,7 +41,7 @@ kwild setup genesis-hash --snapshot "/path/to/snapshot.sql.gz" --genesis "~/.kwi
 )
 
 func GenesisHashCmd() *cobra.Command {
-	var genesisFile, rootDir, snapshotFile string
+	var genesisFile, snapshotFile string
 
 	cmd := &cobra.Command{
 		Use:     "genesis-hash",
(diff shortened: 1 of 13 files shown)


app/setup/genesis-hash.go
```diff
@@ -61,43 +64,30 @@ func GenesisHashCmd() *cobra.Command {
...
-				err = ensureTmpKwilAdminDir(dir)
+				// clean up any previous temp admin snapshots
+				err = cleanupTmpKwilAdminDir(dir)
 				if err != nil {
 					return display.PrintErr(cmd, err)
 				}
-				defer cleanupTmpKwilAdminDir(dir) // clean up temp admin snapshots directory on exit after app hash computation
 
-				// clean up any previous temp admin snapshots
-				err = cleanupTmpKwilAdminDir(dir)
```
jchappelow, line 97: @charithabandi AFAICT, this was deleting the directory that was just created above, and the `PGDump` function was recreating it internally.  Please double check this for me.

app/setup/genesis-hash.go
```diff
@@ -107,15 +97,22 @@ func GenesisHashCmd() *cobra.Command {
...
+				}
+			}
+			return display.PrintCmd(cmd, &genesisHashRes{
+				Hash: base64.StdEncoding.EncodeToString(appHash),
+			})
 		},
 	}
 
 	cmd.Flags().StringVarP(&genesisFile, "genesis", "g", "", "optional path to the genesis file to patch with the computed app hash")
-	cmd.Flags().StringVarP(&rootDir, "root-dir", "r", "", "optional path to the root directory of the kwild node from which the genesis hash will be computed")
```
jchappelow, line 115: Root is bound globally.

brennanjl: Should we have root dir bound globally?

Entire subcommand (e. validators) have no use for it.

app/setup/genesis-hash.go
charithabandi: Yeah, its temporary dir for storing snapshots till we get the genesis hash, we don't actually need it after.

jchappelow: I know that's why we defer the cleanup, but I'm saying the original order in this function made no sense.  It created it and immediately removed it, then created it again.

app/setup/genesis-hash.go
jchappelow: Almost all commands actually do.  In the case of all the commands that use admin RPC, the on-disk config is used as a default when `--rpcserver` is to set on the CLI.  This means root directory can affect it.
We have the ability to hide it for other commands as needed though.  For example `key gen` had it's own output flag, and the root flag their was confusing, so I hid it.

brennanjl (approved):

Issue #1285, opened 2025-01-23
TxQuery needs a pending boolean flag


jchappelow:
The server side implementation currently just checks the block index:

```go
func (n *Node) TxQuery(ctx context.Context, hash types.Hash, prove bool) (*ktypes.TxQueryResponse, error) {
	tx, height, blkHash, blkIdx, err := n.bki.GetTx(hash)
	if err != nil {
		return nil, err
	}
```

This also need to be able to check mempool  (in v0.9 there was a pending/unconfirmed flag for query-tx, right?)

PR #1286, opened 2025-01-23
node: tx query checks mempool

jchappelow:
Resolves https://github.com/kwilteam/kwil-db/issues/1285


jchappelow:
OK, actually ready now.  Acceptance test updated with the syncbroadcast option removed, which we used as a workaround until this bug was fixed.  `Test_Engine` acceptance test is working.

Also restored `TestLocalDevSetup` (and `task dev:up`).

charithabandi (approved):

PR #1289, opened 2025-01-23
add namespace manager

brennanjl:
As discussed, I have added a way for other parts of Kwil to read the current set of namespaces.

Something worth noting is that namespaces can be read in the middle of a block, so it is not transactional within the engine (e.g. if a block has 10 txs, and a namespace is created in tx 1 of 10, the system will see it before the other 9 txs commit).

This can be fixed, but it requires writing to the manager at a higher level (in block processor)

### DIFF
--- a/app/node/deps.go
+++ b/app/node/deps.go
@@ -17,7 +17,7 @@ import (
 // datasets in different postgresql "schema".
 type dbOpener func(ctx context.Context, dbName string, maxConns uint32) (*pg.DB, error)
 
-func newDBOpener(host, port, user, pass string) dbOpener {
+func newDBOpener(host, port, user, pass string, filterSchemas func(string) bool) dbOpener {
 	return func(ctx context.Context, dbName string, maxConns uint32) (*pg.DB, error) {
 		cfg := &pg.DBConfig{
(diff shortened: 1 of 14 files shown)


brennanjl:
@charithabandi The second commit added a function that lists all schema that the engine needs to have pg dumped

app/node/deps.go
```diff
@@ -17,7 +17,7 @@ import (
 // datasets in different postgresql "schema".
 type dbOpener func(ctx context.Context, dbName string, maxConns uint32) (*pg.DB, error)
 
-func newDBOpener(host, port, user, pass string) dbOpener {
+func newDBOpener(host, port, user, pass string, filterSchemas func(string) bool) dbOpener {
 	return func(ctx context.Context, dbName string, maxConns uint32) (*pg.DB, error) {
 		cfg := &pg.DBConfig{
```
jchappelow, line 22: Should "just work" if you give this to `DBConfig` here.

brennanjl: Whoops, I totally meant to do that and forgot

jchappelow (dismissed):
Looks good from DB perspective.

jchappelow (approved):

PR #1291, opened 2025-01-24
txapp: grant ownership to db owner is idempotent

brennanjl:
This fixes a bug where a network that starts from a snapshot fails to start due to an owner already existing.


charithabandi (approved):

brennanjl:
Putting up a different fix

PR #1293, opened 2025-01-24
node: remove namespace manager Ready() panic

brennanjl:
Removes panic on namespace manager


app/node/namespace_manager.go
```diff
@@ -55,8 +55,7 @@ func (n *namespaceManager) Filter(ns string) bool {
 	defer n.mu.RUnlock()
 
 	if !n.ready {
-		// this would indicate a bug in our startup process
-		panic("namespace manager not ready")
+		return false
```
charithabandi, line 58: I don't think we need this ready field or Ready() method anymore right?

charithabandi (approved):

PR #1296, opened 2025-01-24
propagate context to node constructor

brennanjl:
FYI @charithabandi. This fixes a minor bug with context propagation.


charithabandi:
@brennanjl I am gonna address few of these in my PR. We can close this.

PR #1297, opened 2025-01-24
autogen, peercount, logging updates

jchappelow:
This change primarily is cleanup and logging minimizations.  Part of this is avoiding endless spam of the warning `No connected peers and no known addresses to dial!` when running a single node network.  This lead to some `autogen` fixes.

### DIFF
--- a/app/node/node.go
+++ b/app/node/node.go
@@ -318,7 +318,7 @@ func loadGenesisAndPrivateKey(rootDir string, autogen bool, dbOwner string) (pri
 		}
 		// If the genesis file exists and --autogen was used, disallow a genesis
 		// file with either multiple validators or a different leader the self.
-		if autogen && (len(genCfg.Validators) > 0 || !genCfg.Leader.PublicKey.Equals(privKey.Public())) {
+		if autogen && (len(genCfg.Validators) > 1 || !genCfg.Leader.PublicKey.Equals(privKey.Public())) {
 			return nil, nil, errors.New("cannot use --autogen with genesis config for a multi-node network")
 		}
 		return privKey, genCfg, nil
--- a/app/node/node.go
+++ b/app/node/node.go
@@ -130,11 +130,17 @@ func runNode(ctx context.Context, rootDir string, cfg *config.Config, autogen bo
 
 	// if running in autogen mode, and config.toml does not exist, write it
 	tomlFile := config.ConfigFilePath(rootDir)
-	if autogen && !fileExists(tomlFile) {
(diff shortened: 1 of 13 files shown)


node/consensus/leader.go
```diff
@@ -306,47 +309,55 @@ func (ce *ConsensusEngine) processVotes(ctx context.Context) error {
...
 
-		ce.log.Infoln("Announce committed block", ce.state.blkProp.blk.Header.Height, ce.state.blkProp.blkHash,
-			ce.state.blockRes.paramUpdates)
-		// Broadcast the blockAnn message
-		go ce.blkAnnouncer(ctx, ce.state.blkProp.blk, ce.state.commitInfo)
+	// Commit the block and broadcast the blockAnn message
+	if err := ce.commit(ctx); err != nil {
+		ce.haltChan <- fmt.Sprintf("Error committing block %v: %v", blkProp.blkHash, err)
+		return
+	}
```
jchappelow, line 349: @charithabandi Is this the best way to handle this?  I noticed the `error` return from `processVotes` was unused and a `commit` error did not have an immediate effect.  Since all callers of `processVotes` ignored the error return, and there is a halt channel use in the `nack` majority case, I assume the intent is to halt via the channel if the commit failed.

app/node/node.go
```diff
@@ -306,7 +312,15 @@ func loadGenesisAndPrivateKey(rootDir string, autogen bool, dbOwner string) (pri
...
 		}
+		// If the genesis file exists and --autogen was used, disallow a genesis
+		// file with either multiple validators or a different leader the self.
+		if autogen && (len(genCfg.Validators) > 0 || !genCfg.Leader.PublicKey.Equals(privKey.Public())) {
+			return nil, nil, errors.New("cannot use --autogen with genesis config for a multi-node network")
+		}
 		return privKey, genCfg, nil
+	} else if !autogen {
+		// If not using --autogen, the genesis file must exist!
+		return nil, nil, fmt.Errorf("genesis file %s does not exist (did you mean to use --autogen for a test node?)", genFile)
```
jchappelow, line 323: I'm 99% certain we don't want to generate a random genesis file unless `--autogen` was used.

node/consensus/leader.go
charithabandi: Yeah , looks correct to me.

charithabandi (dismissed):

jchappelow:
Ugh, one more thing coming...

charithabandi (approved):

PR #1300, opened 2025-01-27
core/client/example: update for new SQL schema and client

jchappelow:
This updates the client example app.

For now you have to:
- use a local node at 127.0.0.1
- disable gas
- be the `db_owner`

```
$  ./mykwilapp -key 00c2e88...
address: 0xE0f13c95225F34f86E99f83984e286dcaF321E37
Connected to Kwil chain "kwil-testnet", block height 48
Account AccountID{identifier = e0f13c95225f34f86e99f83984e286dcaf321e37, keyType = secp256k1} balance = 0, nonce = 25
Success: "deploy schema" in transaction 5a4deccb71edce66be922dcf4751b30dd752990498931e4c3b57d4ebc0def076
Executing action "tag" to insert data...
Success: "execute action tag" in transaction b6fd13af216a2dbe080df2beb4c1e5cfdce406ecb57ee544a06eae29f4572fff
Success: "execute action tag" in transaction 4deb10eb83b6f9209429a2edbaf2f43cf35d1daf6021b7c92320d853eed7d527
All entries in tags table:
column names: []string{"id", "author", "msg", "val"}
values:
[]interface {}{"5f02d8be-57b0-5a0c-b766-7db567448967", "0xE0f13c95225F34f86E99f83984e286dcaF321E37", "jon was here", 12}
[]interface {}{"f1428604-a4a5-5398-b93d-5655b7e2bc78", "0xE0f13c95225F34f86E99f83984e286dcaF321E37", "jon was here AGAIN", 99}
Success: "delete all" in transaction e490129079816dbf57b1da0482aa8937bb6451c290ab8e6d6302afbbd8892c86
2025/01/27 13:16:37 deleted all!
```


core/client/example/main.go
```diff
@@ -0,0 +1,275 @@
...
+
+	// Check our account's balance.
+	acctInfo, err := cl.GetAccount(ctx, acctID, types.AccountStatusLatest)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("Account %s balance = %v, nonce = %d\n", acctID, acctInfo.Balance, acctInfo.Nonce)
+
+	// List previously deployed database owned by us.
+	/*datasets, err := cl.ListDatabases(ctx, acctID)
```
jchappelow, line 96: I'm curious if there's anything we can do that is like this, to discover current actions and their signatures.

jchappelow:
This should be updated with some ad hoc queries to the info namespace, and again when we deploy a public net, but ok for now, I think.

brennanjl:
Awesome thank you

brennanjl (approved):

PR #1301, opened 2025-01-27
Statesync discovery should occur prior to the db and the rest of the …

charithabandi:
…modules initializations


jchappelow (commented):
Only comment so far is that d.ctx should be removed in this pr

brennanjl (approved):

PR #1302, opened 2025-01-27
Start connect backoff

jchappelow:
This makes the startup peer connection (handled by `maintainMinPeers`) use exponential backoff on a per-peer basis, just like the reconnect logic started when a peer disconnects (initiated by the `network.Notifiee`)

This also fixes the regression with the `network.Notifiee` being de-registered from the Host.

Finally, this has the `P2PService` register dummy stream handlers for all of the "required" protocol IDs.  They are replaced with the actual handlers in `Node`  startup when the CE is ready to take over.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -243,31 +243,17 @@ func (n *Node) Start(ctx context.Context, bootpeers ...string) error {
 		return err
 	}
 
-	bootpeersMA, err := peers.ConvertPeersToMultiAddr(bootpeers)
-	if err != nil {
-		return err
-	}
-
-	// connect to bootstrap peers, if any.
-	//
-	// NOTE: it may be preferable to simply add to Host's peer store here and
-	// let PeerMan manage connections.
-	for i, peer := range bootpeersMA {
-		peerInfo, err := makePeerAddrInfo(peer)
-		if err != nil {
-			n.log.Warnf("invalid bootnode address %v from setting %v", peer, bootpeers[i])
-			continue
-		}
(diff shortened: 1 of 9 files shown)


jchappelow:
Apparently I have a blind spot in my PR list.  I thought this was done and merged.  Bringing out of draft in a sec...

jchappelow:
eh, rebase error....

node/node.go
```diff
@@ -243,31 +243,17 @@ func (n *Node) Start(ctx context.Context, bootpeers ...string) error {
...
-		if err != nil {
-			n.log.Warnf("invalid bootnode address %v from setting %v", peer, bootpeers[i])
-			continue
-		}
-		if err = n.checkPeerProtos(ctx, peerInfo.ID); err != nil {
+	// Check protocol support for connected peers, which were established
+	// earlier during P2PService startup.
+	for i, peer := range n.peers() {
+		if err = n.checkPeerProtos(ctx, peer); err != nil {
 			n.log.Warnf("WARNING: peer does not support required protocols %v: %v", bootpeers[i], err)
```
charithabandi, line 250: hmm, we don't provide bootpeers anymore in the Start constructor. We should probably get the peer info from the `peer` variable in the loop

jchappelow: Yeah, this was my rebase error.  :)
Fix pushed for this and TestDualNodeMocknet

jchappelow: Good catch!

charithabandi (approved):
lgtm!

PR #1305, opened 2025-01-28
erc20 reward ext

Yaiba:
This pr contains erc20 reward extension as our built-in extension, it will be enabled by default


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kwilteam/kwil-db
 
-go 1.22.0
+go 1.23.1
```
Yaiba, line 3: eh need to fix this

node/exts/erc20reward/ext.go
```diff
@@ -0,0 +1,993 @@
...
+	if err != nil {
+		return err
+	}
+
+	if finalized {
+		h.SafeNonce += 1
+	}
+
+	if resultFn != nil {
+		return resultFn([]any{})
```
brennanjl, line 847: Don't call this. It will always be not nil (the engine creates an empty one if it is passed as nil). Additionally, it should only be called if you have results to return. This means the action will return one empty row, instead of returning no rows

node/exts/erc20reward/sql.go
```diff
@@ -0,0 +1,757 @@
...
+		{Type: types.ByteaType, Nullable: false},
+		{Type: types.IntType, Nullable: false},
+		{Type: types.ByteaType, Nullable: false},
+		{Type: types.UUIDType, Nullable: false},
+		{Type: types.ByteaType, Nullable: false},
+	}
+}
+
+func GenPendingRewardID(recipient string, amount string, height int64) *types.UUID {
+	// TODO: seems not unique, because we might have same recipient receive same amount multiple times in same height
```
Yaiba, line 309: @brennanjl This could be an issue because I cannot think of a combination to make sure the id is unique.
This cannot grantee unique even using txID/blockHash in the generation, because a tx can issue to same user same amount multiple times. 
Thoughts?

brennanjl: That's a good point. Definitely a way to do it, might be a bit hacky though.

Immediate idea would be to add some sort of field to tx context (e.g. like how the standard library value have `context.WithValue`)

brennanjl:
```go
txCtx.SetValue(string, any)

txCtx.GetValue(string) (any, bool)
```

brennanjl (approved):

PR #1310, opened 2025-01-28
node: add SetValue and Value in tx context

brennanjl:
Adds the ability for extensions to attach arbitrary info to a transaction context, which can be referenced later

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -71,6 +71,23 @@ type TxContext struct {
 	Caller string
 	// Authenticator is the authenticator used to sign the transaction.
 	Authenticator string
+	// values is a map of values that can be set and retrieved by extensions.
+	values map[string]any
+}
+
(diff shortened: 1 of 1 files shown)


brennanjl:
@Yaiba

common/common.go
```diff
@@ -71,6 +71,27 @@ type TxContext struct {
 	Caller string
 	// Authenticator is the authenticator used to sign the transaction.
 	Authenticator string
+	// values is a map of values that can be set and retrieved by extensions.
+	values map[string]any
```
Yaiba, line 75: is this thread safe?

brennanjl: Transactions have to be deterministic, so we never have any sort of asynchronicity

Yaiba (approved):
LGTM

PR #1311, opened 2025-01-28
engine: adds SET CURRENT NAMESPACE statement

brennanjl:
Over the past week this is something I have found myself wanting, and while implementing a new addition for the rewards extension I found myself needing it.

Currently, when writing raw SQL, namespaces can only be set when you prefix a statement with `{namespace}`. This PR adds a new statement, where you can call `SET CURRENT NAMESPACE TO some_namespace`. Once called, the default namespace for the rest of the execution will be that namespace.


Yaiba (approved):
LGTM

PR #1312, opened 2025-01-28
Don't clear out resolutions from the votestore if the DB is restored …

charithabandi:
…using a statesync snapshot

If the DB is restored from genesis snapshots during offline or ZDT migrations, the resolutions from the voting table should be cleared out especially the validator and migration resolutions as they are not applicable for the new network. However this should not be done for the statesync snapshots. (This is a regression from the changes I made yesterday)

This PR also adds a new integration test to test out unresolved resolutions in statesync snapshots

Also fixes incomplete statesync tests


jchappelow (approved):
OK with me.

PR #1316, opened 2025-01-28
set lastCommitInfo on startup

charithabandi:


jchappelow (approved):
LGTM.  That will cover node status on startup and replay from blockstore if app is already past zero.

PR #1318, opened 2025-01-29
Evm sync

brennanjl:
Still a lot to do, but this has a lot of the ground work for syncing logs from evm chains (so that we can know how many tokens the contract has, automatically detect the gnosis signers, etc.)

It also adds a new type of hook that I had been thinking about for a while but never quite had the right idea for: `EngineReadyHook`:

```go
// EngineReadyHook is a hook that is called on startup after the engine (and all of its extensions)
// have been initialized. It is meant to be used to perform any setup tasks that require the engine
// to be fully initialized.
type EngineReadyHook func(ctx context.Context, app *common.App) error
```

It also adds as a core feature a utility that has been needed for a while: a tool to automate the processing of logs from a system where order matters. Our resolution system does not guarantee order of logs (e.g. if you submit event1, event2, event3, they may get resolved in any order). IdOS ran into this being an issue, and built their own system. We also ran into this while making a demo that syncs nft transfers. To solve for this, we now have built-in storage that acts as an intermediate store for events that have been finalized, but we have not come to consensus on events that came before it.

It also adds two reserved namespace prefixes: `kwil_` and `pg_`. It will prevent users from making namespaces with these prefixes, however it will let extensions make namespaces with the prefix `kwil_`. This allows us to make built-in extension namespaces that exist at genesis (and the ability to add more in the future) without worrying about conflicting with user schemas. @Yaiba we should use the `kwil_` namespace for the meta_rewards namespace. We should _not_ use it for namespaces created due to extension initialization.

This PR will remove the eth deposits oracle (have not yet removed it at the time of writing), in favor of these more robust built-in oracles (which do the same thing + additional functionalities like transfers back on chain)


brennanjl:
@Yaiba it is mostly untested, but this has a lot of the infrastructure to allow us to sync data from ethereum.

Yaiba (approved):
just skimmed though the code, LGTM

PR #1320, opened 2025-01-29
p2p.external-address

jchappelow:
WIP, just want CI running

```
$  ./kwild start -r .testnet/node0 --p2p.external-address 70.113.64.78:6666
...
2025-01-28 18:41:12.011 [INF] NODE: This node is 0226b3ff29216dac187cea393f8af685ad419ac9644e55dce83d145c8b1af213bd#secp256k1 @ [70.113.64.78:6600 127.0.0.1:6600 192.168.30.232:6600]
```

### DIFF
--- a/node/p2p.go
+++ b/node/p2p.go
@@ -74,7 +74,17 @@ func NewP2PService(cfg *P2PServiceConfig, host host.Host) (*P2PService, error) {
 			return nil, fmt.Errorf("invalid P2P listen port: %s, %w", portStr, err)
 		}
 
-		host, err = newHost(ip, port, cfg.ChainID, cfg.PrivKey, cg, logger)
+		hostCfg := &hostConfig{
+			ip:              ip,
+			port:            port,
+			privKey:         cfg.PrivKey,
+			chainID:         cfg.ChainID,
+			connGater:       cg,
+			logger:          logger,
+			externalAddress: cfg.KwilCfg.P2P.ExternalAddress,
+		}
+
+		host, err = newHost(hostCfg)
 		if err != nil {
 			return nil, fmt.Errorf("cannot create host: %w", err)
 		}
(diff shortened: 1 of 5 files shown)


node/p2p.go
```diff
@@ -137,6 +147,11 @@ func (p *P2PService) Start(ctx context.Context, bootpeers ...string) error {
 			continue
 		}
 
+		// Don't dial ourself.
+		if peerInfo.ID == p.host.ID() {
```
charithabandi, line 151: Ah this is something I noticed, where a node complains that it doesn't support "discovery" protocol, I am guessing this is because it is dialing itself.

node/p2p.go
```diff
@@ -137,6 +147,11 @@ func (p *P2PService) Start(ctx context.Context, bootpeers ...string) error {
...
+			continue
```
jchappelow, line 152: setup testnet puts ourself in the bootpeers... nbd, just noticed it happening

brennanjl:
Will take a look today

node/p2p.go
jchappelow: Hmm, that could be something different.  Not sure why discover protocol would be mentioned in the error unless it was connecting to bootpeers before all the protocol handlers were set.  Could that be a consequence of recent p2p refactoring?

charithabandi (approved):
LGTM! I think we should probably advertise all the addresses, I will think more on it.

node/p2p.go
charithabandi: Yeah with the P2P refactoring, the P2P module is initialized way before the node and node is where most of the handlers are initialized. So yeah, as we discussed offline, we should probably add a mockHandlers, that gets eventually replaced during node initialization, if we want to take approaches where the connectionGater is more rigid. I can make this change in my PR.

brennanjl (approved):

PR #1322, opened 2025-01-29
Leader blocksync

charithabandi:
Two mechanisms are used to prevent the leader from committing a block at an already finalized height:

1. Manual checkpoints: The leader can be configured with a checkpoint height, which is the
    height the leader must sync to before proposing blocks. The node operator should set the
    checkpoint height based on the network's best height.
2. Validator feedback: If the leader proposes a block at an already finalized height, validators
    will respond with a NACK, indicating the leader is out-of-sync. They can provide an OutOfSyncProof
    as evidence, including the BlockHeader and the leader's signature of the block the validator is on.
    This forces the leader to catch up to the validator's block height before proposing blocks again. This
    corrective mechanism allows leader to eventually converge to the best height.

However the nodes can still end up in these weird race scenarios where majority of Validators are in-sync with the leader, but not with the Validator with BestHeight, and the Ack's from the out of sync validators might end up reaching the leader faster than the OutOfSync NACK, eventually making the leader to commit the block. But the situation is not unrecoverable, leader and these Validators can always be wiped out and restarted to the correct state.


jchappelow:
When this merges we need a descriptive commit message, which actually would be good for the PR description too.  ;)

node/consensus/blocksync.go
```diff
@@ -12,100 +11,70 @@ import (
...
+			return fmt.Errorf("leader block sync failed: %w", err)
+		}
+
+		// verify the checkpoint hash and height
+		if err := ce.VerifyCheckpoint(); err != nil {
+			return fmt.Errorf("checkpoint verification failed: %w", err)
+		}
+
+		// Leader has synchronized with the network up to the checkpoint height,
+		// now perform a best effort block sync to catch up further if needed.
```
jchappelow, line 40: :+1: good code comments

node/consensus/blocksync.go
```diff
@@ -156,8 +125,8 @@ func (ce *ConsensusEngine) syncBlocksUntilHeight(ctx context.Context, startHeigh
 // the block is successfully retrieved from the network.
 func (ce *ConsensusEngine) syncBlockWithRetry(ctx context.Context, height int64) error {
 	_, rawblk, ci, err := ce.getBlock(ctx, height)
-	if err != nil { // all kinds of errors?
-		return fmt.Errorf("error requesting block from network: height : %d, error: %w", height, err)
+	if err != nil {
+		return err
```
jchappelow, line 129: Don't want the extra error message context?  The issue with these `retrievers` is that they actually do two big things: `retrieve` and `apply`.  I found it was hard in the caller to tell where the error was coming from.

node/consensus/blocksync.go
```diff
@@ -229,3 +198,84 @@ func blkRetrier(ctx context.Context, maxRetries int64, fn func() error) error {
 		}
 	}
 }
+
+// func (ce *ConsensusEngine) leaderBlockSync(ctx context.Context) error {
```
jchappelow, line 202: I would suggest either removing this or putting it back where it was around but using block comments.

i.e. 

<pre>
/* func (ce *ConsensusEngine) leaderBlockSync(ctx context.Context) error {

...


*/
</pre>

So it's all there there without the ugly `//` comment formatting.

node/consensus/follower.go
```diff
@@ -9,12 +9,15 @@ import (
...
 )
 
-// AcceptProposal checks if the node should download the block corresponding to the proposal.
-// This should not be processed by the leader and the sentry nodes.
-// Validator should only accept the proposal if it is not already processing a block and
-// the proposal is for the next block to be processed.
-// If we receive a new proposal for the same height, abort the execution of the current proposal and
-// start processing the new proposal.
+// AcceptProposal determines if the node should download the block for the given proposal.
+// This function should not be executed by the leader or sentry nodes.
```
jchappelow, line 13: `// This function should not be executed by the leader or sentry nodes`
This doesn't sound right.  This is one of the functions used by the stream handler, so all node types will call it, and the response will be false if not a validator.  Or is this comment really accurate that leader or sentry must not even call this function?  The condition below seems to handle it inside:"

```
if ce.role.Load() != types.RoleValidator {
	return false
}
```

node/consensus/follower.go
```diff
@@ -25,9 +28,59 @@ func (ce *ConsensusEngine) AcceptProposal(height int64, blkID, prevBlockID types
...
+
+		ackRes := &types.AckRes{
+			ACK:        false,
+			NackStatus: &status,
+			BlkHash:    blkID,
+			Height:     height,
+			OutOfSyncProof: &types.OutOfSyncProof{
+				Header:    bestBlk.Header,
+				Signature: bestBlk.Signature,
+			},
```
jchappelow, line 75: :+1: I like it.

node/consensus/follower.go
```diff
@@ -176,7 +229,20 @@ func (ce *ConsensusEngine) processBlockProposal(ctx context.Context, blkPropMsg
 
 	if err := ce.validateBlock(blkPropMsg.blk); err != nil {
 		ce.log.Error("Error validating block, sending NACK", "error", err)
-		go ce.ackBroadcaster(false, blkPropMsg.height, blkPropMsg.blkHash, nil, nil)
+		sig, err := types.SignVote(blkPropMsg.blkHash, false, nil, ce.privKey)
+		if err != nil {
+			ce.log.Error("Error signing the voteInfo", "error", err)
```
jchappelow, line 234: We saw this ignored error from SignVote elsewhere recently.  Are we sure we want to send it with an invalid sig?

node/consensus/follower.go
```diff
@@ -214,7 +278,8 @@ func (ce *ConsensusEngine) processBlockProposal(ctx context.Context, blkPropMsg
 
 	signature, err := types.SignVote(blkPropMsg.blkHash, true, &ce.state.blockRes.appHash, ce.privKey)
 	if err != nil {
-		return fmt.Errorf("error signing the voteInfo: %w", err)
+		ce.log.Error("Error signing the voteInfo", "error", err)
+		return err
```
jchappelow, line 282: Again confused why we strip error message context out AND log.  Why do this?

node/consensus/leader.go
```diff
@@ -239,6 +239,50 @@ func (ce *ConsensusEngine) addVote(ctx context.Context, vote *vote, sender strin
 		return nil
 	}
 
+	// check if the vote is a Nack vote and the leader is out of sync
+	if !vote.ack && vote.nackStatus != nil && *vote.nackStatus == types.NackStatusOutOfSync && vote.outOfSyncProof != nil {
```
jchappelow, line 243: I would suggest a vote method like `OutOfSync() (*proof, bool)` or similar.

node/consensus/leader.go
```diff
@@ -239,6 +239,50 @@ func (ce *ConsensusEngine) addVote(ctx context.Context, vote *vote, sender strin
...
+		}
+		hash := vote.outOfSyncProof.Header.Hash()
+		valid, err := ce.pubKey.Verify(hash[:], vote.outOfSyncProof.Signature)
+		if err != nil {
+			ce.log.Errorf("Error verifying the out-of-sync proof: %v", err)
+			return nil
+		}
+		if !valid {
+			ce.log.Warn("Invalid vote: out-of-sync proof verification failed")
+			return nil
```
jchappelow, line 256: Will you command on the `nil` returns here?  I assume this is because it's not a leader node failure/bug, rather malfunction of the peer that sent this proof.

And more generally, this `addVote` method's docs do not say when a non-nil error will be returned, or how it should be handled  the caller.

node/types/commit.go
```diff
@@ -69,6 +79,20 @@ type Signature struct {
 	Data []byte
 }
 
+func EncodeSignature(sig *Signature) []byte {
+	var buf bytes.Buffer
+	sig.WriteTo(&buf)
+	return buf.Bytes()
+}
```
jchappelow, line 86: IMO a `Bytes()` method would be preferable since it is more discoverable.  The decode function is fine tho.

node/consensus/messages.go
```diff
@@ -162,14 +161,25 @@ func (ce *ConsensusEngine) NotifyACK(validatorPK []byte, ack types.AckRes) {
...
+	return &types.AckRes{
+		ACK:            v.ack,
+		AppHash:        v.appHash,
+		BlkHash:        v.blkHash,
+		Height:         v.height,
+		NackStatus:     v.nackStatus,
+		OutOfSyncProof: v.outOfSyncProof,
+		Signature:      v.signature,
+	}
+}
```
jchappelow, line 174: There's probably a good reason I just don't see at a glance, but why have a separate type?  Just a code comment would be good if there's an important reason.

node/types/commit.go
```diff
@@ -199,6 +223,7 @@ func (v *VoteInfo) Verify(blkID Hash, appHash Hash) error {
 	return nil
 }
 
+// TODO: should this sign the nack status as well? and the proof?
 func SignVote(blkID Hash, ack bool, appHash *Hash, privKey crypto.PrivateKey) (*Signature, error) {
```
jchappelow, line 227: Oh, gosh I think so, unless we're pretty confident p2p and the stream handlers are going to validate the authenticity of the entire message (and we check the source of the message is the validator)

node/types/consensus.go
```diff
@@ -62,16 +61,51 @@ func (cr *ConsensusReset) UnmarshalBinary(data []byte) error {
...
+	// other unknown miscellaneous reasons for nack
+	NackStatusUnknown NackStatus = "unknown"
+)
+
+func (ns NackStatus) String() string {
+	return string(ns)
+}
+
+type OutOfSyncProof struct {
+	Header    *types.BlockHeader
```
jchappelow, line 93: I think this struct should have docs, which in particular explain that this header is the alternate block header that the validator already has.

I'm not also starting to wonder if it's important for the nacker to also give the header for the subsequent block so the leader can see that the chain continues.  Otherwise, is there a case such as resets where leader having a replacement block at the same height is actually normal (it really is the best height).  A validator could give this proof in such a case even though there's not a fork at a lower height (again, this is the best height, just a reset or second block after some "normal" validator NACK such as apphash mismatches)?

node/consensus.go
```diff
@@ -269,18 +269,12 @@ func (n *Node) blkPropStreamHandler(s network.Stream) {
...
-		AppHash: appHash,
-		BlkHash: blkID,
-		Height:  height,
-
-		Signature:  signature,
-		PubKeyType: n.pubkey.Type(),
-		PubKey:     n.pubkey.Bytes(),
+func (n *Node) sendACK(msg *AckRes) error {
+	if msg == nil {
+		return errors.New("nil ACK response")
```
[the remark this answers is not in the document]
charithabandi: I added it just for sanity check as we are dereferencing it. But yeah i don't think callers will send a nil msg ever. can be removed.

node/consensus/follower.go
charithabandi: haha its more the former, I will update the comments.

node/consensus/follower.go
charithabandi: Ah no, definitely not. the error return is missing here, thought I fixed this in some PR. Anyways, fixing it here.

node/consensus/follower.go
charithabandi: fixed this.

node/types/consensus.go
charithabandi: Added docs to the struct. 
As discussed offline, the Header prvoided here is the node's best block header, so it's already the latest block

node/consensus/messages.go
charithabandi: okay, the fields are literally the same, so I updated the vote msg as a wrapper to the AckRes to be used by the consensus engine

node/types/commit.go
jchappelow: After chat we decided that since the proof is not recorded in chain data, only used in a p2p capacity, that it is best not to change the signature message, and instead check when handling the gossip message that the message originator/sender is the same as the `signature`'s pubkey.  In other words, prevent a node or validator from retransmitting a valid message with the original signature, but with the proof modified.

jchappelow:
LGTM, but might be a little last minute for a `beta.2`.  What's your feeling about it @charithabandi ?

charithabandi:
I am fairly confident, ran lots of tests for various sync scenarios.

node/node.go
```diff
@@ -305,15 +307,16 @@ func (n *Node) Start(ctx context.Context) error {
...
+		if err := n.startDiscoveryRequestGossip(ctx, ps); err != nil {
+			cancel()
+			return err
+		}
 
-	if err := n.startDiscoveryResponseGossip(ctx, ps); err != nil {
-		cancel()
-		return err
-	}
+		if err := n.startDiscoveryResponseGossip(ctx, ps); err != nil {
```
jchappelow, line 316: Will you pls remove the protocol IDs and stream handlers for this, unless the removal is possibly temporary?  `ProtocolIDDiscover` and `node/peers.(*PeerMan).DiscoveryStreamHandler`

charithabandi: Will remove this. Not temporary.

jchappelow (approved):

PR #1324, opened 2025-01-30
node startup and shutdown fixes

jchappelow:


jchappelow:
I *still* have not discovered what is closing the DB connection the instant the run context is canceled, but I don't see any hangs or big problems.

```
^Cshutdown signal received
2025-01-30 15:35:12.038 [INF] ADMINRPC: JSON-RPC server shutting down...
2025-01-30 15:35:12.038 [WRN] BP: Failed to rollback the consensus transaction {err=conn closed}
2025-01-30 15:35:12.038 [WRN] CONS: Error executing block {height=372 hash=9b29eee6a0a559b2ce6f6e46bd8b60f8c1c6838975a6924850d8e886755e5edd error=failed to finalize the block execution: conn closed}
2025-01-30 15:35:12.038 [INF] RPC: JSON-RPC server shutting down...
```

:confounded:

node/consensus/blocksync.go
```diff
@@ -118,8 +118,14 @@ func (ce *ConsensusEngine) replayBlockFromNetwork(ctx context.Context, requester
...
-			ce.log.Info("block request from the network failed", "height", height, "error", err)
-			break
+			if errors.Is(err, context.Canceled) {
+				return err
+			}
+			if errors.Is(err, types.ErrBlkNotFound) || errors.Is(err, types.ErrNotFound) {
+				break // no peers have this block, assume block sync is complete, continue with consensus
+			}
+			ce.log.Warn("unexpected error requesting block from the network", "height", height, "error", err)
+			break // should become an error if we are good about returning typed errors and handling above
```
jchappelow, line 128: ughhhh `2025-01-30 15:43:04.955 [WRN] CONS: unexpected error requesting block from the network {height=393 error=error committing block: COMMIT PREPARED failed: context canceled}`

node/block_processor/processor.go
```diff
@@ -251,10 +251,7 @@ func (bp *BlockProcessor) CheckTx(ctx context.Context, tx *ktypes.Transaction, h
 		}
 	}
 
-	readTx, err := bp.db.BeginReadTx(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to begin read transaction: %w", err)
-	}
+	readTx := bp.db.BeginDelayedReadTx()
```
jchappelow, line 254: DB comms for every `CheckTx` is rough.  Since acct store has cache, a delayed/lazy read transaction is a pretty good solution.

node/consensus/blocksync.go
```diff
@@ -118,8 +118,14 @@ func (ce *ConsensusEngine) replayBlockFromNetwork(ctx context.Context, requester
 
 	for {
 		if err := requester(ctx, height); err != nil {
-			ce.log.Info("block request from the network failed", "height", height, "error", err)
-			break
```
jchappelow, line 122: This break was allowing caller logic to proceed in bad cases like db failure from the requester's apply block component.  (it's more than just a requester)

node/consensus/engine.go
```diff
@@ -431,14 +431,21 @@ func (ce *ConsensusEngine) runConsensusEventLoop(ctx context.Context) error {
 	// The catchupticker resets with each processed consensus message that successfully advances the node's state
 
 	for {
+		select { // ignore other ready signals if we're shutting down
+		case <-ctx.Done():
+			ce.log.Info("Shutting down the consensus engine")
+			return
+		default:
+		}
```
jchappelow, line 439: The `select-case` prioritization problem that can cause important signals to be ignored indefinitely.

node/consensus/follower.go
```diff
@@ -98,7 +98,11 @@ func (ce *ConsensusEngine) AcceptCommit(height int64, blkID types.Hash, ci *type
 		// check if this is the first time we are hearing about this block and not already downloaded it.
 		blk, _, err := ce.blockStore.Get(blkID)
 		if err != nil {
-			return true
+			if errors.Is(err, types.ErrNotFound) || errors.Is(err, types.ErrBlkNotFound) {
+				return true // we want it
+			}
+			ce.log.Error("Unexpected error getting block from blockstore", "error", err)
+			return false
```
jchappelow, line 105: We don't want to start massive download operations because the block store is down.

node/consensus/follower.go
```diff
@@ -210,7 +214,7 @@ func (ce *ConsensusEngine) processBlockProposal(ctx context.Context, blkPropMsg
 
 	signature, err := types.SignVote(blkPropMsg.blkHash, true, &ce.state.blockRes.appHash, ce.privKey)
 	if err != nil {
-		ce.log.Error("Error signing the voteInfo", "error", err)
+		ce.log.Error("Error signing the voteInfo", "error", err) // still send the ack with this????
```
jchappelow, line 217: I don't know if there's a reason to send and ack with a bad signature? @charithabandi OK to make this fatal here?

node/consensus/follower.go
```diff
@@ -340,11 +346,11 @@ func (ce *ConsensusEngine) processAndCommit(ctx context.Context, blk *ktypes.Blo
 
 	// Commit the block if the appHash and commitInfo is valid
 	if err := ce.acceptCommitInfo(ci, blkID); err != nil {
-		return fmt.Errorf("error validating commitInfo: %v", err)
+		return fmt.Errorf("error validating commitInfo: %w", err)
```
jchappelow, line 349: Important to wrap errors, else we can't tell what the cause was, which is important up the call stack. I made the same mistake in `pg` pkg.

node/pg/db.go
```diff
@@ -221,7 +221,7 @@ func NewDB(ctx context.Context, cfg *DBConfig) (*DB, error) {
 		<-db.repl.done      // wait for capture goroutine to end (broadcast channel)
 		cancel(db.repl.err) // set the cause
 
-		db.pool.Close()
+		// db.pool.Close()
```
jchappelow, line 224: afaict, this was a mistake to put here as the `cancel` will shut it down anyway, and forcibly closign here prevents clean shutdown of higher level systems.

node/pg/query.go
```diff
@@ -236,10 +242,13 @@ func queryTx(ctx context.Context, oidToDataType map[uint32]*datatype, dbTx txBeg
...
 
-func queryRowFunc(ctx context.Context, conn *pgx.Conn, sql string,
+func queryRowFunc(ctx context.Context, conn *pgx.Conn, stmt string,
 	scans []any, fn func() error, args ...any) error {
-	rows, _ := conn.Query(ctx, sql, args...)
+	rows, _ := conn.Query(ctx, stmt, args...)
 	_, err := pgx.ForEachRow(rows, scans, fn)
+	if sql.IsFatalDBError(err) {
+		err = errors.Join(err, sql.ErrDBFailure)
+	}
```
jchappelow, line 251: These are pretty critical in the context of user queries.  We had a very bad non-determinism bug some time ago where transaction's nested transactions had an error (memory or something), and because it wasn't recognized as an actual postgres backend failure, which did not crash the actual postgres process, the node would proceed as if it were user error ...

brennanjl (dismissed):
a few comments but overall lgtm

node/block_processor/processor.go
```diff
@@ -344,11 +341,21 @@ func (bp *BlockProcessor) ExecuteBlock(ctx context.Context, req *ktypes.BlockExe
...
 	if err != nil {
-		bp.consensusTx = nil // safety measure
 		return nil, fmt.Errorf("failed to begin the consensus transaction: %w", err)
 	}
+	bp.consensusTx = tx
+
+	var success bool
+	defer func() {
+		if !success {
+			if err := bp.consensusTx.Rollback(context.Background()); err != nil {
```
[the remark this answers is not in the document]
jchappelow: This BP method only does precommit, not commit, which happens in a different method later when the CE is ready.  So if we defer unconditional rollback in here it will rollback the prepared transaction that we have **pre**committed, but not yet actually committed.  The context is a bit hard to track.

go.mod
```diff
@@ -213,4 +213,4 @@ require (
 	google.golang.org/protobuf v1.35.2 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.3.0 // indirect
-)
\ No newline at end of file
+)
```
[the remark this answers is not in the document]
jchappelow: Totally confused me too.  I went back and did a tidy and no change. LOl.  Will see what this is about.

node/consensus/follower.go
charithabandi: Yeah, it should return with an Error, I might have missed that.

go.mod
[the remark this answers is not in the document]
jchappelow: OH, the EVM sync PR removed the newlines at the end of the file https://github.com/kwilteam/kwil-db/commit/c812c4745546b354aec99a9050f299746a9e2b38#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L216
`task tidy` seems to have put them back.  Out of curiosity does your go toolchain remove them?

node/pg/db.go
jchappelow: Will remove this comment.  Pretty sure it is not correct now.

go.mod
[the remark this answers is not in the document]
charithabandi: mine doesn't

charithabandi (dismissed):

charithabandi (approved):

PR #1325, opened 2025-01-31
setup genesis,init edges, concise leader/validator population, db oth…

jchappelow:
Updates corresponding to node and admin docs updates in https://github.com/kwilteam/docs/pull/168

### DIFF
--- a/app/key/cmd.go
+++ b/app/key/cmd.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/kwilteam/kwil-db/app/shared/display"
 	"github.com/kwilteam/kwil-db/core/crypto"
+	"github.com/kwilteam/kwil-db/core/crypto/auth"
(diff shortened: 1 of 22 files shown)


jchappelow:
Again this corresponds to https://github.com/kwilteam/docs/pull/168
I expect there will be more docs revisions and some more final cmd cleanup when we do the final pass through docs after migrations are updated and retested.

jchappelow:
I think this is mostly uncontroversial, but one notable change is that wherever there is a `--leader` flag, I've made that required rather than inferring it is the first validator.  So the behavior is switched since you can just specify leader and it will make that a validator, but you can't do the other way around.  I felt a more conservative approach to indicating the leader explicitly was safer.  The two commands that use this are now simplified with a shared helper.

brennanjl (approved):

PR #1328, opened 2025-01-31
engine: unexport unused constants and functions

brennanjl:
I noticed that there were some values in the engine that were exported unnecessarily, and could be moved internally. These were mostly constants relating to operations (e.g. arithmetic, unary, comparison) and privileges.


jchappelow (approved):
LGTM.  @Yaiba ?

Yaiba (approved):
LGTM.  Yeah those are only needed internally by engine, no need to expose them.

PR #1330, opened 2025-01-31
Validator updates in the commitInfo

charithabandi:


jchappelow (dismissed):
Looks good, just a couple thoughts:

could encode key type in the integer format.  we do this in places:

```
	kd, ok := KeyTypeDefinition(kt)
	if !ok {
		return 0, fmt.Errorf("invalid key type: %s", kt)
	}
	err := binary.Write(cw, binary.LittleEndian, kd.EncodeFlag())
```

Could put version in the serialization, unless you feel the parent struct's version (commitInfo) covers that base?  Arguably does

core/types/types.go
```diff
@@ -128,6 +128,45 @@ func (v *Validator) String() string {
 	return fmt.Sprintf("Validator{pubkey = %x, keyType = %s, power = %d}", v.Identifier, v.KeyType, v.Power)
 }
 
+func (v *Validator) MarshalBinary() ([]byte, error) {
+	buf := new(bytes.Buffer)
+	if err := WriteBytes(buf, v.Identifier[:]); err != nil {
```
jchappelow, line 133: Oh, hmm, `Validator` embeds `AccountID`, which defines its own marshalBinary.  Should probably update that and use it.

charithabandi:
@jchappelow right now the param updates are not included in the Header, some kind of hash to verify that the leader is working with the correct consensus parameters. I will try to add that. Do you think anything else is missing that should be part of the header, which is not.

jchappelow:
> @jchappelow right now the param updates are not included in the Header, some kind of hash to verify that the leader is working with the correct consensus parameters. I will try to add that. Do you think anything else is missing that should be part of the header, which is not.

Nothing else comes to mind.

jchappelow (approved):
LGTM.  Any changes?  Did you want to do the params hash in the header now?  Different PR would be fine / better imo.

PR #1331, opened 2025-01-31
Genesis checks shouldn't involve migration status

charithabandi:
fixes the node start issue:

```
❯ ./.build/kwild start -r .net2
Error: genesis configuration failed sanity checks: migration status is invalid: ""
❯
```

### DIFF
--- a/app/node/node.go
+++ b/app/node/node.go
@@ -122,6 +122,13 @@ func runNode(ctx context.Context, rootDir string, cfg *config.Config, autogen bo
 		}
 	}
 
+	// migration status defaults
+	if genConfig.Migration.IsMigration() {
+		genConfig.MigrationStatus = types.GenesisMigration
+	} else {
+		genConfig.MigrationStatus = types.NoActiveMigration
+	}
+
 	if err := genConfig.SanityChecks(); err != nil {
 		return fmt.Errorf("genesis configuration failed sanity checks: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


jchappelow (approved):

jchappelow:
Should it then set it to NoMigrationActive on load then?

charithabandi:
Yeah, I am testing out to see what block processor is using, give me 2 mins.

jchappelow:
Was recently working on these methods, and it seemed trouble:

```
func (ms MigrationStatus) NoneActive() bool {
	return ms == NoActiveMigration || ms == ""
}

func (ms MigrationStatus) Active() bool {
	return !ms.NoneActive()
}

func (ms MigrationStatus) Valid() bool {
	switch ms {
	case NoActiveMigration, ActivationPeriod, MigrationInProgress,
		MigrationCompleted, GenesisMigration:
		return true
	default:
		return false
	}
}
```

The allowance for `""` in `NoneActive` was to match the conditions that were previously sprinkled through the code, but in `Valid` it cannot be `""`.

Best to ensure it's set on load I think.

jchappelow:
I'll never be able to find this PR again, so linking to where we had some logic for this before in a different spot: https://github.com/kwilteam/kwil-db/pull/1245/files#diff-aa6b286ff0444a43911cce3584a811e1c7c1858fff4d2780a548363cf2289ef7L141-L144
Just to be sure we have all bases covered, including the geneisMigration state

app/node/node.go
```diff
@@ -122,6 +122,13 @@ func runNode(ctx context.Context, rootDir string, cfg *config.Config, autogen bo
 		}
 	}
 
+	// migration status defaults
+	if genConfig.Migration.IsMigration() {
+		genConfig.MigrationStatus = types.GenesisMigration
+	} else {
+		genConfig.MigrationStatus = types.NoActiveMigration
```
jchappelow, line 129: Ah, this makes sense.  Thx.

jchappelow (approved):

PR #1332, opened 2025-01-31
ext(erc20rw): support list wallet rewards

Yaiba:
This pr support a new extension method "list_wallet_rewards", which returns `["chain","chain_id","contract","etherscan","created_at","param_recipient","param_amount","param_block_hash", "param_root", "param_proofs"]`;  to support this I created a table `erc20rw_recipient_reward` when finalize a epoch.

Also rename :
- `erc20rw_pending_reward` table to `erc20rw_rewards`
- `erc20rw_epoch_reward` table to `erc20rw_epochs`
- `erc20rw_pending_signatures` table to `erc20rw_epoch_votes`

### DIFF
--- a/contrib/docker/kwild.dockerfile
+++ b/contrib/docker/kwild.dockerfile
@@ -18,10 +18,10 @@ RUN chmod +x /app/dist/kwild /app/dist/kwil-cli
 
 FROM ubuntu:24.04
 WORKDIR /app
-RUN mkdir -p /var/run/kwil && chmod 777 /var/run/kwil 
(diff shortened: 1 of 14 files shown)


brennanjl (dismissed):

Yaiba:
Update: 
- remove the `erc20rw_` prefix from all tables in reward ext
- update dockerfile so `start` is not the default subcmd to run, so we can use the image to call other subcmd, for example: `setup ...`

brennanjl (approved):

PR #1333, opened 2025-01-31
use lrucache for account store records

charithabandi:


jchappelow (commented):
Looks good, except I think this needs tests to ensure the lru params are having the desired effect.  Probably that means a new type to isolate, and use for the records, but that might be overkill as far as the product goes.

jchappelow (approved):

PR #1337, opened 2025-02-03
rework cmd app errors, permit non-zero exit code and custom output

jchappelow:
This changes how the command line apps handle errors to ensure we have non-zero exit codes to the OS/shell while also maintaining our own output format.  This is important for scripting among other things.

The problem with cobra is that if we return an error from say `RunE`, [cobra will insist on prefixing the error message](https://github.com/spf13/cobra/issues/2226) with a text like `Error: ` which is no good for reasons.  We can sorta set this prefix, but it's a somewhat flawed implementation whereby we always end up with at least two spaces prefixing the output we want.  This is why we have had the `shared/display.PrintErr` always eating the error, I believe.

We get around this issue by continuing to swallow the error, but storing the error in the command's context, and then in `main()` pulling it out and setting the exit code if it is not nil.  (we don't actually need the `error`, just a bool, but I include the whole thing for debugging).

Finally, this PR also resolves related issues with startup and shutdown:
- the `buildServer` function was receiving the background context (my bad) so we couldn't really cancel things like a peer dial that would have to timeout instead
- the `P2Pservice` initializer used a background context
- the bootnode loop in `(*P2PService).Start` always returned nil even if the context was cancelled and a peer connect returned a non-nil error.  This was logically done to allow the loop to `continue` to other peers, but if the context was cancelled, we need to abort startup.  Before this fix, it would go on to initialize other systems, eventually hitting one that cared about the context cancellation.

### DIFF
--- a/app/key/info.go
+++ b/app/key/info.go
@@ -68,11 +68,6 @@ func InfoCmd() *cobra.Command {
 			cmd.Usage()
 			return errors.New("must provide with the private key file or hex string")
 		},
-		// PostRunE: func(cmd *cobra.Command, args []string) error {
-		// 	cmdP := cmd.Parent()
-		// 	cmdP.SetContext(cmd.Context())
-		// 	return nil
-		// },
 	}
 
 	cmd.Flags().StringVarP(&privkeyFile, "key-file", "o", "", "file containing the private key to display")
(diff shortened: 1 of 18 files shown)


brennanjl (approved):
Nice, looks like a good fix, thanks for doing this.

Will let you merge so that you can wait for your integration tests to finish

charithabandi (approved):

PR #1341, opened 2025-02-04
node: use bootpeers in mocknet test instead of force connect

jchappelow:
This makes `TestDualNodeMocket` use bootpeers like real apps rather than having the mocknet do it's `ConnectAllButSelf` magic.  Let's see if CI and my machine like this better.

More generally, we've started doing much more outside of the `Node` constructor and `Start`, so it might make sense to have these quick in-memory tests in some other node_test package that uses the server construction stuff in `app/node` directly so this is closer to reality.  This test came about before there was even an app using `Node`.

### DIFF
--- a/node/node_live_test.go
+++ b/node/node_live_test.go
@@ -5,6 +5,7 @@ package node
 import (
 	"context"
 	"encoding/hex"
+	"net"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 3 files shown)


node/node_live_test.go
```diff
@@ -381,7 +385,8 @@ func TestDualNodeMocknet(t *testing.T) {
 	}
 	ps2, err := NewP2PService(ctx, psCfg2, h2)
 	require.NoError(t, err)
-	err = ps2.Start(ctx)
+	err = ps2.Start(ctx, peerStr1)
```
jchappelow, line 388: ^ The bootpeers change here is having p2p service number 2 connect to the host number 1's p2p service.

I still don't get what happens in the CI failures, but it complains about missing protocol IDS, which is strange since all the protocol ids should have been set long before the nodes were linked and connected.

charithabandi (approved):

jchappelow:
Will merge this after the beta.2

PR #1345, opened 2025-02-05
leader election

charithabandi:
This PR introduces a way to update leader if the current leader is offline and the chain is halted because of offline leader.

`kwild validators promote <leaderCandidate> <height>`
In scenarios where leader is offline, there are multiple ways you can recover the chain. 

1. Fix the leader and bring the leader back up. 
2. If leader is not fixable or network doesn't want the previous leader, network should decide on the next leader and at least majority of the validators should promote the new candidate at the halt height using the above command for the new leader to take over and start proposing blocks. If majority of validators don't promote the same leader, the chain can end up in a broken state where different validators treat different nodes as the leader, so this command should be used with caution and to be used in scenarios when the chain is halted due to leader being offline and other methods (using param update resolutions) for changing the leader.

### DIFF
--- a/app/validator/promote.go
+++ b/app/validator/promote.go
@@ -1,0 +1,78 @@
+package validator
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/kwilteam/kwil-db/app/rpc"
+	"github.com/kwilteam/kwil-db/app/shared/display"
+	"github.com/kwilteam/kwil-db/config"
+	"github.com/spf13/cobra"
+)
+
+var (
+	promoteExample = `# Promote a new leader starting from block height 1000
+kwild validators promote 02c57268fc884fa88425c7e5c19d3af263d1c64dd8b8f3f8c0fb31bb622d1fdab8#secp256k1 1000`
+
+	promoteLong = `This command promotes a new leader starting from the specified block height. If the majority of validators agree, the new leader will be promoted from the given height. It is crucial for the validator being promoted to also promote themselves, otherwise, they will not propose a block when required. The specified height must not be an already committed block height, or the command will be rejected. This command will create or update the leader-updates.json file in the node's root directory, and when these updates are applied by the node at the specified height, the file will be deleted.`
+)
+
+func promoteCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "replace-leader <candidate> <height>",
+		Short:   "Promote a validator to leader starting from the specified height.",
+		Long:    promoteLong,
+		Example: promoteExample,
+		Args:    cobra.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+
+			clt, err := rpc.AdminSvcClient(ctx, cmd)
+			if err != nil {
+				return display.PrintErr(cmd, err)
+			}
+
+			pubKey, keyType, err := config.DecodePubKeyAndType(args[0])
+			if err != nil {
+				return display.PrintErr(cmd, err)
+			}
+
+			height, err := strconv.ParseInt(args[1], 10, 64)
+			if err != nil {
+				return display.PrintErr(cmd, err)
+			}
+
+			err = clt.Promote(ctx, pubKey, keyType, height)
+			if err != nil {
+				return display.PrintErr(cmd, err)
+			}
+
+			return display.PrintCmd(cmd, promoteStatus{
+				Height: height,
+				PubKey: args[0],
+			})
+		},
+	}
+
+	return cmd
+}
+
+type promoteStatus struct {
+	Height int64  `json:"height"`
+	PubKey string `json:"pubkey"`
+}
+
+func (ps promoteStatus) MarshalText() ([]byte, error) {
+	return []byte(fmt.Sprintf("Node will start accepting proposals from %s from height %d.\nThe new leader and majority of the validators must run this command for the leader replacement to be successful.", ps.PubKey, ps.Height)), nil
+}
+
+func (ps promoteStatus) MarshalJSON() ([]byte, error) {
+	return json.Marshal(struct {
+		Height int64  `json:"height"`
+		PubKey string `json:"pubkey"`
+	}{
+		Height: ps.Height,
+		PubKey: ps.PubKey,
+	})
+}
--- a/node/consensus/engine.go
+++ b/node/consensus/engine.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -85,6 +86,13 @@ type ConsensusEngine struct {
 	longRunningTxs []ktypes.Hash
 	numResets      int64
 
+	// leader updates, tracks if any non-transaction based leader updates that need to be
+	// applied at configured heights. These updates are to be applied before the
+	// block at height is proposed or after the height-1 block is committed.
+	leaderUpdates *leaderUpdate
+	leaderMtx     sync.RWMutex
+	leaderFile    string // file to persist the leader updates and load from on startup
+
 	// Channels
 	newBlockProposal chan struct{} // triggers block production in the leader
 	// newRound triggers the start of a new round in the consensus engine.
@@ -127,8 +135,16 @@ type ConsensusEngine struct {
 	catchupTimeout time.Duration
 }
 
+type leaderUpdate struct {
+	// Candidate is the new leader candidate
+	Candidate crypto.PublicKey
+	// Height is the height at which the leader update should be applied
+	Height int64
+}
+
 // Config is the struct given to the constructor, [New].
 type Config struct {
+	RootDir string
 	// Signer is the private key of the node.
 	PrivateKey crypto.PrivateKey
 	// Leader is the public key of the leader.
@@ -240,6 +256,13 @@ type state struct {
 	votes map[string]*types.VoteInfo
 
 	commitInfo *types.CommitInfo
+
+	// Promoted leader uses these updates to distinguish the leader updates occurred due to
+	// "replace-leader" admin command vs the leader updates due to parameter updates
+	// and include the NewLeader field in the block header to notify the network about
+	// this leader update. This field is applicable only to the promoted leader
+	// and only on the block heights where the node became a new leader through the replace cmd.
+	leaderUpdate *leaderUpdate
 }
 
 type blockResult struct {
@@ -262,7 +285,7 @@ type lastCommit struct {
 }
 
 // New creates a new consensus engine.
-func New(cfg *Config) *ConsensusEngine {
+func New(cfg *Config) (*ConsensusEngine, error) {
 	logger := cfg.Logger
 	if logger == nil {
 		// logger = log.DiscardLogger // for prod
@@ -285,6 +308,8 @@ func New(cfg *Config) *ConsensusEngine {
 		broadcastTxTimeout:  cfg.BroadcastTxTimeout,
 		checkpoint:          cfg.Checkpoint,
 		db:                  cfg.DB,
+		leaderUpdates:       nil,
+		leaderFile:          config.LeaderUpdatesFilePath(cfg.RootDir),
 		state: state{
 			blkProp:  nil,
 			blockRes: nil,
@@ -328,7 +353,11 @@ func New(cfg *Config) *ConsensusEngine {
 		ce.proposeTimeout = defaultProposeTimeout
 	}
 
-	return ce
+	// load the leader updates from the file if any
+	if err := ce.loadLeaderUpdates(); err != nil {
+		return nil, fmt.Errorf("error loading leader updates: %w", err)
+	}
+	return ce, nil
 }
 
 func (ce *ConsensusEngine) Start(ctx context.Context, fns BroadcastFns, peerFns WhitelistFns) error {
@@ -353,6 +382,9 @@ func (ce *ConsensusEngine) Start(ctx context.Context, fns BroadcastFns, peerFns 
 	if err := ce.catchup(ctx); err != nil {
 		return fmt.Errorf("error catching up: %w", err)
 	}
+
+	// apply leader updates if any before starting the consensus event loop
+	ce.applyLeaderUpdates()
 
 	// Start the mining process if the node is a leader. Validators and sentry
 	// nodes are activated when they receive a block proposal or block announce msg.
@@ -458,12 +490,25 @@ func (ce *ConsensusEngine) runConsensusEventLoop(ctx context.Context) {
 			return
 
 		case <-ce.newRound:
+			// check if there are any leader updates to be made before starting a new round
+			ce.applyLeaderUpdates()
+
+			// if the node is not a leader, ignore the new round signal
+			if ce.role.Load() != types.RoleLeader {
+				continue
+			}
+
 			go ce.newBlockRound(ctx)
 
 		case <-ce.newBlockProposal:
 			params := ce.blockProcessor.ConsensusParams()
 			if params.MigrationStatus == ktypes.MigrationCompleted {
 				ce.log.Info("Network halted due to migration, no more blocks will be produced")
+				continue
+			}
+
+			if ce.role.Load() != types.RoleLeader {
+				continue
 			}
 
 			if err := ce.proposeBlock(ctx); err != nil {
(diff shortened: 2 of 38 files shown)


charithabandi:
Logs of follower switching to a leader with the admin command.
```
2025-02-06 10:29:02.907 [INF] CONS: Role updated {from=validator to=leader}
2025-02-06 10:29:02.907 [INF] CONS: Starting a new consensus round {height=150}
2025-02-06 10:29:03.935 [INF] CONS: Created a new block proposal {height=150 hash=65517d4e6de99129c66e63005c07acb42024e3740e93dfd9e934ed1bcac831bd}
2025-02-06 10:29:03.956 [INF] BP: AppState updates:  {prevAppHash=3775d24a5e59ff7f8357a1f551ce58d61d98280dff7bb98e2fcc3a746bdacf79 changesetsHash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 valUpdatesHash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 accountsHash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 txResultsHash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 paramUpdatesHash=df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119}
2025-02-06 10:29:03.957 [INF] CONS: Executed block {height=150 blockID=65517d4e6de99129c66e63005c07acb42024e3740e93dfd9e934ed1bcac831bd numTxs=0 appHash=e60668251231d5e4d3757ecb48a00e746ee24a7d748ebb715bbf4758fd679107}
2025-02-06 10:29:03.957 [INF] CONS: Waiting for votes from the validators {height=150 hash=65517d4e6de99129c66e63005c07acb42024e3740e93dfd9e934ed1bcac831bd}
2025-02-06 10:29:03.961 [INF] CONS: Consensus message received {type=vote sender=021072159608e8bfa10102cc74d3e1b533dfdf1904538a61de42811cc3066de014}
2025-02-06 10:29:03.961 [INF] CONS: Adding vote {height=150 blkHash=65517d4e6de99129c66e63005c07acb42024e3740e93dfd9e934ed1bcac831bd appHash=e60668251231d5e4d3757ecb48a00e746ee24a7d748ebb715bbf4758fd679107 sender=021072159608e8bfa10102cc74d3e1b533dfdf1904538a61de42811cc3066de014}
2025-02-06 10:29:03.961 [INF] CONS: Majority of the validators have accepted the block, proceeding to commit the block {height=150 hash=65517d4e6de99129c66e63005c07acb42024e3740e93dfd9e934ed1bcac831bd acks=2 nacks=0}
2025-02-06 10:29:03.964 [INF] CONS: Committed Block {height=150 hash=65517d4e6de99129c66e63005c07acb42024e3740e93dfd9e934ed1bcac831bd appHash=e60668251231d5e4d3757ecb48a00e746ee24a7d748ebb715bbf4758fd679107 updates={}}
2025-02-06 10:29:03.964 [INF] CONS: Announce committed block 150 65517d4e6de99129c66e63005c07acb42024e3740e93dfd9e934ed1bcac831bd {}
```

node/consensus/engine.go
```diff
@@ -524,17 +551,66 @@ func (ce *ConsensusEngine) handleConsensusMessages(ctx context.Context, msg cons
...
+		if ce.leaderUpdates.Height < ce.lastCommitHeight() {
+			ce.log.Warn("Stale leader update, clearing it", "height", ce.leaderUpdates.Height, "validator", hex.EncodeToString(candidate.Bytes()))
+			ce.leaderUpdates = nil
+			ce.saveLeaderUpdates(nil)
+			return
+		}
+
+		if ce.leaderUpdates.Height == ce.lastCommitHeight()+1 {
+			ce.state.mtx.Lock()
+			defer ce.state.mtx.Unlock()
```
jchappelow, line 592: Lock ordering here makes me a bit nervous.  outer lock in this method is leaderMtx, inner is state.mtx.  Sure this doesn't happen in reverse anywhere else.  We could get into deadlock if not careful about this.

node/consensus/engine.go
```diff
@@ -524,17 +551,66 @@ func (ce *ConsensusEngine) handleConsensusMessages(ctx context.Context, msg cons
...
+
+			// ensure that the candidate is a validator
+			if _, ok := ce.validatorSet[hex.EncodeToString(candidate.Bytes())]; !ok {
+				ce.log.Warn("Invalid leader update, candidate is not a validator", "candidate", hex.EncodeToString(candidate.Bytes()))
+				ce.leaderUpdates = nil
+				ce.saveLeaderUpdates(nil)
+				return
+			}
+
+			// umm, what if we are in blocksync mode? trying to catchup with the network, should we apply the leader update at this height, if we know that some other leader already committed a block at this height? only accept this update if not in the catchup mode?
```
jchappelow, line 602: I don't quite understand this comment.  If a node doing sync sees a committed block with a new leader set in the header, it would certainly apply the leader update.  I think I'm confused by "if we know that some other leader already committed a block at this height".  What does that mean?

node/consensus/engine.go
```diff
@@ -627,9 +703,18 @@ func (ce *ConsensusEngine) catchup(ctx context.Context) error {
 // updateRole updates the validator set and role of the node based on the final state of the validator set.
 // This is called at the end of the commit phase or at the end of the catchup phase during bootstrapping.
```
jchappelow, line 704: I don't think this comment is right.  `updateValidatorSetAndRole` is called at the end of each commit, and the only other place is initialization.  "the end of the catchup phase during bootstrapping" doesn't make sense to me

node/consensus/engine.go
```diff
@@ -524,17 +551,66 @@ func (ce *ConsensusEngine) handleConsensusMessages(ctx context.Context, msg cons
...
 
 	case *blockAnnounce:
+		preRole := ce.role.Load()
 		if err := ce.commitBlock(ctx, v.blk, v.ci); err != nil {
 			ce.log.Error("Error processing committed block announcement", "error", err)
 			return
 		}
 
+		// apply the leader updates if any
+		ce.applyLeaderUpdates()
```
jchappelow, line 561: So, `commitBlock` is a few lines up, where it  does `updateValidatorSetAndRole`, which does:

```go
	// update the leader
	prevLeader := ce.leader
	ce.leader = params.Leader
```

And now in `applyLeaderUpdates` it does `ce.leader = ce.leaderUpdates.Candidate`.

If I understand this correctly, the first deals with consensus param updates based leader change, while the second deals with an update arising from the "offline" field in the block header.  Is that it?

node/consensus/engine.go
```diff
@@ -524,17 +551,66 @@ func (ce *ConsensusEngine) handleConsensusMessages(ctx context.Context, msg cons
...
 	}
 
 }
 
+// check if any leaderUpdates are to be made before processing the next block
+func (ce *ConsensusEngine) applyLeaderUpdates() {
+	ce.leaderMtx.Lock()
+	defer ce.leaderMtx.Unlock()
+
+	if ce.leaderUpdates != nil {
```
jchappelow, line 580: Much better in this case to outdent.

```go
if ce.leaderUpdates == nil {
     return
}

...
```

node/consensus/engine.go
```diff
@@ -524,17 +551,66 @@ func (ce *ConsensusEngine) handleConsensusMessages(ctx context.Context, msg cons
...
+		candidate := ce.leaderUpdates.Candidate
+		// stale updates to be cleared right away
+		if ce.leaderUpdates.Height < ce.lastCommitHeight() {
+			ce.log.Warn("Stale leader update, clearing it", "height", ce.leaderUpdates.Height, "validator", hex.EncodeToString(candidate.Bytes()))
```
jchappelow, line 584: I'm curious if you saw this happen?  It would mean a bug correct? A fairly severe one it seems.

node/consensus/engine.go
```diff
@@ -524,17 +551,66 @@ func (ce *ConsensusEngine) handleConsensusMessages(ctx context.Context, msg cons
...
+			ce.leaderUpdates = nil
+			ce.saveLeaderUpdates(nil)
+			return
+		}
+
+		if ce.leaderUpdates.Height == ce.lastCommitHeight()+1 {
```
jchappelow, line 590: What does it mean if this condition is false?  That's the case where they set a leader update for the future, and we're hanging onto it in `ce.leaderUpdates` until we reach that height?
Also would outdent this and comment on the scenario.

node/consensus/engine.go
jchappelow: Also for what lines of code does this lock really need to be held?  I assume only for `ce.state.leaderUpdate = ...`

node/consensus/engine.go
```diff
@@ -955,3 +1048,108 @@ func (ce *ConsensusEngine) lastCommitHeight() int64 {
...
+
+	// ensure that the candidate to be promoted is a validator
+	// TODO: Is this something that we wanna enforce when trying to apply leader updates
+	// rather than when accepting the admin validators promote command, as the node
+	// might not have synced to the latest block yet, and thus might not have the
+	// apt representation of the validator set.
+	// _, ok := ce.validatorSet[hex.EncodeToString(candidate.Identifier)]
+	// if !ok {
+	// 	return fmt.Errorf("only current validators can be promoted to leader")
+	// }
```
jchappelow, line 1067: No, I don't think so based on our reasoning.  As long as it's checked when this height is reached (it is) and if the candidate is not in the set at that height then the update is ignored and not fatal.

node/consensus/engine.go
```diff
@@ -955,3 +1048,108 @@ func (ce *ConsensusEngine) lastCommitHeight() int64 {
...
+		Height:    height,
+	}
+
+	if err := ce.saveLeaderUpdates(update); err != nil {
+		return fmt.Errorf("error saving the leader update: %w", err)
+	}
+
+	ce.leaderMtx.Lock()
+	defer ce.leaderMtx.Unlock()
+	ce.leaderUpdates = update
```
jchappelow, line 1081: I've looked at every use of `saveLeaderUpdates` and I think the very end of that method should do this field update, which would also match `loadLeaderUpdates`.  Could call it `storeLeaderUpdates`

node/consensus/engine.go
```diff
@@ -955,3 +1048,108 @@ func (ce *ConsensusEngine) lastCommitHeight() int64 {
...
+	}
+
+	// load the leader updates from the file
+	data, err := os.ReadFile(ce.leaderFile)
+	if err != nil {
+		return fmt.Errorf("error reading the leader file: %w", err)
+	}
+
+	// unmarshal the leader updates
+	var lu leaderUpdate
```
jchappelow, line 1101: Weirdly placed declaration.  At the very end, you may as well just do 

```go
ce.leaderUpdates = &leaderUpdate{
    Candidate: candidate,
    Height: update.Height,
}
```

jchappelow (commented):
Still going through but want to provide some feedback.

node/consensus/engine.go
charithabandi: Yes correct

node/consensus/engine.go
charithabandi: Yeah, I removed this comment in the latest commit, that I am gonna push now. not a valid concern

node/consensus/engine.go
charithabandi: Yeah, we apply updates only when we reach the height. Yup, gonna outdent and push the changes soon

node/consensus/engine.go
charithabandi: done

node/consensus/engine.go
charithabandi: umm, the scenarios where I can see this getting applied is during block sync, where you tried to promote a new leader, but there is not enough majority and the old leader is still up and producing blocks, in that case, even if u have decided to use this new leader from this height, but as the block is already committed and approved by majority, you accept the old leader as leader and ignore the updates. Once the block sync is done, u clear out this file. 

So the only case where this admin command works is, if majority approved to the leader change for a specific height.

jchappelow:
Looking for edges and rough spots, I do the silliest thing possible:
- healthy network with two validators
- on the leader's admin RPC service, do `validators promote` for the follower validator's key
- command returns with success, but prints nothing, no mention of the `.testnet/node0/leader-updates.json` file
- height passes and nothing happens
- I restart leader and it doesn't mention the obsolete (height passed) leader-updates file.

So I think a few things could change:
1. have the `validators promote` command prints a report of what it did, at least the json file location, maybe suggest the option to shutdown and delete the file to abort.  Also mention that the file must be kept until the change, but may be removed after (right?)
2. when the height passes, log something?
3. document that the validator being promoted must also promote themselves otherwise they won't ever propose a block when needed
4. on start, print info about the leader update if the json file is found.  Perhaps suggest deleting it if the latest committed is already past this height?

Are these valid changes?  It's a little confusing just looking at the command help how the procedure should go.

charithabandi:
`leader-updates.json` is more of an internal implementation thing, so I didn't include it in the command output. But, I will add logging wherever necessary.

jchappelow:
Another thing I'm slightly concerned about is validator operators mixing up `validators approve` and `validators promote`.
- what is the worst that can happen if a validator uses `promote` instead of `approve`?
- should we rename `promote` to `replace-leader`?  Something that is harder to do accidentally seems best here.

charithabandi:
Haha, yeah that's probably best. I think the worst that can happen is, they will be not able to accept block proposals from the correct leader, but can accept and change back to the correct leader once they hear a  blockAnn msg.

jchappelow:
> `leader-updates.json` is more of an internal implementation thing, so I didn't include it in the command output. But, I will add logging wherever necessary.

I would argue that this is above an implementation thing for two reasons:
1. the only way to cancel/abort this command's effect is to stop and delete the file.
2. you can manually configure leader replacement without the `validators promote` command using this file, given a restart.  You may wish to bootstrap your node with this file from the start if you are on a stuck network and are resyncing your node.

app/validator/promote.go
```diff
@@ -9,11 +11,20 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	promoteExample = `# Promote a new leader starting from block height 1000
+kwild validators promote 02c57268fc884fa88425c7e5c19d3af263d1c64dd8b8f3f8c0fb31bb622d1fdab8#secp256k1 1000`
+
+	promoteLong = `This command promotes a new leader starting from the specified block height. If the majority of validators agree, the new leader will be promoted from the given height. It is crucial for the validator being promoted to also promote themselves, otherwise, they will not propose a block when required. The specified height must not be an already committed block height, or the command will be rejected. This command will create or update the leader-updates.json file in the node's root directory, and when these updates are applied by the node at the specified height, the file will be deleted.`
```
jchappelow, line 18: Is it also correct to say that if the previous leader is still online and does not promote themselves, they will continue to propose new blocks, and even those validators that decided to promote a different validator as the new leader, they will continue to follow the old leader.  And in this case, the whole promote process will need to be done again but at a future height (because their promotion expires / is cleaned out)?

jchappelow: We can save some of that for updates in the `docs` repo. I'm mainly just trying to be clear about how this will go if it is used in production.

app/validator/promote.go
```diff
@@ -37,9 +48,31 @@ func promoteCmd() *cobra.Command {
...
 	return cmd
 }
+
+type promoteStatus struct {
+	Height int64  `json:"height"`
+	PubKey string `json:"pubkey"`
+}
+
+func (ps promoteStatus) MarshalText() ([]byte, error) {
+	return []byte(fmt.Sprintf("Promoted new leader at height %d with pubkey %s", ps.Height, ps.PubKey)), nil
```
jchappelow, line 67: This message is also fine, but I think the takeaway is that their node will begin accepting block proposals from the specified leader at some certain height.  The new leader and other validators must also do the same.  If the leader fails to either go offline or also do the same promote, they will continue to propose blocks the promotion will fail.

node/consensus/messages.go
```diff
@@ -144,7 +144,7 @@ func (ce *ConsensusEngine) NotifyBlockCommit(blk *ktypes.Block, ci *types.Commit
 	go ce.sendConsensusMessage(&consensusMessage{
 		MsgType: blkCommit.Type(),
 		Msg:     blkCommit,
-		Sender:  ce.leader.Bytes(),
+		Sender:  ce.leader.Bytes(), // TODO: wrong
```
jchappelow, line 147: How to make it right?  Use the NewLeader field of the header if it is set?

node/consensus/leader.go
```diff
@@ -295,7 +304,10 @@ func (ce *ConsensusEngine) addVote(ctx context.Context, voteMsg *vote, sender st
 				// and let the leader propose a new block, and the validator can refute it
 				// until it catches up.
 			}
-			ce.newRound <- struct{}{} // signal ce to start a new round
+
+			if ce.role.Load() == types.RoleLeader {
+				ce.newRound <- struct{}{} // signal ce to start a new round
```
jchappelow, line 309: Are these now-conditional sends being stopped the fix for the old leader continuing to propose blocks?  It seems like `(*ConsensusEngine).proposeBlock` should also just not work if role is wrong.

jchappelow (approved):
This was working pretty well last test.  Just a few random comments and questions, but nothing that really troubles me.

node/consensus/messages.go
charithabandi: Yup

app/validator/promote.go
charithabandi: If the old leader is still online, it's now a matter of majority. If majority of the validators agree on the new leader and produce a block, the old leader and the validators that didn't agree on the promotion will be forced to accept the new leader. 
But if the majority of the validators are still supporting the old leader, then the old leader will succeed in producing the next block, and all the validators who promoted a new leader will have to accept the old leader, and they have to redo the promotion again at a future height and ensure that they have the majority to succeed. 
And if both the promoted and the old leader doesn't have a majority, the network halts

jchappelow (approved):

jchappelow:
Good to merge with revised squash commit to cover all of it.  There are a lot of new concepts here worth describing in the commit message.

PR #1346, opened 2025-02-05
feat(erc20signersvc): add erc20 reward signer svc

Yaiba:
This pr adds erc20 reward signer service to kwild's sub service.

Most important change is how erc20 ext process epoch. Now we have three types of epoch:
- confirmed epoch, epoch that has been confirmed on EVM
- finalized epoch, epoch that has been finalized, wait to be confirmed
- collecting epoch, epoch that current collect all new rewards

Except the first epoch, at any given time, there will only be two active epochs: finalized epoch and collecting epoch.
After an finalized epoch is confirmed, when Kwil propose a new epoch:
- the current collecting epoch will be finalized
- a new collecting epoch will be created


config/config.go
```diff
@@ -320,17 +327,18 @@ type Config struct {
...
+	DB            DBConfig                     `toml:"db" comment:"DB (PostgreSQL) related configuration"`
+	RPC           RPCConfig                    `toml:"rpc" comment:"User RPC service configuration"`
+	Admin         AdminConfig                  `toml:"admin" comment:"Admin RPC service configuration"`
+	Snapshots     SnapshotConfig               `toml:"snapshots" comment:"Snapshot creation and provider configuration"`
+	StateSync     StateSyncConfig              `toml:"state_sync" comment:"Statesync configuration (vs block sync)"`
+	Extensions    map[string]map[string]string `toml:"extensions" comment:"extension configuration"`
+	GenesisState  string                       `toml:"genesis_state" comment:"path to the genesis state file, relative to the root directory"`
+	Migrations    MigrationConfig              `toml:"migrations" comment:"zero downtime migration configuration"`
+	Checkpoint    Checkpoint                   `toml:"checkpoint" comment:"checkpoint info for the leader to sync to before proposing a new block"`
+	Erc20RWSigner Erc20RewardSignerConfig      `toml:"erc20_reward_signer" comment:"ERC20 reward signer service configuration"`
```
Yaiba, line 341: I'm not happy with this name, looking for suggestions

config/config.go
```diff
@@ -426,6 +434,35 @@ type Checkpoint struct {
 	Hash   types.Hash `toml:"hash" comment:"checkpoint block hash."`
 }
 
+type Erc20RewardSignerConfig struct {
+	Enable      bool           `toml:"enable" comment:"enable the ERC20 reward signer service"`
+	Targets     []string       `json:"targets" comment:"target reward ext alias for the ERC20 reward"`
+	PrivateKeys []string       `json:"private_keys" comment:"private key for the ERC20 reward target"`
+	SyncEvery   types.Duration `json:"sync_every" comment:"sync interval; a recommend value is same as the block time"`
```
Yaiba, line 441: Since rn the block time is determined by two consensus parameters, not easy for the signerSvc to generate a good value for this configuration. So I'll leave to the user.

config/config.go
```diff
@@ -308,6 +308,13 @@ func DefaultConfig() *Config {
 			Height: 0,
 			Hash:   types.Hash{},
 		},
+		Erc20RWSigner: Erc20RewardSignerConfig{
+			Enable:      false,
+			PrivateKeys: nil,
+			Targets:     nil,
+			// the reasonable value is the block time
+			SyncEvery: types.Duration(1 * time.Minute),
```
Yaiba, line 316: Technically we don't need this configuration, but it maybe not what user wants.
Since rn the block time is determined by two consensus parameters(proposeTimeout and emptyBlockTimeout), not easy for the signerSvc to generate a good value based on that two parameters for this configuration. So I'll leave to the user.

node/services/erc20signersvc/api.go
```diff
@@ -0,0 +1,142 @@
...
+	TotalRewards types.Decimal
+	//MtreeJson    string
+	RewardRoot []byte
+	SafeNonce  int64
+	SignHash   []byte // SignHash is Chain aware, it's from GnosisSafeTx
+	ContractID types.UUID
+	BlockHash  []byte
+	CreatedAt  int64
+	Voters     []string
+	Finalized  bool // TODO: so signerSvc can use this to see it can skip, or, in the SQL, we have option to only return not-finalized epoch
```
Yaiba, line 25: Will support this after Brennan make the changes to the ext.

config/config.go
```diff
@@ -426,6 +434,35 @@ type Checkpoint struct {
 	Hash   types.Hash `toml:"hash" comment:"checkpoint block hash."`
 }
 
+type Erc20RewardSignerConfig struct {
+	Enable      bool           `toml:"enable" comment:"enable the ERC20 reward signer service"`
+	Targets     []string       `json:"targets" comment:"target reward ext alias for the ERC20 reward"`
+	PrivateKeys []string       `json:"private_keys" comment:"private key for the ERC20 reward target"`
```
Yaiba, line 440: here I made this an array, so different 'signer' could be configured for different target

app/node/start.go
```diff
@@ -104,14 +104,14 @@ func StartCmd() *cobra.Command {
...
+		if !strings.HasPrefix(args[i], "--extensions.") {
 			return nil, fmt.Errorf("expected extension flag, got %q", args[i])
 		}
 		// split the flag into the extension name and the flag name
 		// we intentionally do not use SplitN because we want to verify
-		// there are exactly 3 parts.
-		parts := strings.Split(args[i], ".")
-		if len(parts) != 3 {
+		// there are at least 3 parts.
+		parts := strings.SplitN(args[i], ".", 3)
```
Yaiba, line 113: the 3rd part can have values like URL

node/services/erc20signersvc/eth.go
```diff
@@ -0,0 +1,226 @@
...
+	"context"
+	"fmt"
+	"math/big"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/samber/lo"
```
Yaiba, line 13: Initially I used this libaray for functional operations, but rn this is the only place it's used

Yaiba:
others worth mention:
- Add `erc20_reward_signer` config section to kwild's config, support multiple ext instance
- signer service retries if RPC service is not ready or the configured 'target' erc20 instance is not created
- erc20 ext stores temporary `vote` records on epoch, once an epoch is confirmed, those record will be deleted
- use multicall3 if possible, to reduce the rpc roundtrip, and get values from same block

config/config.go
brennanjl: Lets start calling it "erc20 bridge".

So `ERC20BridgeSignerConfig`

node/exts/erc20reward/meta_sql.go
```diff
@@ -435,3 +579,80 @@ func setVersionToCurrent(ctx context.Context, app *common.App) error {
...
+	wallet ethcommon.Address, pending bool, fn func(*Epoch) error) error {
+
+	query := `
+	{kwil_erc20_meta}SELECT e.id, e.created_at_block, e.created_at_unix, e.reward_root, e.reward_amount, e.ended_at, e.block_hash, e.confirmed, array_agg(v.voter) as voters, array_agg(v.amount) as amounts, array_agg(v.nonce) as nonces, array_agg(v.signature) as signatures
+	FROM epoch_rewards AS r
+	JOIN epochs AS e ON r.epoch_id = e.id
+	LEFT JOIN epoch_votes AS v ON v.epoch_id = e.id
+	WHERE recipient = $wallet AND e.instance_id = $instance_id AND e.ended_at IS NOT NULL` // at least finalized
+
+	// TODO: use this after SQL issue is fixed
```
Yaiba, line 633: Ah forgot about this, I'll make the changes

node/exts/erc20reward/meta_schema.sql
```diff
@@ -39,15 +40,18 @@ CREATE TABLE balances (
...
 	id UUID PRIMARY KEY,
     created_at_block INT8 NOT NULL, -- kwil block height
     created_at_unix INT8 NOT NULL, -- unix timestamp (in seconds)
     instance_id UUID NOT NULL REFERENCES reward_instances(id) ON UPDATE RESTRICT ON DELETE RESTRICT,
 	reward_root BYTEA UNIQUE, -- the root of the merkle tree of rewards, it's unique per contract
+    reward_amount NUMERIC(78, 0), -- the total amount in current epoch
     ended_at INT8, -- kwil block height
     block_hash BYTEA, -- the hash of the block that is used in merkle tree leaf, which is the last block of the epoch
-    confirmed BOOLEAN -- whether the epoch has been confirmed on chain
+    confirmed BOOLEAN NOT NULL DEFAULT FALSE -- whether the epoch has been confirmed on chain
```
brennanjl, line 54: I don't think we should use NOT NULL here. If the `confirmed` is null, then that means it is still collecting; it essentially gives us 3 states.

There are queries that use `WHERE confirmed IS NULL`, so we need to make sure to account for this.

brennanjl: Never mind, I see that you updated the query that uses this. I guess no strong opinion on this then.

node/exts/erc20reward/meta_sql.go
```diff
@@ -233,18 +238,20 @@ func createSchema(ctx context.Context, app *common.App) error {
 }
 
 // issueReward issues a reward to a user.
-func issueReward(ctx context.Context, app *common.App, epochID *types.UUID, user ethcommon.Address, amount *types.Decimal) error {
+func issueReward(ctx context.Context, app *common.App, instanceId *types.UUID, epochID *types.UUID, user ethcommon.Address, amount *types.Decimal) error {
 	return app.Engine.ExecuteWithoutEngineCtx(ctx, app.DB, `
 	{kwil_erc20_meta}UPDATE reward_instances
-	SET balance = balance - $amount;
+	SET balance = balance - $amount
+    WHERE id = $instance_id;
```
brennanjl, line 245: good catch

node/exts/erc20reward/meta_sql.go
```diff
@@ -346,49 +352,187 @@ func getRewardsForEpoch(ctx context.Context, app *common.App, epochID *types.UUI
...
-		rewards = append(rewards, &EpochReward{
+		return fn(&EpochReward{
 			Recipient: recipient,
 			Amount:    row.Values[1].(*types.Decimal),
 		})
+	})
+}
+
+// previousEpochConfirmed return whether previous exists and confirmed.
+func previousEpochConfirmed(ctx context.Context, app *common.App, instanceID *types.UUID, endBlock int64) (bool, bool, error) {
```
brennanjl, line 363: nit: can you name the returns so that it is clear what each one means? just because there are two bools

node/exts/erc20reward/meta_sql.go
```diff
@@ -435,3 +579,80 @@ func setVersionToCurrent(ctx context.Context, app *common.App) error {
...
+		confirmed = row.Values[0].(bool)
+		return nil
+	})
+
+	if err != nil {
+		return false, err
+	}
+
+	return confirmed, nil
+}
```
brennanjl, line 604: You should probably add an error here where it errors out if the epoch doesn't exist. I'd do this by simply adding a counter than increments on each callback function passed to the engine. If the counter is 0, then the epoch with that ID does not exist.

This seems more correct than simply returning false

node/exts/erc20reward/meta_extension.go
```diff
@@ -942,107 +925,290 @@ func init() {
...
+							if len(signature) != reward.GnosisSafeSigLength {
+								return fmt.Errorf("signature is not 65 bytes")
+							}
+
+							from, err := ethAddressFromHex(ctx.TxContext.Caller)
+							if err != nil {
+								return err
+							}
+
+							confirmed, err := epochConfirmed(ctx.TxContext.Ctx, app, epochID)
```
brennanjl, line 1099: Seems like there is a bug here.

Epochs have 3 stages:
- collecting
- finalized
- confirmed

Only an epoch in the "finalized" phase can be signed. But this check only ensures that "confirmed" epochs cannot be signed; it does _not_ protect against signers voting for a "collecting" epoch.

node/exts/erc20reward/meta_extension.go
```diff
@@ -68,6 +71,8 @@ var (
 	// so that we know how to decode them
 	logTypeTransfer       = []byte("e20trsnfr")
 	logTypeConfirmedEpoch = []byte("cnfepch")
+
+	mtLRUCache = lru.NewMap[[32]byte, []byte](rewardMerkleTreeLRUSize) // tree root => tree body
```
brennanjl, line 75: nice

config/config.go
```diff
@@ -442,6 +450,40 @@ type Checkpoint struct {
...
+type Erc20RewardSignerConfig struct {
+	Enable      bool           `toml:"enable" comment:"enable the ERC20 reward signer service"`
+	Targets     []string       `toml:"targets" comment:"target reward ext alias for the ERC20 reward"`
+	PrivateKeys []string       `toml:"private_keys" comment:"private key for the ERC20 reward target"`
+	EthRpcs     []string       `toml:"eth_rpcs" comment:"eth rpc address for the ERC20 reward target"`
+	SyncEvery   types.Duration `toml:"sync_every" comment:"sync interval; a recommend value is same as the block time"`
+}
+
+func (cfg Erc20RewardSignerConfig) Validate() error {
+	if (len(cfg.PrivateKeys) != len(cfg.Targets)) && (len(cfg.EthRpcs) != len(cfg.Targets)) {
```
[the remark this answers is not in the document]
brennanjl: It seems like we should have a structure like:

```go
type ERC20BridgeSignerConfig struct {
	PrivateKeyPath string `toml:"private_key_path" comment:"private key file path for the ERC20 reward target"`
	Chain          string `toml:"chain" comment:"chain name for the ERC20 reward target"`
}
```

which would be stored in the top-level node config as:

```go
type Config struct {
	// ...
	ERC20Bridge map[string]ERC20BridgeSignerConfig `toml:"erc20_bridge" comment:"ERC20 bridge signer service configuration"`
```

node/services/erc20signersvc/eth.go
```diff
@@ -0,0 +1,221 @@
+package signersvc
```
brennanjl, line 1: seems sort've weird to have a `services` package, since the erc20 signer service is in no way actually a service of any kind.

We probably should put both this package and the extension package into an `erc20-bridge` package

brennanjl: This will allow us to have one common place to store abis, manage a node's configuration of RPCs (which are used both by validators and signers), etc.

node/services/erc20signersvc/kwil.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (k *erc20rwExtApi) SetTarget(ns string) {
+	k.namespace = ns
+}
+
+func (k *erc20rwExtApi) InstanceInfo(ctx context.Context) (*RewardInstanceInfo, error) {
+	procedure := "info"
+	input := []any{}
+
+	res, err := k.clt.Call(ctx, k.namespace, procedure, input)
```
brennanjl, line 96: You essentially have this node calling itself, but that seems like a weird and unnecessary hack?

Why don't we just properly make it a new module that uses the engine?

config/config.go
```diff
@@ -311,6 +311,13 @@ func DefaultConfig() *Config {
 			Height: 0,
 			Hash:   types.Hash{},
 		},
+		Erc20RWSigner: Erc20RewardSignerConfig{
+			Enable:      false,
+			PrivateKeys: nil,
```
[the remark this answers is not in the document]
Yaiba: hmm, so all keys in one file?

config/config.go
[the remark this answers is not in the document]
Yaiba: yeah thought abut that, I think we should do it

node/exts/erc20reward/meta_extension.go
Yaiba: yeah you're right

node/services/erc20signersvc/kwil.go
```diff
@@ -0,0 +1,177 @@
...
+	RewardAmount   string
+	EndBlockHash   []byte
+	Confirmed      bool
+	Voters         []string
+	VoteAmounts    []string
+	VoteNonces     []int64
+	VoteSignatures [][]byte
+}
+
+type FinalizedReward struct {
```
brennanjl, line 38: It seems weird to doubly declare these. We should really consolidate the signer and extension logic.

node/services/erc20signersvc/kwil.go
Yaiba: yeah it becomes weird now the signerSvc could a reference to the engine, gonna change this

node/exts/erc20reward/named_extension.go
```diff
@@ -222,6 +206,103 @@ func init() {
...
+							{Name: "id", Type: types.UUIDType},
+							{Name: "start_height", Type: types.IntType},
+							{Name: "start_timestamp", Type: types.IntType},
+							{Name: "end_height", Type: types.IntType, Nullable: true},
+							{Name: "reward_root", Type: types.ByteaType, Nullable: true},
+							{Name: "reward_amount", Type: uint256Numeric, Nullable: true},
+							{Name: "end_block_hash", Type: types.ByteaType, Nullable: true},
+							{Name: "confirmed", Type: types.BoolType},
+							{Name: "voters", Type: types.TextArrayType, Nullable: true},
+							{Name: "vote_amounts", Type: types.TextArrayType, Nullable: true}, // cannot successful return uint256NumericArray, as empty value
```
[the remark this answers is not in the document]
Yaiba: an empty uint256NumericArray return decimal(0,0), I don't know how can set precision on nil decimal value

node/exts/erc20reward/meta_sql.go
Yaiba: yeah make sense

node/services/erc20signersvc/eth.go
Yaiba: I like that, will move

config/config.go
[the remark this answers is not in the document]
brennanjl: No, the user can specify many files paths just like how they specify many keys.

node/exts/erc20reward/named_extension.go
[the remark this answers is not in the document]
brennanjl: Ah I see... This is a shortcoming of the extension interface. I will fix this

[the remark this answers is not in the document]
brennanjl: @Yaiba see here: https://github.com/kwilteam/kwil-db/pull/1394. This should fix the issue

node/services/erc20signersvc/kwil.go
Yaiba: This is a little bit related to the `polling`, in that we kind want to avoid using RPC service.  We've agreed that for `polling` we don't want to react on every block, then I don't see much difference between using end_of_block and just use a ticker.

After a bit experiment, IMO, RPC might be the preferred(and easier) way to communicate with erc20 ext, reasons are:
- although we can use `engine` to call ext VIEW methods, but we cannot call methods(i.e., vote_epoch) that change the state(because it's not deterministic);  thus we need to use **two** sets of API handling to communicate with extension
- when signeSvc call `vote_epoch`, there is no easy way to call the RPC service method.  You need to construct the payload manually, well, if this is the case Idk why we don't just use client SDK to do this.

node/services/erc20signersvc/kwil.go
Yaiba: previously I define the return types in the extension, and use them in the method `return`, later the API could reuse that. Is this what you mean?

config/config.go
```diff
@@ -335,6 +341,7 @@ type Config struct {
 	GenesisState string                       `toml:"genesis_state" comment:"path to the genesis state file, relative to the root directory"`
 	Migrations   MigrationConfig              `toml:"migrations" comment:"zero downtime migration configuration"`
 	Checkpoint   Checkpoint                   `toml:"checkpoint" comment:"checkpoint info for the leader to sync to before proposing a new block"`
+	Erc20Bridge  ERC20BridgeConfig            `toml:"erc20_bridge" comment:"ERC20 bridge configuration"`
```
Yaiba, line 344: This config is used in two places:
- evm-sync extension
- erc20-bridge extension

There are few places like oracle still use the general extension config, which I'm not sure if we should move them under `erc20_bridge` cofiguration.

app/node/build.go
```diff
@@ -504,6 +508,33 @@ func buildConsensusEngine(_ context.Context, d *coreDependencies, db *pg.DB,
...
+		}
+		_ = emptyFile.Close()
+	}
+
+	state, err := signersvc.LoadStateFromFile(stateFile)
+	if err != nil {
+		failBuild(err, "Failed to load erc20 bridge signer state file")
+	}
+
+	rpcUrl := "http://" + d.cfg.RPC.ListenAddress
```
brennanjl, line 528: Will this always work? Are ways that a user could set their listen address where it may not?

If so, an alternative would be to give the signer access to the local mempool

app/node/build.go
```diff
@@ -504,6 +508,33 @@ func buildConsensusEngine(_ context.Context, d *coreDependencies, db *pg.DB,
 	return ce
 }
 
+func buildErc20RWignerMgr(d *coreDependencies) *signersvc.ServiceMgr {
```
brennanjl, line 511: Seems like a typo: `buildErc20RWignerMgr` -> `buildErc20RSignerMgr`

app/node/build.go
Yaiba: now you mentioned it, do we expected to configure a public ip in `ListenAddress`?

brennanjl: Yes

node/exts/erc20-bridge/erc20/meta_extension.go
```diff
@@ -942,107 +926,256 @@ func init() {
...
+							//id := inputs[0].(*types.UUID)
+							epochID := inputs[1].(*types.UUID)
+							nonce := inputs[2].(int64)
+							signature := inputs[3].([]byte)
+
+							if len(signature) != utils.GnosisSafeSigLength {
+								return fmt.Errorf("signature is not 65 bytes")
+							}
+
+							from, err := ethAddressFromHex(ctx.TxContext.Caller)
```
brennanjl, line 1058: Shouldn't we check the signature here?

Yaiba: I commented at line 1063, basically we don't have enough info to verify

Yaiba: I created #1399 for this

brennanjl: > I commented at line 1063

The issue seems to be that I cannot read.

node/exts/erc20-bridge/signersvc/kwil.go
```diff
@@ -39,121 +46,336 @@ type EpochReward struct {
...
+	return uint64(nonce), nil
+}
 
-	ers := make([]*Epoch, len(res.QueryResult.Values))
-	for i, v := range res.QueryResult.Values {
-		er := &Epoch{}
-		err = types.ScanTo(v, &er.ID, &er.StartHeight, &er.StartTimestamp, &er.EndHeight,
-			&er.RewardRoot, &er.RewardAmount, &er.EndBlockHash, &er.Confirmed, &er.Voters, &er.VoteNonces, &er.VoteSignatures)
+// execute mimics client.Client.Execute, without client options.
+func (k *signerClient) execute(ctx context.Context, namespace string, txSigner auth.Signer, action string, tuples [][]any) (types.Hash, error) {
```
Yaiba, line 312: still a bit hacky, if it's worth it we should implement a internal `TxSvcClient`. For now this is all I needed for signer service

node/exts/erc20-bridge/signersvc/kwil.go
```diff
@@ -39,121 +46,336 @@ type EpochReward struct {
...
-type erc20rwExtApi struct {
-	clt       *client.Client
-	namespace string
+type DB interface {
+	sql.ReadTxMaker
+	sql.DelayedReadTxMaker
 }
 
-var _ erc20ExtAPI = (*erc20rwExtApi)(nil)
+type signerClient struct {
```
Yaiba, line 75: Another way to implement signerClient is to call the `usersvc.Service`. This approach will have unnecessary wrap/unwrap the request/response types, should be easier to implement

jchappelow:
Sorry, we removed `types.ResultBroadcastTx` recently since it was modeled after a very odd cometbft type that conflated tx execution code with a broadcast error code (even before the tx is executed).  The returns are now a `types.Hash` and a `*types.TxResult`.  Changes were made in https://github.com/kwilteam/kwil-db/pull/1405

Yaiba:
> Sorry, we removed `types.ResultBroadcastTx` recently since it was modeled after a very odd cometbft type that conflated tx execution code with a broadcast error code (even before the tx is executed). The returns are now a `types.Hash` and a `*types.TxResult`. Changes were made in #1405

thanks for the reminder, I'll make the change accordingly

charithabandi (approved):

node/services/erc20signersvc/eth.go
jchappelow: Zero deps with it's release tag.  👍

Issue #1342, opened 2025-02-04
null values in postgres array break changeset encoding


jchappelow:
Given an array like `'{1, 2, null}'::int8[]`, changeset encoding in a `datatype`'s `SerializeChangeset` implementation is generally broken.  For example, in the integer array type, we do a `strconv.ParseInt`, which fails because the third element in that array is the string `"NULL"`.

Note that for scalar types, `convertPgxTuple` handles this fine because `col.DataType` will be `pglogrepl.TupleDataTypeNull` and we don't look at any value.  For arrays, it's not a null type, but there is a null value in the array.

Also, this slipped passed most of our tests until recently when we fixed the "schema filter" with the addition of a namespace manager.  So now schemas that were ignored in the changeset capturing code now hit this path.

I see the issue and will fix.

PR #1348, opened 2025-02-05
[WIP] interpreter,pg: fix NULL in array types

jchappelow:
Resolves https://github.com/kwilteam/kwil-db/issues/1342, which causes node halt when migrations are enabled as the changeset serialization causes an error.  WIP

### DIFF
--- a/node/engine/interpreter/interpreter_test.go
+++ b/node/engine/interpreter/interpreter_test.go
@@ -12,15 +12,23 @@ import (
 	"testing"
 
 	"github.com/kwilteam/kwil-db/common"
+	"github.com/kwilteam/kwil-db/core/log"
 	"github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/extensions/precompiles"
 	"github.com/kwilteam/kwil-db/node/engine"
 	"github.com/kwilteam/kwil-db/node/engine/interpreter"
 	"github.com/kwilteam/kwil-db/node/pg"
+	pgtest "github.com/kwilteam/kwil-db/node/pg/test"
 	"github.com/kwilteam/kwil-db/node/types/sql"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+func TestMain(m *testing.M) {
+	pg.UseLogger(log.New(log.WithName("PG"), log.WithFormat(log.FormatUnstructured),
+		log.WithLevel(log.LevelDebug)))
+	m.Run()
+}
 
 const (
 	defaultCaller    = "owner"
@@ -675,9 +683,7 @@ func Test_SQL(t *testing.T) {
 		},
 	}
 
-	db, err := newTestDB()
-	require.NoError(t, err)
-	defer db.Close()
+	db := newTestDB(t, nil, nil)
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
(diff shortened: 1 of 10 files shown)


brennanjl (approved):

jchappelow:
Note for posterity:

```
psql://postgres@127.0.0.1:5432/postgres 
# select unnest('{"hello",world,"NULL",NULL}'::text[]);
┌────────┐
│ unnest │
├────────┤
│ hello  │
│ world  │
│ NULL   │
│ ∅      │
└────────┘
(4 rows)

Time: 18.949 ms
psql://postgres@127.0.0.1:5432/postgres 
# select null;
┌──────────┐
│ ?column? │
├──────────┤
│ ∅        │
└──────────┘
(1 row)

Time: 1.642 ms
psql://postgres@127.0.0.1:5432/postgres 
# select "null";
ERROR:  column "null" does not exist
LINE 1: select "null";
               ^
Time: 1.757 ms
psql://postgres@127.0.0.1:5432/postgres 
# select 'null';
┌──────────┐
│ ?column? │
├──────────┤
│ null     │
└──────────┘
(1 row)

Time: 2.622 ms
```

Double quotes is the way of strings inside of the array notation, which is also the format in the logical replication message's `TupleColumn` `Data []byte` field.  But single quotes is the way in the context of scalar text values.

Thus when reading the logical replication `Data []byte`, for the `TEXT` type, special handling is required for nulls:
- `NULL` by itself not surrounded by quotes -> `nil`
- `"NULL"`  -> `[]byte("NULL")` (remove the quotes)
- all other strings -> pass through to `[]byte`

If we do not do this, when we deserialize the changeset `[]byte` it will end up with go `string` values still containing the quotes, which is a problem when we go to apply them like:

```
var val any

// func to deserialize into val here

tx.Execute(ctx, "INSERT INTO tbl (col) VALUES ($1)", val)
```

In above, to insert `NULL`, `val = nil`.  To insert a string `"NULL"`, `val = "NULL"`. 

```

func Test_NullStr(t *testing.T) {
	ctx := context.Background()
	db, err := NewPool(ctx, &cfg.PoolConfig)
	require.NoError(t, err)
	defer db.Close()

	tx, err := db.BeginTx(ctx)
	require.NoError(t, err)
	defer tx.Rollback(ctx)

	_, err = tx.Execute(ctx, `CREATE TABLE test (s text);`)
	require.NoError(t, err)

	var val any
	var valOut *string

	// nil string -> OK, it's NULL
	val = nil
	_, err = tx.Execute(ctx, "INSERT INTO test (s) VALUES ($1)", val)
	require.NoError(t, err)

	err = QueryRowFunc(ctx, tx, "SELECT s FROM test", []any{&valOut}, func() error { return nil })
	require.NoError(t, err)

	t.Log(valOut) // <nil>

	// string NULL -> OK, not really NULL
	val = `NULL`
	valOut = nil
	_, err = tx.Execute(ctx, "INSERT INTO test (s) VALUES ($1)", val)
	require.NoError(t, err)

	err = QueryRowFunc(ctx, tx, "SELECT s FROM test", []any{&valOut}, func() error { return nil })
	require.NoError(t, err)

	t.Log(*valOut) // NULL

	// string with quotes -> wrong
	val = `"NULL"`
	valOut = nil
	_, err = tx.Execute(ctx, "INSERT INTO test (s) VALUES ($1)", val)
	require.NoError(t, err)

	err = QueryRowFunc(ctx, tx, "SELECT s FROM test", []any{&valOut}, func() error { return nil })
	require.NoError(t, err)

	t.Log(*valOut) // "NULL"
}
```

So, when we get a `"NULL"` from pg logical replication `Data []byte`, that has to be the string `NULL` with the quotes removed.

PR #1358, opened 2025-02-07
stress tool for v0.10

jchappelow:
You must have to have the key to `db_owner` to deploy a test schema namespace, and use `-key` to specify (only secp256k1, and just the hex bytes, no key type).

When it makes a new namespace, take note of the printed namespace.  Then to run multiple instanecs, copy that randomly generated namespace and use that with a second instance with `-ns` and a different (or random key).

(previous stress tool just had each instance depoy a new db, but that's not really possible and more unless each instance uses the same key, and that causes horrible nonce issues)

Say the DB owner is set to an ethereum address with a known corresponding private key.  Start the stress tool with the `-key` flag using that key:

```
./stress -key 0610bb3b2c52c4c06e9ba7a2ce2a507f618c5ca7c93b7b3e7601a7f5182b7133
```

This will automatically create a new namespace in which it will start calling actions:

```
[INF] STRESS: act schema: deployed namespace "blahblahca4yfwzg"
```

(I'll improve that log).

Grab that "blahblahxnvxnqod" string and use it with a second instance using the `-ns` flag, but either a different key or no key (random):

```
./stress -ns blahblahuw4q1dd2
```


node/engine/parse/grammar/KuneiformParser.g4
```diff
@@ -347,6 +347,7 @@ insert_statement:
         | (select_statement)
     )
     upsert_clause?
+    (RETURNING sql_expr_list)?
```
jchappelow, line 350: @brennanjl with this change I was able to get a schema with `INSERT ... RETURNING id;` past antlr (it both deploys fine, and passes `kwil-cli utils parse`), but I end up with an engine/interpreter error if I try to `INSERT ... RETURNING id;` in an action.  The error indicated that it got mixed up counting the number of arguments in the inserted values.  I can reproduce the error (don't have it at present), but I figured you might know what's the issue.

To clarify, the use case is the fairly typical scenario where's there's an autogenerated column, such as a generated uuid or a serial in this simplified example:

```
kwild=# create table asdf.t (a serial, b int);
CREATE TABLE
Time: 3.836 ms

   
kwild=# insert INTO asdf.t (b) values (2) returning a;
 a 
---
 1
(1 row)

INSERT 0 1
Time: 1.732 ms

kwild=# insert INTO asdf.t (b) values (767) returning a;
 a 
---
 2
(1 row)
```

jchappelow:
Works fine for me as of Friday, so r4r, but pls don't merge yet as I need to clean up.

Will probably revert the Antlr change as it's really not integral to this work: https://github.com/kwilteam/kwil-db/pull/1358#discussion_r1947332012

node/engine/parse/grammar/KuneiformParser.g4
brennanjl: Yeah, there are additional changes that would need to be made throughout the engine:
- We need to update our own AST with the `RETURNING` clause (we have an intermediate representation that Antlr writes to).
- We need to add logic for re-generating Postgres SQL
- We need to add logic in the query planner to inform us of what we are expecting to receive, and how it needs to order it (since returns have to be deterministic, which they commonly are not in SQL DBs)

These are all relatively easy to do actually. The first two tasks are 15 minutes of work, but the query planner might be a couple hours / half day. That being said, I think this would be a nice feature; I've found myself wanting `RETURNING` before.

I've made an issue here as a reminder to add this: https://github.com/kwilteam/kwil-db/issues/1361

jchappelow:
Reverted the antlr changes.  Just the stress tool now.

core/client/README.md
```diff
@@ -336,8 +336,8 @@ if err != nil {
 }
 ```
 
-The `Call` method returns the data in the `core/types/client.Records`
-type. See the [godocs](https://pkg.go.dev/github.com/kwilteam/kwil-db/core/types/client)
+The `Call` method returns the data in the `core/client/types.Records`
```
jchappelow, line 339: I have a separate branch for this README.  ok to disregarde this file and the example's go mod.  Will revert in this PR.

test/stress/README.md
```diff
@@ -0,0 +1,9 @@
+# Stress tool
+
+## New stress tool plan
```
jchappelow, line 3: ugh, accidental add, will remove

jchappelow:
Iterating with @charithabandi .  Nothing big coming, just not ready to merge.

charithabandi (approved):
@jchappelow looks good to me!

charithabandi (approved):

PR #1363, opened 2025-02-11
engine: fix extension startup during migration

brennanjl:
Fixes an issue where the new erc20 extensions are unable to start during migration. Also fixes an issue where paths for genesis state are not expanded in `kwild setup init`

PR #1364, opened 2025-02-11
Fix deadlock

brennanjl:
Fixes several race conditions and deadlock issues in the erc20 extension.

It is based on https://github.com/kwilteam/kwil-db/pull/1363, so https://github.com/kwilteam/kwil-db/pull/1363 needs to go in first (cc @charithabandi, we don't need you to review this but it is blocked on your review of the other one)

The first commit is https://github.com/kwilteam/kwil-db/pull/1363, the second is what is actually relevant to this PR

### DIFF
--- a/node/exts/erc20reward/meta_extension.go
+++ b/node/exts/erc20reward/meta_extension.go
@@ -41,6 +41,7 @@ import (
 	evmsync "github.com/kwilteam/kwil-db/node/exts/evm-sync"
 	"github.com/kwilteam/kwil-db/node/exts/evm-sync/chains"
 	"github.com/kwilteam/kwil-db/node/types/sql"
+	"github.com/kwilteam/kwil-db/node/utils/syncmap"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


Yaiba (dismissed):
lgtm

node/exts/erc20reward/meta_extension.go
```diff
@@ -355,16 +366,26 @@ func init() {
 
 									err = setActiveStatus(ctx.TxContext.Ctx, app, &id, true)
 									if err != nil {
+										info.mu.RUnlock()
 										return err
 									}
+
+									info.mu.RUnlock()
+									info.mu.Lock()
```
jchappelow, line 374: Somethign is funny here.

jchappelow: oh I misread... still reviewing.  Just a sec

node/exts/erc20reward/meta_extension.go
```diff
@@ -355,16 +366,26 @@ func init() {
 
 									err = setActiveStatus(ctx.TxContext.Ctx, app, &id, true)
```
jchappelow, line 367: What I was trying to understand is if `setActiveStatus` can end up doing a full RW lock.  Is that a possibility?

brennanjl: Nope. The engine will need a read lock to copy the state (which it uses to maintain transactionality), but it won't need a RW lock unless one of those methods is called directly (which they are not, the extension never calls itself)

jchappelow (approved):
I think it'll work to avoid deadlock from a recursive rw lock.  It's a bit difficult to reason about; no better suggestions at present.

brennanjl:
>  It's a bit difficult to reason about

Yeah I don't love it but I don't really see another option right now

PR #1366, opened 2025-02-11
ce,node: CE crash must stop node

jchappelow:
The ensures that fatal CE errors (those which kill the consensus event loop other than parent context cancellation) are returned to the the caller of Start. This fixes a halted / dead CE not stopping the node.

This also reduces a lot of duplicate error logging when the error is continually passed up the call stack.


charithabandi (approved):
LGTM!

I will do audit of consensusMessage processing, looks like we are ignoring errors there today or tomorrow.

jchappelow:
@charithabandi Rebased and resolved.  Thanks for the reminder on this.

charithabandi (approved):

PR #1368, opened 2025-02-11
[WIP] common,bp,txapp: call notices and row results in log, and use log from tx routes

jchappelow:
JUST TESTING.

I'm experimenting with putting results from action execution in the transaction log:
- notices
- scanned rows from an action return, either a single value, single record/row, or a table

I think there are gotchas when a table is returned, but I like the possibility to have action returns go somewhere.

### DIFF
--- a/node/txapp/routes.go
+++ b/node/txapp/routes.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"strings"
 	"sync"
 	"time"
 
@@ -136,7 +137,7 @@ func (d *baseRoute) Price(ctx context.Context, router *TxApp, db sql.DB, tx *typ
 func (d *baseRoute) Execute(ctx *common.TxContext, router *TxApp, db sql.DB, tx *types.Transaction) *TxResponse {
 	dbTx, err := db.BeginTx(ctx.Ctx)
 	if err != nil {
-		return txRes(nil, types.CodeUnknownError, err)
+		return txRes(nil, types.CodeUnknownError, "", err)
 	}
 
 	spend, code, err := router.checkAndSpend(ctx, tx, d, dbTx)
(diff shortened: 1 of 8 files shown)


brennanjl:
> I think there are gotchas when a table is returned, but I like the possibility to have action returns go somewhere.

Curious, what might these be?

jchappelow:
> > I think there are gotchas when a table is returned, but I like the possibility to have action returns go somewhere.
> 
> Curious, what might these be?

I need to think it though, but it seems like if this is the case then it adds some considerations to the action author.  For instance, perhaps an action is best suited to serve other actions, so it returns largish tables, and the other actions return more reasonably (small) results.  So directly calling the action with a large table return ends up with a transaction with a large log.  Maybe modifiers can avoid that, or perhaps log length limits?

node/txapp/routes.go
```diff
@@ -314,19 +317,69 @@ func (d *executeActionRoute) PreTx(ctx *common.TxContext, svc *common.Service, t
...
-func (d *executeActionRoute) InTx(ctx *common.TxContext, app *common.App, tx *types.Transaction) (types.TxCode, error) {
+const unknownColName = "?column?" // TODO: get from interpreter pkg
+
+func (d *executeActionRoute) InTx(ctx *common.TxContext, app *common.App, tx *types.Transaction) (types.TxCode, string, error) {
+	var logBuilder strings.Builder
 	for i := range d.args {
-		// TODO: once we are able to store execution logs in the block store, we should propagate the discarded
-		// return value here.
-		_, err := app.Engine.Call(makeEngineCtx(ctx), app.DB, d.namespace, d.action, d.args[i], func(r *common.Row) error {
-			// we throw away all results for execute actions
```
jchappelow, line 322: @charithabandi Since you were asking, this is the very old TODO I was attempting to address.  There are two things we can put in the tx log: (1) the db tx "logs" which are emitted notices, and (2) returned data from the action.

My hesitation with this PR is twofold: (1) if the schema author is not cautious about the size of data returned and possibly modifiers (like owner or private), this could result in enormous txResults that bloat the chain and network traffic in commit info, and (2) what is the format and structure for the tx log (I'm just experimenting with a human readable format here).

jchappelow:
Closing in favor of https://github.com/kwilteam/kwil-db/pull/1439

PR #1369, opened 2025-02-11
kwild: re-add extension config flags

brennanjl:
It seems like we forgot to re-add the ability to [configure extensions from the command line](<https://prerelease.kwil.com/docs/daemon/config/settings#configuration-source-priority>). This re-adds that

### DIFF
--- a/app/node/start.go
+++ b/app/node/start.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -27,7 +28,6 @@ func StartCmd() *cobra.Command {
 		CompletionOptions: cobra.CompletionOptions{
 			DisableDefaultCmd: true,
 		},
-		Args:    cobra.NoArgs,
 		Version: version.KwilVersion,
 		Example: custom.BinaryConfig.NodeCmd + " start -r .testnet",
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


brennanjl:
Im open to other ways of doing it, this is just how we did it before (but we can definitely change it)

app/node/start.go
```diff
@@ -72,3 +72,54 @@ func StartCmd() *cobra.Command {
 
 	return cmd
 }
+
+// parseExtensionFlags parses the extension flags from the command line and
+// returns a map of extension names to their configured values
+func parseExtensionFlags(args []string) (map[string]map[string]string, error) {
```
jchappelow, line 78: I guess it was just a forgotten feature, my bad.  I think you need to call it in the `StartCmd` `RunE` still?

brennanjl: Right...

app/node/start.go
```diff
@@ -40,6 +40,10 @@ func StartCmd() *cobra.Command {
 
 			cfg := conf.ActiveConfig()
 
+			// we don't need to worry about order of priority with applying the extension
+			// flag configs because flags are always highest priority
+			cfg.Extensions = extConfs
```
jchappelow, line 45: I have an idea for this, wanna try something quick.  Otherwise I think the `--` solution is acceptable.

brennanjl: Ok. I don't love `--`, so I'm open to something else. I got it from the POSIX guidelines (guideline 10):https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html

jchappelow (approved):

PR #1371, opened 2025-02-12
mempool: limits, fix RecheckTx, etc.

jchappelow:
This adds a to a maximum bytes limit to the `Mempool`. 

`Store` now returns `(found, rejected bool)` where the first is the same (already there) and the second means it would push the mempool past its limit.

`Size` now returns both the tx count and total bytes.

`PeekN(maxTxns, totalSizeLimit int)` has the new second parameter which may further limit the number of transactions returned.  However, the caller could easily have truncated the slice since `PeekN` by definition does not modify `Mempool` state.

A bug in `RecheckTxs` is fixed.  It did not remove things correctly since it was ranging over a slice and calling a method, but that can only be done with a range on a map.  The range did not run over the full slice as intended.  Transactions were left that remained, and some could be removed that were not supposed to be.  New unit tests cover this.


node/consensus/leader.go
```diff
@@ -199,7 +199,8 @@ func (ce *ConsensusEngine) proposeBlock(ctx context.Context) error {
 // This method orders the transactions in the nonce order and also
 // does basic gas and balance checks and enforces the block size limits.
 func (ce *ConsensusEngine) createBlockProposal(ctx context.Context) (*blockProposal, error) {
-	nTxs := ce.mempool.PeekN(blockTxCount)
+	totalTxSizeLimit := ce.ConsensusParams().MaxBlockSize
+	nTxs := ce.mempool.PeekN(blockTxCount, int(totalTxSizeLimit))
```
charithabandi, line 203: update the blockTxCount to maxLimit?

charithabandi (approved):

PR #1373, opened 2025-02-13
planner: fix update and upsert bugs

brennanjl:
This PR fixes two bugs in the query planner that @Yaiba found.

The first bug occurred when we have an UPDATE that does not have a FROM nor WHERE, and references the updated table as part of the expression. It was unable to reference the updated table in this case.

The second bug was not technically a bug, but it was a really confusing error message. If a user had an ON CONFLICT clause targeting a composite index, and it referenced an unknown column, it would say that a unique index did not exist on those columns, rather than complaining that the column itself did not exist.


Yaiba (approved):
LGTM

PR #1375, opened 2025-02-13
fix null value bug in engine

brennanjl:
This fixes a bug found by @Yaiba, where a nil array was not detected as nil, but instead detected as a 0-length array, if it was passed via an extension.


jchappelow (approved):
LGTM, but @Yaiba knows for sure.

Issue #1347, opened 2025-02-05
Bug: Badger cache issue when starting node


KwilLuke:
Chain was running for awhile (reach block ~8k). Then hit this issue:

```
./kwild start -r ./node0
2025-02-04 18:25:00.137 [INF] KWILD: Starting kwild version 0.10.0-beta.2+e74fe75ae
2025-02-04 18:25:00.138 [INF] KWILD: Loading the genesis configuration from node0/genesis.json
2025-02-04 18:25:00.154 [INF] KWILD: Node public key: 0226b3ff29216dac187cea393f8af685ad419ac9644e55dce83d145c8b1af213bd (secp256k1)
2025-02-04 18:25:00.154 [INF] KWILD: loading TLS key pair for the admin server if TLS enabled {key_file=admin.key cert_file=admin.cert}
2025-02-04 18:25:00.222 [INF] PEERS: Loaded address book with 0 peers
badger 2025/02/04 18:26:00 WARNING: Block cache might be too small. Metrics: hit: 515 miss: 8441 keys-added: 262 keys-updated: 92 keys-evicted: 212 cost-added: 823565314 cost-evicted: 556016591 sets-dropped: 0 sets-rejected: 8086 gets-dropped: 0 gets-kept: 8896 gets-total: 8956 hit-ratio: 0.06
badger 2025/02/04 18:26:00 WARNING: Cache life expectancy (in seconds): 
 -- Histogram: 
Min value: 0 
Max value: 0 
Count: 212 
50p: 2.00 
75p: 2.00 
90p: 2.00 
[0, 2) 212 100.00% 100.00%
 --
```

jchappelow:
I want to look into this more in the future, but since you were able to startup after reboot, and I was able to replay the node from the root directory zip sent to me, I don't think we need to worry more about this for now.

```
2025-02-05 10:00:39.108 [INF] CONS: Replayed blocks from the blockstore {from=1 to=7980 elapsed=1m11.837747464s appHash=d91d65f376c054964386a11dcb499e4b57e796767e724f982c8f01259db09890}
```

Also ok after restarting kwild.

jchappelow:
Got while node was running, not starting, also didn't seem to cause a problem:

```
2025-02-11 14:40:17.928 [INF] CONS: Starting a new consensus round {height=15275}
2025-02-11 14:40:17.928 [WRN] NODE: No peers to advertise block to
[...]

PR #1376, opened 2025-02-14
consensus,store: enforce block size limits, compact serializations, misc badger fixes

jchappelow:
The PR starts with enforcing the consensus parameter (in genesis.json) for max block size.

This also led to several changes in block and transaction serialization, primarily the use of variable length integers in length-prefixed data, which includes many fields in the transaction and transaction body.  None of the payloads that JS would have to serialize are changed.  The size of a block with many fairly small transactions is drastically reduced.

The `SerializeSize` methods are added or improved, as these are required for efficient size checks without actually allocating huge amounts of memory.

Finally, this addresses some minor issues related to the block store:
- much of the config was not working as intended, namely the logger was being overriden
- our default of no-compression (an override from badger defaults) did not set badger's internal "block" cache to 0 (disabled) which the badger docs recommend if no compression or encryption are used
- expose a `store.compression` option

Related: https://github.com/kwilteam/kwil-db/issues/1347

### DIFF
--- a/app/node/build.go
+++ b/app/node/build.go
@@ -367,7 +367,7 @@ func schemaExists(ctx context.Context, db sql.Executor, schema string) (bool, er
 
 func buildBlockStore(d *coreDependencies, closers *closeFuncs) *store.BlockStore {
 	blkStrDir := config.BlockstoreDir(d.rootDir)
-	bs, err := store.NewBlockStore(blkStrDir)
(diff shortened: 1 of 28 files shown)


jchappelow:
Noticing some things to change.  Back to draft, but overall not much will change.

charithabandi (dismissed):

charithabandi (approved):

PR #1377, opened 2025-02-14
concurrency access on the mempool from txquery and recheckTxs

charithabandi:


jchappelow (approved):
LGTM.  It's an expensive stringer.  Should be used sparingly.  (if the caller already hash hash, construct log themselves).

jchappelow:
That said, the stringer is clear ok in rare uses.

jchappelow (approved):

PR #1378, opened 2025-02-14
node/mempool: fix RecheckTx again

jchappelow:
This fixes `node/mempool.(*Mempool).RecheckTx` so that it iterates in order.  The previous slice iteration was not working since it modified the slice during iteration, and the subsequent map iteration broke the caller's assumption that it would apply `CheckTx` in order


charithabandi (approved):
ok, Lgtm!

Issue #1355, opened 2025-02-07
Engine: notify user that an action cannot be used in a SQL statement


brennanjl:
If an action is used in a SQL statement, Postgres simply gives a "unknown function" error. With some minor modifications, our query planner can inform the user that it is an action

PR #1381, opened 2025-02-18
engine: better error messages for actions used in SQL

brennanjl:
Actions are not allowed to be used within SQL. If a user tries to, two things may happen:

1. If the action has the same name of the built-in function, it will just use the built-in function. This is actually sort've misleading from the perspective of the user, because referencing that same function in an action will call the action (and all built-in functions can be called this way as well).
2. If there is not a built-in function, it simply says "function does not exist", rather than informing the user that it is illegal to use actions in a query.

This PR changes this, so that the query planner will inform the user that it cannot reference these functions.

Resolves: https://github.com/kwilteam/kwil-db/issues/1355

### DIFF
--- a/node/engine/interpreter/context.go
+++ b/node/engine/interpreter/context.go
@@ -191,6 +191,19 @@ func (e *executionContext) query(sql string, fn func(*row) error) error {
 
 			return nil, engine.ErrUnknownVariable
 		},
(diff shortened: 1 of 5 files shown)


Yaiba (approved):
LGTM

PR #1387, opened 2025-02-18
[FOR TESTING] node,peers: modify block get request with chunk/idle timeout and informative not found

jchappelow:
This modifies the get-by-height block request to use a `readAll` implementation that allows setting timeouts for the 512 bytes chunks used in the read loop.  This should prevent having to hit the full (20 sec or longer) timeout for the entire response if the connection has completely stalled.

This also changes `ProtocolIDBlockHeight` protocol to include the peer's best height in both the "not found" branch of the protocol, and at the end of the positive path right after the block contents.  The former may be used to help block sync (e.g. break the catch up loop if all the not found responses say we are requesting the best known +1 or greater already); the latter could also help the CE be more intelligent about when it breaks out of initial block sync so it can get into the consensus event loop.


jchappelow:
Included in https://github.com/kwilteam/kwil-db/pull/1397

Issue #1383, opened 2025-02-18
Large integers returned to CLI in scientific notation


brennanjl:
A large int64 that is stored in the database is returned in scientific notation. I believe this is due to how the result is being JSON marshalled, but I am unsure:

```shell
$ kwil-cli query 'select 1000000000000000'
|       |
+-------+
| 1e+15 |
```

PR #1389, opened 2025-02-18
cli: fix scientific formatting from JSON

brennanjl:
resolves https://github.com/kwilteam/kwil-db/issues/1383

### DIFF
--- a/cmd/kwil-cli/cmds/call-action.go
+++ b/cmd/kwil-cli/cmds/call-action.go
@@ -2,8 +2,10 @@ package cmds
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	"github.com/kwilteam/kwil-db/app/shared/display"
 	"github.com/kwilteam/kwil-db/cmd/kwil-cli/client"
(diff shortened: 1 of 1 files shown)


cmd/kwil-cli/cmds/call-action.go
```diff
@@ -152,7 +154,20 @@ func getStringRows(v [][]any) [][]string {
 	for _, r := range v {
 		var row []string
 		for _, c := range r {
-			row = append(row, fmt.Sprintf("%v", c))
+			var col string
+			switch c2 := c.(type) {
+			case []byte:
+				col = base64.StdEncoding.EncodeToString(c2)
+			case float64:
```
jchappelow, line 161: Regarding the curretn slack convo.  Here's what I think we could decode with json.Number enabled and have a case for `json.Number`, then try `Int64()` fallback to `Float64()`.  However, I don't know how JavaScript clients will work.  A javascript `number` is a `float64` (effective).  So for consistency, it feels like integers (if they can be very large) should be strings in the JSON.

jchappelow (approved):

cmd/kwil-cli/cmds/call-action.go
brennanjl: This is a separate issue.

Regardless of what happened via JSON RPC, our CLI tooling would convert large numbers to scientific notation.  For example:

```
$ go run ./cmd/kwil-cli query 'select 1000000000000000'
|       |
+-------+
| 1e+15 |
```

JSON RPC returns the correct value here (1000000000000000), but the CLI formats it as scientific notation.

PR #1394, opened 2025-02-19
engine: add soft type casting for extension returns

brennanjl:
This fixes a weird edge case with the engine that we overlooked, where a null or 0-length value might not have the data type declared by the engine, causing unintuitive errors (or making it impossible to pass certain values, such as 0-length decimal arrays).

This is solved by adding a "soft cast", which will checks if a value is null or a zero length array of the expected type, and if so, casts it to the expected type


Yaiba (approved):
lgtm

PR #1397, opened 2025-02-19
Performance improvements in P2P and ConsensusEngine

charithabandi:
Performance improvements in the P2P and CE:
- updated the getBlockByHeight with chunk/idle timeouts and stream deadline checks to avoid long waits and return best height when the requested block is not found. Best height is used to guess if the block requested is not committed by the leader yet.
- CE now maintains a state which identifies if the block is available to the CE, if yes, the AcceptCommits will communicate with the node's stream handlers to not download the block again. 
- CE will release the block fetecher lock only upon the commit and the block is available in the blockstore
- BroadcastTx will now return ErrMempoolFull if the mempool is full
- Block validation now includes checks to ensure that the number of voteBody and voteID events are under the maxVotesPerTx limits.

### DIFF
--- a/node/block.go
+++ b/node/block.go
@@ -173,7 +173,8 @@ func (n *Node) blkAnnStreamHandler(s network.Stream) {
 	}
 
 	// re-announce
-	n.ce.NotifyBlockCommit(blk, ci)
+	n.log.Infof("downloaded block %v of height %d from %v, notifying ce of the block", blkid, height, s.Conn().RemotePeer())
+	n.ce.NotifyBlockCommit(blk, ci, blkHash)
 	go func() {
 		n.announceRawBlk(context.Background(), blkHash, height, rawBlk, blk.Header, ci, s.Conn().RemotePeer(), reqMsg.LeaderSig) // re-announce with the leader's signature
 	}()
--- a/node/block.go
+++ b/node/block.go
@@ -115,6 +115,9 @@ func (n *Node) blkAnnStreamHandler(s network.Stream) {
 	// If we are a validator and this is the commit ann for a proposed block
 	// that we already started executing, consensus engine will handle it.
 	if !n.ce.AcceptCommit(height, blkHash, hdr, ci, sig) {
+		// this either means that the ce already has the block or it is not
+		// ready to accept it yet.  In either case, we don't need to do anything
+		// here.
 		return
 	}
 
@@ -122,7 +125,6 @@ func (n *Node) blkAnnStreamHandler(s network.Stream) {
 	// code like a sentry node might do.
 
 	need, done := n.bki.PreFetch(blkHash)
-	defer done()
 	if !need {
 		n.log.Debug("ALREADY HAVE OR FETCHING BLOCK")
 		return // we have or are currently fetching it, do nothing, assuming we have already re-announced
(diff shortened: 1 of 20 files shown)


node/nogossip.go
```diff
@@ -50,7 +50,7 @@ func (n *Node) txAnnStreamHandler(s network.Stream) {
 	// First try to get from this stream.
 	rawTx, err := requestTx(s, []byte(getMsg))
 	if err != nil {
-		n.log.Warnf("announcer failed to provide %v, trying other peers", txHash)
+		n.log.Warnf("announcer failed to provide %v due to error %w, trying other peers", txHash, err)
```
jchappelow, line 53: `%w` only works in `fmt.Errorf`, not logging

node/block.go
```diff
@@ -121,51 +133,65 @@ func (n *Node) blkAnnStreamHandler(s network.Stream) {
 	n.log.Debug("retrieving new block", "blockID", blkid)
 	t0 := time.Now()
 
+	peerID := s.Conn().RemotePeer()
+
 	// First try to get from this stream.
 	rawBlk, err := request(s, []byte(getMsg), blkReadLimit)
 	if err != nil {
-		n.log.Warnf("announcer failed to provide %v, trying other peers", blkid)
+		n.log.Warnf("announcer failed to provide %v due to error: %w, trying other peers", blkid, err)
```
jchappelow, line 141: `%w` in logs here too

node/block.go
```diff
@@ -286,33 +311,151 @@ func (n *Node) getBlk(ctx context.Context, blkHash types.Hash) (int64, []byte, *
...
+	b := make([]byte, 0, readChunk)
+
+	for {
+		// Check absolute deadline for the entire resource.
+		if time.Now().After(deadline) {
+			return nil, fmt.Errorf("timeout")
+		}
+
+		// Set read deadline for this chunk.
+		s.SetReadDeadline(time.Now().Add(idleTimeout))
```
jchappelow, line 391: I think I may revisit the timeout for the first iteration of this loop.

node/consensus/block.go
```diff
@@ -63,6 +63,35 @@ func (ce *ConsensusEngine) validateBlock(blk *ktypes.Block) error {
 		return fmt.Errorf("block size %d exceeds max block size %d", blockTxnsSize, maxBlockSize)
 	}
 
+	// maxVotesPerTx
```
jchappelow, line 66: Maybe revise comment to the extent of `// Ensure that the number of event and resolution IDs within validator vote transactions does not exceed the max votes per transaction consensus limit.`

node/consensus/block.go
```diff
@@ -231,7 +264,7 @@ func (ce *ConsensusEngine) commit(ctx context.Context) error {
 	ce.mempool.RecheckTxs(ctx, ce.recheckTx)
 
 	ce.log.Info("Committed Block", "height", height, "hash", blkProp.blkHash.String(),
-		"appHash", appHash.String(), "updates", ce.state.blockRes.paramUpdates)
+		"appHash", appHash.String(), "numTxs", blkProp.blk.Header.NumTxns, "updates", ce.state.blockRes.paramUpdates)
```
jchappelow, line 267: I think I was a little sloppy putting the `"updates"` here.  I'm pretty sure we log update separately if there are any.

node/consensus/messages.go
```diff
@@ -119,7 +122,7 @@ func (ce *ConsensusEngine) NotifyBlockProposal(blk *ktypes.Block) {
 }
 
 // NotifyBlockCommit is used by the p2p stream handler to notify the consensus engine of a committed block.
-func (ce *ConsensusEngine) NotifyBlockCommit(blk *ktypes.Block, ci *types.CommitInfo) {
+func (ce *ConsensusEngine) NotifyBlockCommit(blk *ktypes.Block, ci *types.CommitInfo, blkID types.Hash, doneFn func()) {
```
jchappelow, line 125: looks like `blkID` is unused.

jchappelow (changes requested):
Just a few minor nits.

node/consensus/block.go
charithabandi: yeah, we log them in the block processor, removing this

jchappelow (approved):

jchappelow (approved):

PR #1401, opened 2025-02-21
engine: cache prepared statements

brennanjl:
This PR adds caching for statements in the interpreter, removing the need to redundantly parse, analyze, and generate SQL for statements that don't change. The cache is invalidated any time any table is changed; this is because a change in the DB schema may change what is necessary to make a query deterministic.

This PR also fixes a bug that I coincidentally found while clicking through the query planner looking for something else. It was not properly applying default ordering to window functions.

### DIFF
--- a/node/engine/interpreter/context.go
+++ b/node/engine/interpreter/context.go
@@ -3,6 +3,7 @@ package interpreter
 import (
 	"fmt"
 	"strings"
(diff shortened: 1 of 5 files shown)


jchappelow (approved):
Looks great. Engine hotspot is completely gone.

PR #1405, opened 2025-02-21
fix broadcast errors and BroadcastTx returns

jchappelow:
This restores the use of the RPC `BroadcastError` type that contains a code indicating the reason.

This also fixes a few issues with error wrapping and aliasing.

The `BroadcastTx` function now returns a `types.Hash` and a `*types.TxResult`, where the result is only non-nil if sync was enabled (waiting for commit).  This is essentially dropping come baggage from how a "Tx code" was returned from the somewhat polymorphic broadcast result, but it didn't make much sense in unless "commit" mode was used.


jchappelow:
This is working for what it was intended, but I'm putting it in draft as this revealed another issue with how txapp's accounts mempool view is coordinated with the `mempool.Mempool`.  We need to find a solution to effectively lock mempool, do the capacity check, do the application's checktx, and then unlock mempool.  This can be accomplished in a few ways.  One is to pass the Store method or the Mempool constructor some validation function to be run just before insert and return.  Another is to make exported lock/unlock methods for CE to use to make these operations atomic.  Another is to

jchappelow:
Temporary fix applied: mempool store before checktx, remove if checktx failed.  Updated CE and tx ann stream handler.
Retesting in mempool full scenario...

jchappelow:
@charithabandi squashed and rebased

charithabandi (approved):

PR #1407, opened 2025-02-24
make default chainID consistent

charithabandi:


jchappelow (approved):
OK.  But these would ideally be separate commits.

jchappelow (approved):
LGTM.  Let's use "Rebase and merge" for this one instead of squash.

PR #1409, opened 2025-02-24
node/metrics: add telemetry package

jchappelow:


jchappelow:
Will add a config for the OTEL collector and make this r4r.

jchappelow:
This is going to be a WIP for a while, but this creates a usable foundation.

charithabandi (approved):
Haven't tested it, but glanced it multiple times over the past few days, let's go with it.

PR #1416, opened 2025-02-25
node,peers: proposal annoucement and PEX fixes

jchappelow:
This fixes several things:
- the block proposal announcement stream handler `blkPropStreamHandler` could be abused to download block proposals multiple times because the sequence of `AcceptProposal` => download => `NotifyBlockProposal` was not made atomic.  This is fixed with a mutex.
- `AcceptProposal` does leader signature verification *first* before any other action such as retrieving the block from the block store.  This allows the proposal handler to return quite quickly if the annoucement was illegitimate (spam)
- followers, such as normal validators, now immediately reannounce a block proposal to their peers on first receipt of a new valid proposal.  This allows sparse network connectivity, including validators that are not directly connected to the leader.
- It was discovered that both gossipsub and dht external modules were initiating new peer connections using the `host.Host` directly.  This is undesirable if PEX is disabled in the application, so both are now created with appropriate options to disable their own peer discovery mechanisms if kwild is started with `p2p.pex = false`.
- Finally, we were ignoring the `cfg.P2P.TargetConnections` field.  It is now passed to the peer manager, where it is considered in `maintainMinPeers`.  Previously a hard coded value was used.

### DIFF
--- a/node/consensus.go
+++ b/node/consensus.go
@@ -217,6 +217,12 @@ func (n *Node) blkPropStreamHandler(s network.Stream) {
 		return
 	}
 
+	// This requires atomicity of AcceptProposal -> download -> NotifyBlockProposal.
+	// We also must not ignore any proposal messages since they may be real
+	// (signed by leader) leader while others may be spam.
+	n.blkPropHandlerMtx.Lock()
+	defer n.blkPropHandlerMtx.Unlock()
+
 	height := prop.Height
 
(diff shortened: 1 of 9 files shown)


jchappelow:
Will also "Rebase and merge" this one so the pex changes are separate from the block proposal protocol changes.

charithabandi (approved):

PR #1418, opened 2025-02-26
DO NOT MERGE test: re-enable acceptance test, with kgw

Yaiba:
This add kgw test support, also re-enable acceptance test in ci


Yaiba:
close in favor of #1419

Issue #1349, opened 2025-02-05
taming logs during sync


jchappelow:
During sync (either replay from block store or network sync), the logs look like this for each block:

```
2025-02-05 10:00:38.027 [INF] CONS: Executed block {height=7880 blockID=508be33c63baac12d33fb3654a061f362d9d5eb4a723072dad610c9b10f71609 numTxs=0 appHash=fdbafd67a554e050ec1e6389d1edaa8b97547dc9b963b6c4406d6357fa6d21cb}
2025-02-05 10:00:38.032 [INF] CONS: Committed Block {height=7880 hash=508be33c63baac12d33fb3654a061f362d9d5eb4a723072dad610c9b10f71609 appHash=fdbafd67a554e050ec1e6389d1edaa8b97547dc9b963b6c4406d6357fa6d21cb updates={}}
2025-02-05 10:00:38.033 [INF] CONS: Processing committed block {height=7881 blockID=d82a024443e637f22552038f650eeb39c61eb49c1f958e3b0a6684bf6ab97e83 appHash=8ad7a1ebc65ea925ec319864dea023d2898d27d92b77ccab12144bf4ba525797}
2025-02-05 10:00:38.039 [INF] BP: AppState updates:  {prevAppHash=fdbafd67a554e050ec1e6389d1edaa8b97547dc9b963b6c4406d6357fa6d21cb changesetsHash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 valUpdatesHash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 accountsHash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 txResultsHash=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 paramUpdatesHash=df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119}
```

We should handle logging differently in this stage.  Virtually every node out there will not even log for every block during sync (aka IBD), usually just reporting on block sync rate and current block every so often (e.g. ~100 blocks or on a time interval).  cometbft is a strange exception that is extremely noisy, and we should not consider that normal.

PR #1420, opened 2025-02-26
update checkpoint config and fix log crashes

charithabandi:


config/config.go
```diff
@@ -448,8 +448,8 @@ type MigrationConfig struct {
 
 type Checkpoint struct {
 	// Height 0 indicates no checkpoint is set. The leader will attempt regular block sync.
-	Height int64      `toml:"height" comment:"checkpoint height for the leader. If the leader is behind this height, it will sync to this height before attempting to propose a new block."`
-	Hash   types.Hash `toml:"hash" comment:"checkpoint block hash."`
```
jchappelow, line 452: Oof, I never noticed this:

```
# checkpoint info for the leader to sync to before proposing a new block
[checkpoint]
# checkpoint height for the leader. If the leader is behind this height, it will sync to this height before attempting to propose a new block.
height = 0
# checkpoint block hash.
hash = '0000000000000000000000000000000000000000000000000000000000000000'
```

node/consensus/messages.go
```diff
@@ -32,6 +32,19 @@ func (ce *ConsensusEngine) sendConsensusMessage(msg *consensusMessage) {
...
 }
 
+func (cm *consensusMessage) String() string {
+	switch v := cm.Msg.(type) {
+	case *blockProposal:
+		return cm.Msg.(*blockProposal).String()
+	case *blockAnnounce:
+		return cm.Msg.(*blockAnnounce).String()
+	case *vote:
+		return cm.Msg.(*vote).String()
```
jchappelow, line 42: Do any of these Stringers get long?    Switching the log from type to full string could get long?

It looks like in most cases we show the details elsewhere anyway?

```
2025-02-26 17:55:59.184 [INF] NODE: Accept proposal? {height=2823 blockID=80e82f7d1c7221352833b758b2623ebbe176a16f10894bd1aea8d06b151f14d8 prevHash=7222dbed39d06608a5c0d9847c04116d8995711732b3f27882f948eec96de2db from_peer=0226b3ff29216dac187cea393f8af685ad419ac9644e55dce83d145c8b1af213bd#secp256k1}
2025-02-26 17:55:59.185 [INF] NODE: processing block proposal {height=2823 hash=80e82f7d1c7221352833b758b2623ebbe176a16f10894bd1aea8d06b151f14d8 from=0226b3ff29216dac187cea393f8af685ad419ac9644e55dce83d145c8b1af213bd#secp256k1}
2025-02-26 17:55:59.185 [INF] CONS: Consensus message received {type=block_proposal sender=0226b3ff29216dac187cea393f8af685ad419ac9644e55dce83d145c8b1af213bd}
```

Related: https://github.com/kwilteam/kwil-db/issues/1349

jchappelow (dismissed):
LGTM, just a question about the logs getting much longer.

Maybe we log the full msg string in Debug, and just the type in Info?

config/config.go
charithabandi: Hahha yeah, I kinda didn't notice too, it should be okay now.

node/consensus/messages.go
charithabandi: Will handle the sync logs issue in a different PR, need to think through that a bit.

jchappelow:
conflicts, but will approve again

jchappelow (approved):

PR #1421, opened 2025-02-26
node,consensus: no duplicate proposals allowed

jchappelow:
This addresses a strange proposal announcement amplification that can occur given the right timing, which @charithabandi  observed once but could not easily reproduce.

In short, to support reannoucement of block proposals as well as preventing malicious spam proposals from stacking up behind the consensus event loop, we made the entire proposal stream handler atomic w.r.t. `AcceptProposal` => download = `NotifyBlockProposal`.  **However**, `NotifyBlockProposal` is mostly asynchronous, meaning the next call to `AcceptProposal` will return true and allow the same proposal, which includes a full block, to be downloaded again and queued in another consensus msg for `NotifyBlockProposal`.

To resolve this, we take a nearly identical approach to that used with `AcceptCommit` and its stream handler.  The main difference being that we completely block any subsequent proposal until the prior is processed (either fully rejected or fully accepted and block execution begun).

### DIFF
--- a/node/consensus.go
+++ b/node/consensus.go
@@ -10,6 +10,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"slices"
+	"sync"
 
 	ktypes "github.com/kwilteam/kwil-db/core/types"
 	"github.com/kwilteam/kwil-db/node/peers"
@@ -160,7 +162,7 @@ func (bp *blockProp) WriteTo(w io.Writer) (int64, error) {
 	return n, nil*/
 }
 
-func (n *Node) announceBlkProp(ctx context.Context, blk *ktypes.Block) {
+func (n *Node) announceBlkProp(ctx context.Context, blk *ktypes.Block, skipPeers ...peer.ID) {
 	rawBlk := ktypes.EncodeBlock(blk)
 	blkHash := blk.Hash()
 	height := blk.Header.Height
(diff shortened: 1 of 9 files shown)


node/interfaces.go
```diff
@@ -18,7 +18,7 @@ type ConsensusEngine interface {
 	InCatchup() bool
 
 	AcceptProposal(height int64, blkID, prevBlkID types.Hash, leaderSig []byte, timestamp int64) bool
-	NotifyBlockProposal(blk *ktypes.Block)
+	NotifyBlockProposal(blk *ktypes.Block, doen func())
```
charithabandi, line 21: `done` spelling

node/consensus/follower.go
```diff
@@ -120,7 +118,11 @@ func (ce *ConsensusEngine) AcceptProposal(height int64, blkID, prevBlockID types
 // 2. If the node is a validator and missed the block proposal message.
 func (ce *ConsensusEngine) AcceptCommit(height int64, blkID types.Hash, hdr *ktypes.BlockHeader, ci *types.CommitInfo, leaderSig []byte) bool {
 	if ce.stateInfo.hasBlock.Load() == height { // ce is notified about the blkAnn message already
-		return false
+		// that we processed correct proposal
+		if ce.stateInfo.blkProp != nil && ce.stateInfo.blkProp.blkHash == blkID {
```
charithabandi, line 122: umm, stateInfo read lock should be moved up here.

node/consensus.go
```diff
@@ -217,13 +219,23 @@ func (n *Node) blkPropStreamHandler(s network.Stream) {
...
-	height := prop.Height
+	var ceAsyncHandling bool // true once we've handed off to CE to asynchronously handle it, and CE takes the done func
+	defer func() {
+		if !ceAsyncHandling {
+			<-n.blkPropHandling
+		}
+	}()
+
+	from := s.Conn().RemotePeer()
+	n.log.Info("Accept proposal?", "height", height, "blockID", prop.Hash, "prevHash", prop.PrevHash,
```
charithabandi, line 237: `AcceptProposal` logs the same, so this can be removed.

charithabandi (approved):

PR #1422, opened 2025-02-26
core/types,node/types: add cached tx hash to ktypes.Transaction, create node/types.Tx

jchappelow:
This updates `core/types.Transaction` a.k.a. `ktypes.Transation` with an unexported `cachedHash` field, which is only populated or referenced via the new `HashCache` method, rather than modifying the `Hash` method to do this automatically.  A few reasons are outlined in the docs, but they include concurrency concerns, needless allocs in many cases, the tx mutability gotchas particularly in client side applications.

This also removes the `node/types.NamedTx` which was used in a handful of places to pair the hash with the `core/types.Transaction`, replacing `NamedTx` with `node/types.Tx` for brevity.  This new types adds a constructor `NewTx` that computes the has on construction.  The `Hash()` method of this type shadows the embedded `types.Transaction`'s method to return this stored hash.  This should be less likely to become an issue since it is "internal" to the `node/...` packages, and the hash being computed in the constructor avoids any concurrency concerns.  Plus it is documented explicitly as an _immutable_ transaction type.

The `node/types.Tx` with it's construction-time hash storage is primarily used in mempool and the CE.
The `HashCache` method is used in other places that deal with a `core/types.Block`, which is necessarily defined in terms of a `core/types.Transaction`.

### DIFF
--- a/node/consensus/block.go
+++ b/node/consensus/block.go
@@ -34,7 +34,7 @@ func (ce *ConsensusEngine) validateBlock(blk *ktypes.Block) error {
 	}
 
 	// Verify the merkle root of the block transactions
-	merkleRoot := blk.MerkleRoot()
+	merkleRoot := blk.CalcMerkleRoot() // NOTE: this expects CalcMerkleRoot to use tx.HashCache() to prepare the Transaction's internal hash cache
 	if merkleRoot != blk.Header.MerkleRoot {
 		return fmt.Errorf("merkleroot mismatch, expected %v, got %v", merkleRoot, blk.Header.MerkleRoot)
 	}
--- a/node/consensus/block.go
+++ b/node/consensus/block.go
@@ -117,14 +117,13 @@ func (ce *ConsensusEngine) lastBlock() (int64, types.Hash, time.Time) {
 // It is an error if the transaction is already in the mempool.
 // It is an error if the transaction fails CheckTx.
 // This method holds the mempool lock for the duration of the call.
-func (ce *ConsensusEngine) QueueTx(ctx context.Context, tx *ktypes.Transaction) error {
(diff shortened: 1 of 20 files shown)


node/consensus/block.go
```diff
@@ -34,7 +34,7 @@ func (ce *ConsensusEngine) validateBlock(blk *ktypes.Block) error {
 	}
 
 	// Verify the merkle root of the block transactions
-	merkleRoot := blk.MerkleRoot()
+	merkleRoot := blk.CalcMerkleRoot() // NOTE: this expects CalcMerkleRoot to use tx.HashCache() to prepare the Transaction's internal hash cache
```
jchappelow, line 37: `validateBlock` is a key method that is one of the first points at which a node will begin processing a block it has received, e.g. from p2p, block store, etc.  Ensuring that this method warms the internal tx hash cached in `core/types.Transaction` benefits many other spots downstream.

node/consensus/block.go
```diff
@@ -117,14 +117,13 @@ func (ce *ConsensusEngine) lastBlock() (int64, types.Hash, time.Time) {
...
 // This method holds the mempool lock for the duration of the call.
-func (ce *ConsensusEngine) QueueTx(ctx context.Context, tx *ktypes.Transaction) error {
+func (ce *ConsensusEngine) QueueTx(ctx context.Context, tx *types.Tx) error {
 	height, _, timestamp := ce.lastBlock()
 
 	ce.mempoolMtx.Lock()
 	defer ce.mempoolMtx.Unlock()
 
-	txHash := tx.Hash()
-	have, rejected := ce.mempool.Store(txHash, tx)
```
jchappelow, line 127: This silly optimization of having `mempool.Store` take a hash and the transaction separately was silly and now obsolete.

charithabandi (approved):

PR #1430, opened 2025-02-28
node/consensus: priority lock for CE business to preempt tx influx

jchappelow:
This adds `node/consensus.PriorityLock` that is like a normal mutex but:

 - `Lock` guarantees FIFO when acquired, no issues with sleeping goroutines in the `sync.Mutex` breaking ordering
 - `PriorityLock` jumps the line, getting the lock ahead of others that used `Lock`

This is intended in the very specific use case of CE where the `mempoolMtx` was locked in:

- `commit`: block commit, priority
- `prepareBlockProposal`: leader preparing block proposal, priority
- `QueueTx`: ingest new tx from either p2p announce or RPC broadcast, NOT priority

All of the "priority" cases are sequential, never concurrent with each other.  Only the `QueueTx` spot can be concurrent and at any time, as it is triggered from external sources.

This eliminate the one bad spot of mutex contention revealed by running (for a brief period) with `--profile-mode=mutex` while stress tool is flooding the network.

### DIFF
--- a/node/consensus/lock.go
+++ b/node/consensus/lock.go
@@ -1,0 +1,71 @@
+package consensus
+
+import "sync"
+
+type PriorityLockQueue struct {
+	mtx    sync.Mutex
+	active bool
+	queue  []chan struct{}
+}
+
+type queueFunc func(q []chan struct{}, c chan struct{}) []chan struct{}
+
+func appendTo[E any](q []E, c E) []E {
+	return append(q, c)
+}
+
+func prependTo[E any](q []E, c E) []E {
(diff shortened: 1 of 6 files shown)


charithabandi (approved):

node/consensus/lock.go
```diff
@@ -0,0 +1,71 @@
+package consensus
+
+import "sync"
+
+type PriorityLockQueue struct {
```
jchappelow, line 5: I need to doc and clean up this PR.

PR #1439, opened 2025-03-03
Notices

brennanjl:
This PR adds `NOTICE` calls to the transaction execution / call result.

It also separates execution errors (which are created by calling `ERROR()` or by things like primary key conflicts) from internal errors (like the DB being OOM). When calling a view action, it returns this as a separate field. When executing an action, it includes the error as the final log.

Finally, it fixes an issue where ERROR could not be called in a SQL statement (which should have been supported).


node/engine/interpreter/interpreter.go
```diff
@@ -378,6 +379,40 @@ func NewInterpreter(ctx context.Context, db sql.DB, service *common.Service, acc
...
+	if err == nil {
+		return nil, false
+	}
+
+	if u, ok := err.(userDefinedErr); ok {
+		return u.err, true
+	}
+
+	if strings.Contains(err.Error(), "SQLSTATE") {
+		return err, true
```
jchappelow, line 409: `true` here seems wrong.

node/engine/interpreter/interpreter.go
```diff
@@ -378,6 +379,40 @@ func NewInterpreter(ctx context.Context, db sql.DB, service *common.Service, acc
...
+	}
+
+	// TODO: detect pg error
+	return err, false
```
jchappelow, line 413: `node/types/sql.IsFatalDBError` seems like what you are thinking of here

node/engine/interpreter/interpreter.go
```diff
@@ -422,7 +457,7 @@ func funcDefToExecutable(funcName string, funcDef *engine.ScalarFunctionDefiniti
 				if !args[0].Null() {
 					msg = args[0].RawValue().(string)
 				}
-				return fmt.Errorf("error raised while executing: %s", msg)
+				return newUserDefinedErr(errors.New(msg))
```
jchappelow, line 460: Seems like somewhere on the call stack something is going to wrap this e.g. `%w` or `errors.Join`, in which case `err.(userDefinedErr)` in `unwrapExecutionErr` will fail to match. No?
Probably want to use `errors.As` in the `unwrapExecutionErr` *and* define the `Error` method with a pointer receiver and return a `*userDefinedErr`, so that it's impossible to miss.

node/engine/interpreter/interpreter.go
brennanjl: Ah ok. I will add checks for error codes then

node/engine/interpreter/interpreter.go
brennanjl: Correct.
Just fixed this.

jchappelow (approved):

PR #1442, opened 2025-03-04
testnet setup command to generate non-deterministic private keys

charithabandi:
With this PR
- The `kwild setup testnet` command will now generate non-deterministic private keys for the nodes.
- The runtime sanity checks for the genesis config, now checks if the `pubkey` and the `key_type` of the genesis validators are valid.
- `setup` commands now does validation checks on the validator flags

### DIFF
--- a/app/setup/testnet.go
+++ b/app/setup/testnet.go
@@ -1,10 +1,8 @@
 package setup
 
 import (
-	"crypto/sha256"
-	"encoding/binary"
+	"crypto/rand"
 	"fmt"
-	"math/rand/v2"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


app/setup/testnet.go
```diff
@@ -291,6 +291,7 @@ func GenerateNodeRoot(ncfg *NodeGenConfig) error {
 	return GenerateNodeDir(ncfg.RootDir, ncfg.Genesis, cfg, ncfg.NodeKey, "")
 }
 
+/*
 type deterministicPRNG struct {
```
jchappelow, line 295: You can remove this.   This was a workaround for Go <1.23 not having a `Read` method on the reader from `rand.NewChaCha8`.

jchappelow (approved):

PR #1444, opened 2025-03-05
rpc: add authenticated query

brennanjl:
Adds an RPC endpoint for authenticated queries. It also automatically uses the endpoint if the client has a private key.

This PR also fixes a bug I found in our roles system, where certain types of fine grained permissions could be lost if a node restarts.

### DIFF
--- a/core/client/client.go
+++ b/core/client/client.go
@@ -44,7 +44,8 @@ type Client struct {
 
 	noWarnings bool // silence warning logs
 
-	authCallRPC bool
+	// nodeRequiresAuth is true if the remote node requires authenticated call RPCs.
+	nodeRequiresAuth bool
 }
 
 // SvcClient is a trapdoor to access the underlying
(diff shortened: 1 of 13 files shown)


jchappelow (dismissed):
LGTM, although reviewed quickly.  `node/engine/interpreter/roles.go` has a lot of commented code.  Plans for that in the future or useful alternative or remove?

brennanjl:
Ah whoops I meant to remove that. I commented it out to try something else and then forgot to delete it.

jchappelow (approved):

jchappelow:
Thanks @brennanjl this worked to keep node data if so desired:

```sql
alter table kwild_engine.role_privileges add column granted bool default true;
alter table kwild_engine.role_privileges add unique (privilege_type, namespace_id, role_id);
drop view info.role_privileges;
CREATE VIEW info.role_privileges AS
SELECT
    r.name AS role_name,
    p.privilege_type::text AS privilege,
    n.name AS namespace,
    p.granted AS granted
FROM
    kwild_engine.role_privileges p
JOIN
    kwild_engine.roles r
    ON p.role_id = r.id
LEFT JOIN
    kwild_engine.namespaces n
    ON p.namespace_id = n.id
ORDER BY
    1, 2, 3, 4;
```

PR #1450, opened 2025-03-05
fix(bridge signer): handle nil value from CallWithoutEngineCtx

Yaiba:
This pr fix the nil value handling in bridge client

### DIFF
--- a/node/exts/erc20-bridge/erc20/meta_extension.go
+++ b/node/exts/erc20-bridge/erc20/meta_extension.go
@@ -1246,7 +1246,7 @@ func init() {
 				}
 
 				if leafNum == 0 {
-					app.Service.Logger.Info("no rewards to distribute, delay finalized current epoch")
+					app.Service.Logger.Debug("no rewards to distribute, delay finalized current epoch")
 					return nil
 				}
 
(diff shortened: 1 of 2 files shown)


node/exts/erc20-bridge/signersvc/kwil.go
```diff
@@ -114,29 +114,37 @@ func (k *signerClient) InstanceInfo(ctx context.Context, namespace string) (*Rew
 			return fmt.Errorf("failed to get epoch period")
 		}
 
-		info.Erc20, ok = row.Values[3].(string)
-		if !ok {
-			return fmt.Errorf("failed to get erc20")
+		if row.Values[3] != nil {
```
Yaiba, line 117: those fields are 'nullable' in method declaration

node/exts/erc20-bridge/erc20/meta_extension.go
```diff
@@ -1246,7 +1246,7 @@ func init() {
 				}
 
 				if leafNum == 0 {
-					app.Service.Logger.Info("no rewards to distribute, delay finalized current epoch")
+					app.Service.Logger.Debug("no rewards to distribute, delay finalized current epoch")
```
Yaiba, line 1249: this will be called on every block if no rewards, could be noisy

jchappelow (approved):

PR #1457, opened 2025-03-06
Fix/bridge: properly unuse a erc20 bridge

Yaiba:
This pr fix issue that when `unuse` an erc20 bridge ext, the listener topic is still registered.

Also add the support to update `period` config when re-use the existing erc20 bridge

### DIFF
--- a/node/exts/evm-sync/instance.go
+++ b/node/exts/evm-sync/instance.go
@@ -79,9 +79,9 @@ type EVMEventListenerConfig struct {
 	// Resolve is the function that will be called the Kwil network
 	// has confirmed events from Ethereum.
 	Resolve string
-	// ReuseTopic indicate whether to use the existing topic.
+	// TopicExists indicate whether the targeting topic exists.
 	// This is useful for erc20-bridge, we never delete the topics.
-	ReuseTopic bool
+	TopicExists bool
 }
 
 // RegisterListener registers a new listener.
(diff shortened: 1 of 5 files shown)


node/exts/evm-sync/instance.go
```diff
@@ -142,6 +142,11 @@ func (l *globalListenerManager) UnregisterListener(uniqueName string) error {
 	close(info.done)
 	delete(l.listeners, uniqueName)
 
+	err := orderedsync.Synchronizer.UnregisterTopic(ctx, db, eng, uniqueName)
```
Yaiba, line 145: hmm I can just return err here

Yaiba:
Just changed to skip to register topic when re-use erc20-bridge, because erc20-bridge instance will always live, we need to keep the topics

jchappelow (dismissed):
LGTM, but I don't have enough context here to be confident.

jchappelow:
We can start tagging after this (and a few passes of integration tests)

brennanjl:
@Yaiba there is not a need for all of the logic handling whether or not to create the topic. RegisterTopic is idempotent; it has an `ON CONFLICT DO NOTHING` clause.

brennanjl (approved):

Yaiba:
After this is merged, I need redeploy on testnet and test the bridge again

PR #1458, opened 2025-03-06
docker: fix GOWORK build

jchappelow:
The go.work file from the builder machine is not necessarily right for the builder, which specified a Go version in the Dockerfile such as 'FROM golang:1.23 AS build'.

Since we converted to the Dockerfile doing GOWORK builds, the solution is to do 'go work init . ./core ./test' inside the builder.

### DIFF
--- a/contrib/docker/kwild.debug.dockerfile
+++ b/contrib/docker/kwild.debug.dockerfile
@@ -15,6 +15,8 @@ RUN chmod 777 /var/run/kwil
 
 COPY . .
 
+RUN rm -f go.work && go work init . ./core ./test
+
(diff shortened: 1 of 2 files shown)


jchappelow:
Can someone with Go 1.23 still installed please confirm this works?

charithabandi:
Screenshot 2025-03-06 at 11 01 49 AM
 Yeah, works. I am on `1.23.7`

contrib/docker/kwild.debug.dockerfile
```diff
@@ -15,6 +15,8 @@ RUN chmod 777 /var/run/kwil
 
 COPY . .
 
+RUN rm -f go.work && go work init . ./core ./test
```
Yaiba, line 18: I did both(ci and docker) when doing KGW build, but I never really dig too much in this .
Now I think of it, isn't the path in the go.work are relative? shouldn't matter unless we use abs path right?

Yaiba: ah, nvm, the go version is different

charithabandi (approved):

PR #1460, opened 2025-03-06
fix: handle conflict when register topic

Yaiba:
Add handle conflicts back when register topic since it'll be called every time the engine starts


charithabandi (approved):

brennanjl:
We should close this, we need a different solution

Yaiba:
close in favor of #1461

PR #1461, opened 2025-03-06
bridge: fix improper topic register on startup

brennanjl:
Fixes an issue where we would modify application state on startup within the erc20 bridge. This state change has been moved to occur when the first `USE` is called for a bridge.


Yaiba (approved):
LGTM

PR #1464, opened 2025-03-07
engine: fix conn busy error

brennanjl:
Fixes an error where the network can halt with a `conn busy` error. This is a short-term fix; it prevents users from making scalar function calls while looping over a SQL statement. We can fix this by having go implementations of the scalar functions, but getting go implementations that match Postgres's is very non-trivial, and will likely introduce bugs.


brennanjl:
Also includes a very small fix for window functions

jchappelow (approved):

PR #1467, opened 2025-03-10
test: multi-key stress tool

jchappelow:
`-cl N` runs with `N` simultaneous instances with different keys.

The first one is special if `-key` is used.  The first one is responsible for deploying the namespace, or ensuring it is already deployed.  Only then do the rest follow.

For instance, to start 40 concurrent clients with unique keys, each doing 10 action executions ("posts") per block, with a post content length of up to 1461 bytes (variable length with (`-vl`), on a namespaces (`stress_xqbsioz1`) which should already exists since no key is given for the first client: 

```
./stress -ec 10 -el 1462 -ns stress_xqbsioz1 -vl -host http://44.x.y.z:8484 -cb -ei 1ms -ne -cl 40
```

### DIFF
--- a/core/client/client.go
+++ b/core/client/client.go
@@ -434,7 +434,6 @@ func WaitForTx(ctx context.Context, txQuery func(context.Context, types.Hash) (*
 		if err != nil {
 			// Only error out if it's something other than not found.
 			if !errors.Is(err, rpcclient.ErrNotFound) {
-				fmt.Println("not found, still trying")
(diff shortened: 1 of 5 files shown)


jchappelow:
I can see the sentry node become overwhelmed and get temporarily stuck and then catch up when RPC is bombarding like this.  RPC rate limiting would probably be a good idea in this situation.

charithabandi (approved):

PR #1468, opened 2025-03-10
engine: fix actions returning many decimals

brennanjl:
Fixes a bug found by @outerlook where actions that return a table of numerics sometimes incorrectly return duplicates of the first row for every subsequent row.

### DIFF
--- a/node/engine/interpreter/context.go
+++ b/node/engine/interpreter/context.go
@@ -2,6 +2,7 @@ package interpreter
 
 import (
 	"fmt"
+	"slices"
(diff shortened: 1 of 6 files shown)


jchappelow (dismissed):

jchappelow (approved):

jchappelow:
Fixes a bug where actions that return a table of numerics sometimes incorrectly
return duplicates of the first row for every subsequent row.

jchappelow (approved):

PR #1473, opened 2025-03-12
Fix the contention on the DB read accesses

charithabandi:
This PR fixes, the DB read access contention that's holding back the consensus engine from making progress. CE will now use reservedReads for DB access within the consensus path. 

Also fixes the `Closed the DB with an active transaction` issue in the failure scenarios during catchup.

### DIFF
--- a/node/consensus/engine.go
+++ b/node/consensus/engine.go
@@ -416,6 +416,10 @@ func (ce *ConsensusEngine) Start(ctx context.Context, fns BroadcastFns, peerFns 
 	ce.log.Info("Starting the consensus engine")
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
+	defer func() {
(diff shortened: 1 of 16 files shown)


node/consensus/engine.go
```diff
@@ -416,6 +416,10 @@ func (ce *ConsensusEngine) Start(ctx context.Context, fns BroadcastFns, peerFns
 	ce.log.Info("Starting the consensus engine")
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
+	defer func() {
+		ce.close() // Closes any consensus transactions opened either during catchup or during the regular consensus flow.
```
charithabandi, line 420: Earlier this wasn't being called during catchup errors as the function exists pretty earlier, so moving this to defer to cover all the exits which involves block processing (consensus transactions)

jchappelow (approved):

PR #1474, opened 2025-03-13
Transaction size limits for mempool

charithabandi:


jchappelow (dismissed):
LGTM.  But I think the limit should be sanity checked / capped by the max block size, otherwise user can change config to max block size forgetting to lower max tx size.

jchappelow (approved):

PR #1478, opened 2025-03-13
client,kwil-cli: skip auth option for query

jchappelow:
Authenticated query became the default recently if the client has a Signer (private key) configured.

This provides a way to use legacy unauthenticated queries even when your client has a private key configured.

The use case is that very often a simple query into the `'info'` name space requires no authentication out of the box, but doing auth incurs an extra RPC round trip to get a challenge (and associated server overhead with creating, persisting, and verifying the challenge and signature).  In the specific case of stress testing, with enough keys it will fail to get a challenge.

### DIFF
--- a/cmd/kwil-cli/cmds/query.go
+++ b/cmd/kwil-cli/cmds/query.go
@@ -82,7 +82,7 @@ func queryCmd() *cobra.Command {
 			}
 
 			return client.DialClient(cmd.Context(), cmd, dialFlags, func(ctx context.Context, cl clientType.Client, conf *config.KwilCliConfig) error {
-				res, err := cl.Query(ctx, sqlStmt, params)
+				res, err := cl.Query(ctx, sqlStmt, params, !rpcAuth)
 				if err != nil {
 					return display.PrintErr(cmd, err)
 				}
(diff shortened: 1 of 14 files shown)


core/client/client.go
```diff
@@ -347,13 +347,13 @@ func (c *Client) Call(ctx context.Context, namespace string, action string, inpu
...
 // Query executes a query.
-func (c *Client) Query(ctx context.Context, query string, params map[string]any) (*types.QueryResult, error) {
+func (c *Client) Query(ctx context.Context, query string, params map[string]any, skipAuth bool) (*types.QueryResult, error) {
 	if params == nil {
 		params = make(map[string]any)
 	}
 
 	// if a private key is configured, we should automatically authenticate.
-	if c.Signer() != nil {
+	if !skipAuth && c.Signer() != nil {
```
jchappelow, line 356: This is the reason for the "skip" (negation) semantics.  I did not want to have an error as the outcome in the case of `auth bool` where there is no signer, so I have a "do not authenticate" flag that only changes behavior if the client happened to have a key set.

cmd/kwil-cli/cmds/query.go
```diff
@@ -82,7 +82,7 @@ func queryCmd() *cobra.Command {
 			}
 
 			return client.DialClient(cmd.Context(), cmd, dialFlags, func(ctx context.Context, cl clientType.Client, conf *config.KwilCliConfig) error {
-				res, err := cl.Query(ctx, sqlStmt, params)
+				res, err := cl.Query(ctx, sqlStmt, params, !rpcAuth)
```
jchappelow, line 85: I'm not sure why `queryCmd` has the `rpc-auth` flag TBH.  We could just have this be `false` to never skip auth if key is configured.

jchappelow: We disallow query in private mode, so this `rpc-auth` flag setting the `AuthenticatedCalls` client flag seems to be totally useless otherwise:

```
func (svc *Service) Query(ctx context.Context, req *userjson.QueryRequest) (*userjson.QueryResponse, *jsonrpc.Error) {
	ctxExec, cancel := context.WithTimeout(ctx, svc.readTxTimeout)
	defer cancel()

	if svc.privateMode {
		return nil, jsonrpc.NewError(jsonrpc.ErrorNoQueryWithPrivateRPC,
			"query is prohibited when authenticated calls are enforced (private mode)", nil)
	}
```

brennanjl: I agree, I think we can remove

brennanjl (approved):
Other than my comment on unused flag lgtm

PR #1488, opened 2025-03-20
Numeric cast

brennanjl:
Fixes two engine bugs regarding type casting.

The first bug caused type casts on parenthesis to not be applied properly. The second bug caused type casts of integers or text to numerics to not properly apply scale and precision.

Also made a few minor improvements to error messages. These break consensus, but since the above bug fixes break consensus as well I figured this was not a big deal.


brennanjl:
Converting to draft because I noticed that the same issues exist for non-int types when typecasting

Yaiba (approved):

PR #1493, opened 2025-04-07
core: add EncodeValue fuzz testing

brennanjl:
Adds fuzz testing for the `EncodeValue` function. It also fixes 3 bugs in `EncodeValue`, and removes some redundant code and makes minor improvements:

# Bugs

- nil byte slices were not being encoded properly. They were being encoded with data in the `Data` field, which is incorrect.
- typed nil values were not being encoded properly, (e.g. `var nilVal []string = nil`).
- nil pointers were not being encoded

# Improvements
- Makes an error message for empty arrays more clear. Previously, it read that all array values were empty, however it actually checked if there were no values at all
- Removes a redundant case statement that is impossible

### DIFF
--- a/core/types/payloads.go
+++ b/core/types/payloads.go
@@ -659,22 +659,32 @@ func EncodeValue(v any) (*EncodedValue, error) {
 			},
 			Data: nil,
 		}, nil
+	} else if bts, ok := v.([]byte); ok {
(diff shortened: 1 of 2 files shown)


Yaiba (approved):
lgtm

Issue #1496, opened 2025-04-08
core/types.ReadCompactString len out of range


Yaiba:
```
// Code generated by llm-fuzzer; DO NOT EDIT.
package types_test

import (
	"testing"
	types "github.com/kwilteam/kwil-db/core/types"
	io "io"
	"bytes"
)


func TestReadCompactString(t *testing.T) {
	data := []byte("\x8f\x8f\x8f\x80\xa9\xa9\xa9\xa9\xa9\x01")
	var r io.Reader = bytes.NewReader(data)
	types.ReadCompactString(r)
}
```

Got 

```
=== RUN   TestReadCompactString
--- FAIL: TestReadCompactString (0.00s)
panic: runtime error: makeslice: len out of range [recovered]
	panic: runtime error: makeslice: len out of range
```

Yaiba:
ah this is basically the same as #1495

ukane-philemon:
Should be marked as closed.

PR #1497, opened 2025-04-08
core/types: fix possible negative makeslice

### LINKED ISSUE #1494: DecodeBlockHeader len out of range
```go
// Code generated by llm-fuzzer; DO NOT EDIT.
package types_test

import (
	"strings"
	"testing"
	types "github.com/kwilteam/kwil-db/core/types"
)


func TestDecodeBlockHeader(t *testing.T) {
	data := strings.NewReader("0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001")

	_, err := types.DecodeBlockHeader(data)
	if err != nil {
		return
	}
}
```

got 
```
panic: runtime error: makeslice: len out of range [recovered]
	panic: runtime error: makeslice: len out of range
```

chappjc:
Resolves #1494 and https://github.com/kwilteam/kwil-db/issues/1496


Yaiba:
There are new issues popping up, 
```--- FAIL: FuzzDecodeBlock (20.90s)
    fuzzing process hung or terminated unexpectedly while minimizing: EOF
```

This is the fuzz code I'm using

```
// Code generated by llm-fuzzer; DO NOT EDIT.
package types_test

import (
	"testing"
	types "github.com/kwilteam/kwil-db/core/types"
)


func FuzzDecodeBlock(f *testing.F) {
	// Seed corpus
	testcases := []struct{
		rawBlk []byte
	}{
		{rawBlk: []byte{}},
		{rawBlk: []byte("genesis block")},
		{rawBlk: []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9}},
	}

	for _, tc := range testcases {
		f.Add(tc.rawBlk)
	}

	f.Fuzz(func(t *testing.T, rawBlk []byte) {
		// Add necessary setup and initialization

		// Call the function being tested
		_, err := types.DecodeBlock(rawBlk)
		// if returns error and error is not nil, short circuit
		if err != nil {
			return
		}
	})
}
```

chappjc:
That's the allocation for a very large number of transactions.  Adding sanity check to ensure input bytes can actually contain the encoded number.

PR #1498, opened 2025-04-09
engine: fix window function ranking bug

brennanjl:
This PR fixes a bug found by @outerlook where default ordering applied to window functions actually changed the functional output of a window function. Ordering on window functions actually changes the window itself, thus leading to a semantically different execution result.

This PR removes the orderings applied to aggregates used in functions altogether, as further investigation has shown that they actually aren't necessary. I've included in the code a link to a resource that explains how window functions break ties, and why default ordering is thus not a problem.


Yaiba (approved):
LGTM

Yaiba (approved):

PR #1500, opened 2025-04-15
data race accessing ce.leader

charithabandi:
This PR fixes a data race issue in the consensus engine that occurs when accessing the `ce.leader` field. The issue arises when block proposal handling and role updates happen concurrently. Specifically, `ce.leader` is incorrectly accessed in `NotifyBlockProposal` to set the block proposal's sender as the current leader. This is problematic not only because the field is accessed without proper locking but also because it assumes that the proposal always comes from the current leader, which is not guaranteed (maybe a leader change scenario)

The fix involves including the sender information directly in the block proposal message, and accessing `ce.leader` fields in the stream handlers and their CE counterparts.

### DIFF
--- a/node/consensus.go
+++ b/node/consensus.go
@@ -30,11 +30,12 @@ type (
 )
 
 type blockProp struct {
-	Height    int64
-	Hash      types.Hash
-	PrevHash  types.Hash
-	Stamp     int64
-	LeaderSig []byte
+	Height       int64
+	Hash         types.Hash
+	PrevHash     types.Hash
+	Stamp        int64
+	LeaderSig    []byte
+	SenderPubKey []byte
 	// Replacing *types.Hash
 }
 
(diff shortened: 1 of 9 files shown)


chappjc:
Small request: will you update the PR desc and/or commit message:
- what was the scenario where there was a data race
- what specific CE design patterns lead to the problem
- was is a problem in production or just test where the notify function was called manually? was there even a logic race in the stream handler where the leader could change between AcceptProposal and NotifyBlockProposal
- does the message structure change wire (p2p) protocol or just internals?

node/consensus.go
```diff
@@ -114,6 +119,17 @@ func (bp *blockProp) ReadFrom(r io.Reader) (int64, error) {
...
+	var pubkeyLen int64
+	if err := binary.Read(r, binary.LittleEndian, &pubkeyLen); err != nil {
+		return n, err
+	}
+	n += 8
+	bp.SenderPubKey = make([]byte, pubkeyLen)
+	nr, err = io.ReadFull(r, bp.SenderPubKey)
+	n += int64(nr)
+	if err != nil {
+		return int64(nr), err
```
chappjc, line 131: wouldn't `return int64(n), err`  be more correct here?

chappjc (approved):

PR #1504, opened 2025-04-28
engine: fix bug on partial INSERT ... SELECT

brennanjl:
Resolves https://github.com/trufnetwork/kwil-db/issues/1503

CC @outerlook


MicBun (approved):
nice
cc @outerlook

PR #1508, opened 2025-05-12
add better test coverage for parser

brennanjl:
This PR adds better test coverage for the parser and engine. It also finds a very minor bug in the parser, where it would incorrectly parse out default ordering. This issue isn't seen when using Kwil holistically because it is hidden by the default ordering, but if you use the `parse` package as a standalone package you would hit it.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,6 @@ require (
 	github.com/samber/lo v1.47.0
 	github.com/spf13/cobra v1.9.1
 	github.com/spf13/pflag v1.0.6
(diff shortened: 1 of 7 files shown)


MicBun (approved):

MicBun:
@brennanjl, the PR broke the unit test. Could this be checked?

brennanjl:
Yep. Might be a couple days, I'm on vacation this week. Should be able to take a look today or tomorrow.

MicBun (approved):

PR #1509, opened 2025-05-22
testing: show errors in all use cases, handle use case when an error

sapience:
In [this](https://github.com/trufnetwork/kwil-db/commit/79d66da7f16950eaabfb285480ce05b676df0173) commit there was introduced testing framework better handling. But the implementation has issues:
1) In some cases errors are not shown properly
2) No fail if we expect an error but it was not occurred 

This PR fixes the issues.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

### DIFF
--- a/testing/testing.go
+++ b/testing/testing.go
@@ -339,15 +339,19 @@ func (e *TestCase) runExecution(ctx context.Context, platform *Platform) error {
 	// but there are times where it might be returned as a separate error
 	// (e.g. in case of an extension erroring).
 	// Therefore, we need to check both.
-	var receivedErr error
+
+	var (
+		receivedErr    error
+		expectsErr     bool
+		expectsErrText string
+	)
+
 	if err != nil {
 		receivedErr = err
 	} else if res.Error != nil {
 		receivedErr = res.Error
 	}
 
-	expectsErr := false
-	expectsErrText := ""
 	if e.Err != nil {
 		expectsErr = true
 		expectsErrText = e.Err.Error()
--- a/testing/testing.go
+++ b/testing/testing.go
@@ -346,30 +346,40 @@ func (e *TestCase) runExecution(ctx context.Context, platform *Platform) error {
 		receivedErr = res.Error
 	}
 
+	expectsErr := false
+	expectsErrText := ""
+	if e.Err != nil {
+		expectsErr = true
+		expectsErrText = e.Err.Error()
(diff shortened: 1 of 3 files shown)


MicBun:
Hi @sapience, thank you for your first contribution!
```
    testing.go:315: executing action "get_users" against namespace "users"
    testing.go:50: test failed: test "conflicting wallet address - failure" failed: expected error message to contain substring "duplicate key value", but the error didn't happen
```
The unit test failed on your modification. Would you like to fix it?

testing/testing.go
```diff
@@ -346,32 +346,42 @@ func (e *TestCase) runExecution(ctx context.Context, platform *Platform) error {
...
 		if !expectsErr {
-			return fmt.Errorf(`unexpected error: %w`, err)
+			return fmt.Errorf(`unexpected error: %w`, receivedErr)
 		}
 
 		return nil
 	}
 
+	if expectsErr {
+		return fmt.Errorf(`expected error message to contain substring "%s", but the error didn't happen`, expectsErrText)
```
MicBun, line 382:
```
    testing.go:315: executing action "get_users" against namespace "users"
    testing.go:50: test failed: test "conflicting wallet address - failure" failed: expected error message to contain substring "duplicate key value", but the error didn't happen
```
The unit test is failing here.

MicBun (changes requested):

testing/testing.go
```diff
@@ -346,32 +346,42 @@ func (e *TestCase) runExecution(ctx context.Context, platform *Platform) error {
 		receivedErr = res.Error
 	}
 
+	expectsErr := false
+	expectsErrText := ""
```
ukane-philemon, line 350: To maintain code style as observed in this test func(see L342 above), consider: 
```suggestion
	var expectsErr bool
	var expectsErrText string
```

ukane-philemon (approved):
LGTM

testing/testing.go
chappjc: Agree. Don't unnecessarily declare with `:=` when Go's zero-value initialization is part of the spec.  Consistency is the other consideration.

sapience: Thanks. Fixed.

PR #1514, opened 2025-05-31
multi: Add new RPC method `chain.block_header`

### LINKED ISSUE #1506: Problem: Fetch time information fetches the whole block
Currently, retrieving block header fields (like timestamp or merkle root) via `chain.block` necessitates fetching the entire block, including all transaction data. This is inefficient for clients only needing header metadata.

```
curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"chain.block","params":{"height":100},"id":1}' https://nodeaddress:port/rpc/v1
```

Propose adding a new RPC method (e.g., `chain.get_header`) or an option to `chain.block` to return only the `BlockHeader` object for a specified height or hash, excluding the `txns` payload. This would significantly reduce data transfer for header-only use cases.

cc @MicBun 

I believe this turns out to be a priority if we want to decrease load on our nodes, or increase performance of the website, as each block could have several MB of information being fetched

ukane-philemon:
## Description
This commit adds a new RPC method chain.block_header, which involved implementaion of new GetBlockHeaderByHeight and GetBlockHeader methods on *Node, *MemBS and *BlockStore.

Also, the stale rpc spec version was updated.

## Related Issue
Resolves #1506

## Motivation and Context
Feature request.

## How Has This Been Tested?
After configuring PostgreSQL, I started a test kwild daemon using `kwild start --autogen` and manually tested the new rpc method

```txt

~ kwil-db % curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"chain.block_header","params":{"height":100},"id":1}' http://0.0.0.0:8484/rpc/v1 | jq
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "Version": 0,
    "Height": 100,
    "NumTxns": 0,
    "PrevHash": "caefd257cbf82d59339d69e046ba45bdfd89edf12c23bf2c38e22c392f8df7e5",
    "PrevAppHash": "1e8453a02266c0e25409d6bc9f41ec2134c613aa6d036bc2a3a649c7d3f0e99b",
    "Timestamp": "2025-05-31T17:31:48.585Z",
    "MerkleRoot": "0000000000000000000000000000000000000000000000000000000000000000",
    "ValidatorSetHash": "af5570f5a1810b7af78caf4bc70a660f0df51e42baf91d4de5b2328de0e83dfc",
    "NetworkParamsHash": "d3e010a95bb2425e5e349c342e421d39f42c1e5af6c72befeac25a8067605451",
    "NewLeader": null
  }
}

```
Unit tests were also written to cover new methods.

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] My change requires a change to the documentation. (maybe?)
- [ ] I have updated the documentation accordingly.

### Checklist Explanation:
This is a new feature, and there's nowhere on the [Kwil Docs](https://docs.kwil.com) that requires an update; the RPC spec has been updated, so I think that's sufficient.

## How to Review this PR:
Look at the new methods introduced for code style check and correctness, then run the tests and spin a kwil daemon with `--autogen` flag to test the `chain.block_header` RPC method.

## Additional Information:
The new RPC method only returns the block header info, which is missing the block hash. If a block header is fetched with the height, the caller won't have the hash in the response unless they hit the `chain.block` method.

We can either add it to the `BlockHeaderResponse` or ignore.

### DIFF
--- a/node/store/store.go
+++ b/node/store/store.go
@@ -551,6 +551,60 @@ func (bki *BlockStore) GetByHeight(height int64) (types.Hash, *ktypes.Block, *kt
 	return hashes.hash, blk, ci, err
 }
 
+// GetBlockHeader retrieves the block header for the given block hash. If the
+// block is not found, it returns types.ErrNotFound. The header is the decoded
+// header, not the raw bytes. This is useful for getting the header without
+// loading the entire block, which can be large.
+func (bki *BlockStore) GetBlockHeader(blkHash types.Hash) (*ktypes.BlockHeader, error) {
+	if !bki.Have(blkHash) {
+		return nil, types.ErrNotFound
+	}
+
+	var block *ktypes.Block
+	err := bki.db.View(func(txn *badger.Txn) error {
+		blockKey := slices.Concat(nsBlock, blkHash[:])
+		item, err := txn.Get(blockKey)
+		if err != nil {
+			return err
+		}
+
+		return item.Value(func(val []byte) error {
+			block, err = ktypes.DecodeBlock(val)
+			return err
+		})
+	})
+
+	if errors.Is(err, badger.ErrKeyNotFound) {
+		return nil, types.ErrNotFound
+	}
+
+	return block.Header, nil
+}
+
+// GetBlockHeaderByHeight retrieves the block header for the given block height.
+// If the block is not found, it returns types.ErrNotFound. The header is the
+// decoded header, not the raw bytes. This is useful for getting the header
+// without loading the entire block, which can be large.
+func (bki *BlockStore) GetBlockHeaderByHeight(height int64) (*ktypes.BlockHeader, error) {
+	bki.mtx.RLock()
+	hashes, have := bki.hashes[height]
+	bki.mtx.RUnlock()
+	if !have {
+		return nil, types.ErrNotFound
+	}
+	// NOTE: Could just use bki.GetBlockHeader here but it seems the appHash
+	// check below is important when fetching data by height.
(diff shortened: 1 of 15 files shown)


ukane-philemon:
cc: @MicBun

node/store/store.go
```diff
@@ -551,6 +551,60 @@ func (bki *BlockStore) GetByHeight(height int64) (types.Hash, *ktypes.Block, *kt
...
+	}
+	// NOTE: Could just use bki.GetBlockHeader here but it seems the appHash
+	// check below is important when fetching data by height.
+	blk, ci, err := bki.Get(hashes.hash)
+	if err != nil {
+		return nil, err
+	}
+	// NOTE: appHash should match hashes.appHash, so check it here.
+	if ci.AppHash != hashes.appHash {
+		return nil, fmt.Errorf("appHash mismatch: %s != %s", ci.AppHash, hashes.appHash)
```
ukane-philemon, line 603: Why is appHash check important when retreiving block info by height?

outerlook (approved):
> The new RPC method only returns the block header info, which is missing the block hash. If a block header is fetched with the height, the caller won't have the hash in the response unless they hit the chain.block method.
> 
> We can either add it to the BlockHeaderResponse or ignore.

IMO it's fine to ignore for this PR and open a new issue for it to be prioritized as  demand appears

MicBun (approved):

PR #1523, opened 2025-06-09
chore: better performance on `array_append`

### LINKED ISSUE #1522: Bug: `array_append` performance is a bottleneck
This issue is put on hold until we can prioritize the performance of our operations again

**Problem:**
The interpreter's implementation of `array_append` exhibits significant performance degradation when used repeatedly, such as within a loop to construct an array.

**Analysis:**
The `array_append` function must operate immutably, returning a new array rather than modifying the original. To achieve this, it performs a deep copy of the entire existing array before appending the new element.
Two factors contribute to the slowness:
 1. **Copy Method:** The original deep copy mechanism (`copyVal`) inefficiently copies the array by serialising the *entire* structure and all its elements to a string (`stringifyValue`) and then parsing it all back (`parseValue`) to reconstruct the array. This string conversion overhead is extremely high.
 2.  **Complexity:**
      - A *single* `array_append` call on an array of size `N` is `O(N)` because all `N` elements must be copied.
      - Building an array of size `N` by calling `array_append` repeatedly in a loop results in an overall time complexity of **`O(N^2)`**, as it performs copies of size 0, then 1, then 2, ..., up to N-1.
 
This `O(N^2)` behaviour, combined with the highly inefficient string-based copy method, means that while performance is acceptable for small arrays or few appends, it degrades rapidly. After a certain point, adding each new item becomes prohibitively expensive as the cost to copy the entire array (via string conversion) grows with every iteration.

We rely heavily on `array_append`, making the impact of this issue very high.

**Impact Example:**
Getting 5 years of daily data locally via `get_index_change` (which involves building a large array) takes > 60 seconds. By replacing the inefficient string-based copy with a direct memory copy (while still maintaining O(N^2) loop complexity), the same operation completes in less than 9 seconds, demonstrating that the copy method is the primary bottleneck.

**Proposed Fix:**
Replace the use of the generic `copyVal` (stringify/parse) within `array_append` with a type-aware, efficient deep-copy function (`copyArray`) that directly copies the underlying data slices in memory, avoiding all string serialization/deserialization overhead. Handle the `arrayOfNulls` type specifically.
 
cc @MicBun

outerlook:
## Description

The `array_append` function was slow for building large arrays due to an inefficient deep-copy method that serialized the array to a string and back. An internal benchmark improved from >60s to <9s with this change.

This PR addresses the bottleneck and fixes related array-handling bugs:
- Replaced the string-based copy in `array_append` with `copyArray`, an efficient, type-aware memory copy.
- Improved the slice growth strategy in `setArr` to reduce reallocations.
- Fixed `arrayOfNulls` to correctly convert to a typed array when a non-null element is set or appended.

## Related Issue

- fix #1522

## Motivation and Context

The previous performance of `array_append` made building large arrays in Kuneiform actions impractical. This optimization is critical for real-world use cases that rely on array manipulation.

## How Has This Been Tested?

- Updated existing unit tests for array functions.
- Added new tests for `nil` and untyped array edge cases.
- Updated tests to reflect the new error propagation behavior.
- All new and existing tests pass.

## Types of changes
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (please describe): Performance optimization.

## Checklist:
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### Checklist Explanation:
The changes are internal performance optimizations and bug fixes. No user-facing Kuneiform syntax or engine APIs were changed, so documentation updates are not required.

## How to Review this PR:
1.  Start with the new `copyArray` function in `node/engine/interpreter/values.go`.
2.  Review the updated `array_append` implementation in `node/engine/interpreter/interpreter.go`.
3.  Examine the changes to `setArr` and `arrayOfNulls` in `values.go`.
4.  Look at the updated tests in `interpreter_test.go` to see the new error propagation behavior and edge case handling.


outerlook:
This PR was built only for testing purposes, as to really investigate and merge demands more effort to understand the impact.

but at least on the happy path, it works.

So I'll put this on hold until we can prioritize the performance of our operations again

But this has a significant impact on all of our operations (insert, reading data, etc)

cc @MicBun

outerlook:
testing https://github.com/trufnetwork/node/blob/main/internal/benchmark/load_test.go

with 400 streams, 5 years of daily data (5 x 365 data points), `get_index_change`

before this PR:

```log
Qty Streams: 400, Branching Factor: 8, Data Points: 1825, Visibility: Public, Procedure: get_index_change, Samples: 1, Memory Usage: 1691 MB
  Mean Duration: 36.800188194s
  Min Duration: 36.800188194s
  Max Duration: 36.800188194s
```

after this PR

```log
Qty Streams: 400, Branching Factor: 8, Data Points: 1825, Visibility: Public, Procedure: get_index_change, Samples: 1, Memory Usage: 1691 MB
  Mean Duration: 9.097067288s
  Min Duration: 9.097067288s
  Max Duration: 9.097067288s
```


What other kinds of queries will share the improvements?
queries that have array_append used inside a for..loop for many loops

MicBun (dismissed):
lgtm, lets wait for CI to finish

PR #1528, opened 2025-06-16
chore: Smarter peer-sampling for block sync

### LINKED ISSUE #1527: Bug: block retrieval process is slow
related to https://github.com/trufnetwork/truf-network/issues/886

### Problem statement

Nodes that join the network several thousand blocks behind often **never progress beyond their initial height**, even though fully‑synced validators are reachable.
Log trace:

```
... got block proposal height=67853 …
... Starting block sync… {height=63124}
... block 63127 not found on peer … ; their best height is 63126
```

What happens:

1. `getBlkHeight` selects a small random slice (≈ 20 %) of connected peers.
2. If those peers are new “archive‑lite” nodes they answer *“not found; best = height‑1”*.
3. After five such answers the algorithm assumes the block is not committed yet and backs off → consensus watchdog keeps toggling between “proposal ignored” and “start sync”, but height never moves (or moves slowly).

Effect:

* Catch‑up latency explodes from minutes to **hours / infinity**.
* Repeated network dials waste bandwidth and log space.
* A swarm of honest but partial peers can unintentionally **DOS** late‑joining full nodes.

---

### Conditions to reproduce

* ≥ 2 fully‑synced validators already at height **H**.
* ≥ 10 new non‑validator peers started with empty block store.
* Start a fresh full node with `--bootnodes` pointing to the mixed set.
* Observe that the new node stops around `H‑4 k` and prints the log snippet above.

---

### Why it matters

* Operators adding archival nodes after genesis cannot get them online.
* Long‑running networks with many API‑only nodes (no block storage) gradually become *un‑joinable* without manual peer whitelisting.
* Consensus liveness is unaffected, but data availability for new validators and RPC users is at risk.

---

### Possible direction

*Maintain a tiny in‑process cache of each peer’s advertised “best height” and, during block sync, **skip peers whose best < targetHeight** (with a short TTL, e.g. 15 min).*

Other heuristics that could complement or replace the above:

* Dynamically enlarge the peer sample size when the node is far behind.
* Penalise peers that return “not found” for a height they previously claimed to have.

---

### Acceptance criteria

* A node joining **5 000 blocks behind** catches up to the tip within the same wall‑clock time as when only validators are present (currently < 10 min on test‑net).
* The fix does **not** increase outbound traffic when the node is already near the tip.
* No regressions in private‑mode / whitelist deployments.

outerlook:
## Description
- Added a peer-height cache (`peerBest`) that records each peer’s best known height and last-seen time.  
- `getBlkHeight` now:
  – filters out peers that are certainly behind,  
  – samples a larger fraction of peers in small networks (all peers ≤ 5; ≥3 or ⅓ ≤ 15; ≥3 or ⅕ otherwise).  
- Introduced periodic GC (`gcPeerCache`) to cap map size and age-out stale entries.  
- Reduced log noise:
  – consensus ignores proposals > 100 blocks ahead.  
- Inline comments explain “why” for future maintainers.

## Related Issue
Closes #1527 — block sync could stall when the single sampled peer was behind.

## Motivation and Context
Previously we queried only ⌈N·0.2⌉ peers (min 1). If that peer lagged, the node stopped advancing.  
The new cache + adaptive sampling keeps sync progressing while minimising unnecessary network traffic.

## How Has This Been Tested?
1. Added unit tests (`node/blocksync_sampling_test.go`) that cover:  
   • sample-size calculation for various peer counts,  
   • cache filtering behaviour,  
   • cache GC.  
2. All existing integration tests still pass (`go test ./...`).

## Types of changes
- [x] Bug fix (non-breaking)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other

## Checklist
- [x] Code follows project style
- [x] Tests added
- [x] All tests pass
- [ ] Docs update – N/A (internal behaviour)

## How to Review this PR
Start with `node/block.go` changes:  
• const/comment block, `peerInfo`, `peerBest`, new sampling logic, and `gcPeerCache`.  
Skim `node/consensus.go` for the new `FAR_AHEAD_THRESHOLD` guard.  
Run `go test ./node/...` to verify tests.

## Additional Information
None

### DIFF
--- a/node/block.go
+++ b/node/block.go
@@ -21,6 +21,19 @@ import (
 	"github.com/libp2p/go-libp2p/core/peer"
 )
 
+// Block-sync peer-sampling
+//
+// Historical bug: querying only a small fixed fraction (⌈N·0.2⌉, minimum 1)
+// of peers could pick a single behind-height peer and stall block sync.  The
+// revised algorithm below fixes that by:
+//   - keeping a lightweight "peer → best height" cache updated on every
+//     announcement or block response,
+//   - skipping peers that are definitely behind, and
+//   - scaling the sample size with network size (all peers for ≤5, ≥3 or ⅓ for
+//     ≤15, and ≥3 or ⅕ thereafter).
+//
+// When the cache is stale we fall back to the full peer set, so liveness is
+// preserved at the cost of extra bandwidth.
 const (
 	blkReadLimit          = 300_000_000
 	defaultBlkGetTimeout  = 90 * time.Second
@@ -34,7 +47,12 @@ type peerInfo struct {
 	seenAt time.Time
 }
 
-var peerBest sync.Map // map[peer.ID]*peerInfo
+// peerBest remembers the highest block height we have *ever* seen from each
+// peer.  It is an opportunistic heuristic: if we know a peer is at height 10
+// and we need block 20, we can skip querying it.  Entries are evicted by
+// gcPeerCache; if we evict too aggressively we merely sample the peer again
+// later.  sync.Map lets hot paths read/write without explicit locks.
+var peerBest sync.Map // map[peer.ID]peerInfo
 
 func (n *Node) blkGetStreamHandler(s network.Stream) {
 	defer s.Close()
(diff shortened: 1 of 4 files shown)


outerlook:
@MicBun I tried locally, and it worked
- logs about a node not having the block were gone
- seing v1/health endpoint for the blocks seems to be moving forward

I'll also implement at usher labs node to see the effect in the long run, but can you test on your local as well if seems all ok?

PR #1533, opened 2025-06-27
chore: update main module to reflect truf.network

williamrusdyputra:
resolves: https://github.com/trufnetwork/truf-network/issues/632

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -37,7 +37,6 @@ jobs:
 
   test:
     needs: changes
-    # TODO: uncomment this when we have a way to run the test job
     #  only run only if there are changes and non-draft PRs
     if: ${{ needs.changes.outputs.run == 'true' && !github.event.pull_request.draft}}
     runs-on: ubuntu-latest


MicBun:
@williamrusdyputra 
https://github.com/marketplace/actions/backport-action
It looks like backport-action is forked by them to pin the version or having some modification

williamrusdyputra:
> @williamrusdyputra https://github.com/marketplace/actions/backport-action It looks like backport-action is forked by them to pin the version or having some modification
> 
> Same for https://github.com/FantasyJony/openzeppelin-merkle-tree-go
> 
> Can we use the original repo (not the forked one)?

for backport, I think we can since their commit to the repo is only about bumping version (not sure if that will fail something)

for the openzeppelin, they have their own fix regarding an issue, I think I have to copy their commits if we want to start from forked repo too.

MicBun:
I think we should try it and see the result.
Or we can do some workaround if it not available.

How severe the missing repo for us? is it test only or impacting core functionalities?

williamrusdyputra:
@MicBun openzeppelin is used in the core functionalities, especially in merkletree implementation which also affects erc20 bridge extensions, etc.

backport, not used so much, its just a utility function only triggered when they want to backport to a commit I guess, we will not use it anymore I think.

MicBun:
@williamrusdyputra 
I forked the already forked repo here
https://github.com/trufnetwork/openzeppelin-merkle-tree-go

> should I try to fork and copy the commit for openzeppelin? (its a bit risky though)

What are the risk?

As for backport, as it is listed in marketplace action, I think we can directly use it?

williamrusdyputra:
fixing lint error

PR #1534, opened 2025-07-03
fix: race condition in peer management and syncmap

brennanjl:
This PR makes two fixes:

1. It fixes a race condition that can occur while finding peers. This can cause unexpected failure when peers try to find each other.
2. It fixes a race condition when extracting and deleting from a syncmap. This method was not actually used anywhere so it is not currently an issue, but it is exposed to developers in several places.


brennanjl:
@pr-time-tracker bug commit [link](https://github.com/trufnetwork/kwil-db/pull/1534/commits/586e4bc651021649fd8fa419fc5683519b8fd631) && bug author @brennanjl

node/utils/syncmap/map.go
```diff
@@ -49,14 +49,14 @@ func (m *Map[K, V]) Set(key K, value V) {
 
 // Extract removes and returns the key's value, and whether it was found.
 func (m *Map[K, V]) Extract(key K) (value V, ok bool) {
-	m.mu.RLock()
+	m.mu.Lock()
```
ukane-philemon, line 52: Consider adding a ‘defer m.mu.Unlock()’, that way easier it’s to follow and reads better.

Then remove the two other calls to Unlock() below.

ukane-philemon (changes requested):
utAck.

node/utils/syncmap/map.go
brennanjl: Done!

MicBun (approved):
lgtm

PR #1536, opened 2025-07-08
feat: support extension loading during Kwil Testing

MicBun:
## Related Issue

resolves: https://github.com/trufnetwork/kwil-db/issues/1535

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update (including inline docs)

## Checklist:

- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.


testing/testing_extension_support.go
```diff
@@ -0,0 +1,148 @@
...
+				INSERT INTO kwild_engine.extension_initialization_parameters
+				(extension_id, key, value, scalar_type, is_array, metadata)
+				VALUES ($1, $2, $3, $4, $5, $6)
+			`, extensionID, key, valueStr, scalarType, isArray, nil)
+			if err != nil {
+				return fmt.Errorf("failed to insert parameter %s: %w", key, err)
+			}
+		}
+	}
+
```
ukane-philemon, line 102: @MicBun since we are trying to replicate what happens here, https://github.com/trufnetwork/kwil-db/blob/main/node/engine/interpreter/sql.go#L371-#L424

Would it not be better to export what needs to be tested instead of copying selectively?

The `ExtensionConfig` type already exists there as `storedExtension`, but `ExtensionConfig` reads better.

IMO, the original function behaviour might change in the future, and other copies might not be updated, plus it'll take more time to do so.

We also need to export `unregisterExtension` so that a complete test case, from entry to deletion, can be added for extensions.

ukane-philemon (commented):
Good start.

testing/testing_extension_support.go
MicBun: nice hint, will take a look at it in a few hours

PR #1537, opened 2025-07-08
chore: export type map utility

outerlook:
## Description

Exporting `DecodeFromPG` and `OidTypesMap` is very helpful for fixing tx problems in https://github.com/trufnetwork/node/pull/1036. Otherwise it would be a burden to maintain synchronization between our typings.

This PR doesn't touch anything other than exporting those functions


ukane-philemon (approved):
Was wondering if we have test cases for the new exported methods.

MicBun (approved):
lgtm

PR #1538, opened 2025-07-09
fix: race condition in consensus engine state info

brennanjl:
This PR fixes two race conditions in the consensus engine. The first race condition involves using potentially stale data in `QueueTx` if a block is concurrently being committed. The second race condition involves modifying state info without a lock.

EDIT:
I have since found another race condition in a similar line of code, and have added a fix. The race condition involved accessing a map in a concurrent goroutine without acquiring a lock.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved internal locking mechanisms to enhance stability and prevent potential deadlocks during consensus operations. No user-facing changes.

### DIFF
--- a/node/consensus/block.go
+++ b/node/consensus/block.go
@@ -145,9 +145,7 @@ func (ce *ConsensusEngine) QueueTx(ctx context.Context, tx *types.Tx) error {
 	// and send a trigger to the CE if it's in the waiting state to start the new round.
 	if ce.role.Load() == types.RoleLeader {
 		ce.stateInfo.mtx.RLock()
-		status := ce.stateInfo.status
-
-		if status != Committed {
+		if ce.stateInfo.status != Committed {
 			// we do not defer r.Unlock because we dont want to hold a read
 			// lock through the cs.mempoolReadyChan <- struct{}{} call
 			ce.stateInfo.mtx.RUnlock()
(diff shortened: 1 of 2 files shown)


node/consensus/block.go
```diff
@@ -145,14 +145,15 @@ func (ce *ConsensusEngine) QueueTx(ctx context.Context, tx *types.Tx) error {
...
-		if status != Committed {
+		if ce.stateInfo.status != Committed {
+			// we do not defer r.Unlock because we dont want to hold a read
+			// lock through the cs.mempoolReadyChan <- struct{}{} call
+			ce.stateInfo.mtx.RUnlock()
 			// send the mempoolReady trigger only during the
 			// newRound and waiting for blkProposal Timeout to elapse.
 			return nil
 		}
+		ce.stateInfo.mtx.RUnlock()
```
ukane-philemon, line 156: IMO, the new changes (https://github.com/trufnetwork/kwil-db/pull/1538/commits/bc5e19a84a9cafad4dc3c6928446f385551790d5) introduce more "complexity", more LOC, and it's a recommended practice to keep mutex lock and unlock close so it's easy to read and understand what is being protected.

Wdyt, @MicBun?

brennanjl: This might introduce _slightly_ more complexity, but IMO they are necessary.

The two options are:
- leave the vulnerability
- hold the lock for much longer than needed. Typically, this is not an issue, but since the code that it would be waiting on is relying on both an atomic swap and a channel, this could cause some really unintuitive bugs, since we are (unnecessarily) blocking progression of other parts of unknown code based on _other_ parts of unknown code

I've just re-arranged and added some comments to try to make it seem clearer.

MicBun: Great insight, LGTM

MicBun (approved):
LGTM

PR #1539, opened 2025-07-10
fix: race condition in listener manager

brennanjl:
This PR fixes a race condition in the listener manager.

If the node is started and then immediately stopped, there is a case where the listener manager can be stopped before it has truly started. This would cause a panic, as it would try to call a nil function.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability by preventing potential crashes when stopping listeners before they are started.
  * Enhanced thread safety to avoid issues during concurrent start and stop operations for listeners.

### DIFF
--- a/node/listeners/mgr.go
+++ b/node/listeners/mgr.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"slices"
+	"sync"
 	"time"
 
 	common "github.com/trufnetwork/kwil-db/common"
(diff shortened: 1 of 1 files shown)


brennanjl:
@pr-time-tracker bug commit [link](https://github.com/trufnetwork/kwil-db/pull/1539/commits/1f2eea0cb47c53ca1678ef491c39706b1584250c) && bug author @brennanjl

MicBun (approved):
LGMT

PR #1555, opened 2025-08-01
fix: prevent permanent announcement of bad transactions

MicBun:
Orphaned nodes that fall behind the main chain would previously accept
transactions against stale blockchain state and continuously reannounce
them every 30 seconds indefinitely. This occurred because:

1. During block sync, orphaned nodes receive ErrBlkNotFound from peers
2. They incorrectly assume sync is complete and exit the sync loop  
3. They accept new transactions validated against stale state
4. RecheckTxs is never called to clean up invalid transactions
5. Transactions are reannounced every 30 seconds forever

This fix adds a RecheckTxs call when block sync fails with ErrBlkNotFound
or ErrNotFound, ensuring stale transactions are validated against current 
state and removed if invalid.

Changes:
- Add RecheckTxs call in replayBlockFromNetwork when sync fails
- Add comprehensive unit tests verifying the fix works
- Tests confirm RecheckTxs is called and stale transactions are removed

Fixes permanent bad transaction announcements from orphaned nodes.

resolves: https://github.com/trufnetwork/kwil-db/issues/1554

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of missing blocks during synchronization by revalidating mempool transactions, ensuring stale or invalid transactions are removed automatically.

* **Tests**
  * Added new tests to verify that mempool transactions are correctly rechecked and cleaned up when block retrieval errors occur during synchronization.

### DIFF
--- a/node/consensus/blocksync.go
+++ b/node/consensus/blocksync.go
@@ -121,6 +121,10 @@ SYNC:
 			}
 
 			if errors.Is(err, types.ErrBlkNotFound) || errors.Is(err, types.ErrNotFound) {
+				// Recheck all mempool transactions to remove any that are no longer valid
+				// This handles the case where we might be orphaned with stale transactions
+				lh, t := ce.lastBlockInternal()
+				ce.blockProcessor.RecheckTxs(ctx, lh, t)
 				break SYNC // no peers have this block, assume block sync is complete, continue with consensus
 			}
 
--- a/node/consensus/blocksync_test.go
+++ b/node/consensus/blocksync_test.go
@@ -1,0 +1,265 @@
+//go:build pglive
+
+package consensus
+
+import (
+	"context"
+	"math/big"
+	"testing"
+	"time"
+
+	"github.com/trufnetwork/kwil-db/core/crypto/auth"
+	"github.com/trufnetwork/kwil-db/core/log"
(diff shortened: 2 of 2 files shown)


MicBun:
@pr-time-tracker bug commit not caused by previous commit on current maintener

outerlook (approved):
Hey @MicBun, I'll approve this as it really solves an issue, but I need to leave a note that *maybe* this doesn't address our observed issue on our network. Probably the issue description misled the fix.

But see if this makes sense:

1/Before the fork, all nodes were operational, although stale, and receiving transactions with certain nonces.
2/ after the fork, these nodes believed they were on the same network, continuing to send their transactions that are now invalid for other nodes
3/ These old nodes are orphaned (`ErrNoBlock`). So they just can't advance to a new true block and know what a correct nonce is. 

So, the thing is that for these nodes, their old transactions are still valid, so they won't delete these transactions from their mempools during new calls of `RecheckTxs`. The issue was misleading to believe that this could suffice for this particular case.

However, this PR addresses a real bug, which is why I think it's good to have it too. But do these make sense?

not sure what would be the exact fix for this situation... Maybe the node asks for other random nodes about their txs validity (carefully not to introduce attack vectors)? Or receive a response when they announce transactions? Or they just stop emitting if they are orphans. Not sure

PR #1574, opened 2025-08-11
chore: auto-blacklist peers that exhaust connection retries

MicBun:
Automatically blacklist peers that exhaust connection attempts to prevent infinite retry cycles and resource waste from PEX rediscovery.

- Add auto-blacklist logic to maintainMinPeers() retry exhaustion handler
- Permanently blacklist peers with reason "connection_exhaustion" when maxedOut()
- Only activate when both blacklist.enable and auto_blacklist_on_max_retries are true
- Add comprehensive test coverage for all configuration scenarios
- Verify blacklisted peers are rejected by connection gating (prevents PEX re-addition)

resolves: https://github.com/trufnetwork/kwil-db/issues/1562

### DIFF
--- a/node/peers/peers.go
+++ b/node/peers/peers.go
@@ -328,6 +328,13 @@ func (pm *PeerMan) maintainMinPeers(ctx context.Context) {
 				if !bk.try() {
 					if bk.maxedOut() {
 						pm.log.Warnf("Failed to connect to peer %s (%v) after %d attempts", peerIDStringer(pid), pid, bk.attempts)
+
+						// Auto-blacklist peer that exhausted connection retries if enabled
+						if pm.blacklistConfig.Enable && pm.blacklistConfig.AutoBlacklistOnMaxRetries {
+							pm.log.Warnf("Auto-blacklisting peer %s due to connection exhaustion after %d attempts", peerIDStringer(pid), bk.attempts)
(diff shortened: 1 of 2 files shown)


node/peers/peers.go
```diff
@@ -328,7 +328,14 @@ func (pm *PeerMan) maintainMinPeers(ctx context.Context) {
 				if !bk.try() {
 					if bk.maxedOut() {
 						pm.log.Warnf("Failed to connect to peer %s (%v) after %d attempts", peerIDStringer(pid), pid, bk.attempts)
-						pm.removePeer(pid)
+
+						// Auto-blacklist peer that exhausted connection retries if enabled
+						if pm.blacklistConfig.Enable && pm.blacklistConfig.AutoBlacklistOnMaxRetries {
+							pm.log.Warnf("Auto-blacklisting peer %s due to connection exhaustion after %d attempts", peerIDStringer(pid), bk.attempts)
+							pm.BlacklistPeer(pid, "connection_exhaustion", 0) // permanent blacklist (also removes peer)
```
outerlook, line 335: not sure if I agree that it should be permanent... thinking about situations where a peer is just disconnected for bad network, etc. maybe better to have something like 1 hours or something

but we should also stay wary after we implement that to be sure this doesn't happen very frequently with fine nodes

outerlook (approved):

PR #1585, opened 2025-08-19
feat: use @leader contextual variable

MicBun:
feat: use @leader contextual variable

Implements @leader contextual variable for secure leader-only authorization. Uses BlockContext.Proposer for deterministic leader identification, ensuring consistency with other contextual variable like @height.

Technical Implementation:
- Add @leader case to contextual variable system in interpreter
- Use BlockContext.Proposer as data source (same as @height, @block_timestamp)
- Return hex-encoded proposer public key as TEXT type
- Fail-safe behavior: empty string when proposer unavailable

Key Features:
- Deterministic: same block context always returns same leader
- Thread-safe: derived from immutable block execution context
- Zero dependencies: no interface changes or consensus modifications

Testing:
- Unit tests for core @leader functionality
- Multi-validator authorization scenario tests
- Edge case handling (nil proposer, invalid context)
- Consistency tests with other contextual variables
- Deterministic behavior verification

Files Modified:
- node/engine/interpreter/context.go: Added @leader contextual variable
- node/engine/interpreter/context_test.go: Comprehensive test coverage
- core/client/README.md: Updated contextual variables documentation

resolves: https://github.com/trufnetwork/kwil-db/issues/1583
resolves: https://github.com/trufnetwork/kwil-db/issues/1584

## Summary by CodeRabbit

* **New Features**
  * Introduced a new contextual variable: @leader, enabling leader-aware logic in actions.

* **Documentation**
  * Added a “Contextual Variables” section covering @caller, @txid, @height, and @leader with guidance and examples; note: the section was inserted twice.

* **Tests**
  * Added comprehensive tests for @leader covering determinism, interactions with other variables, authorization scenarios, proposer variants, and invalid-context behavior.

### DIFF
--- a/node/engine/interpreter/context.go
+++ b/node/engine/interpreter/context.go
@@ -555,7 +555,9 @@ func (e *executionContext) getVariable(name string) (Value, error) {
 			if e.engineCtx.InvalidTxCtx {
 				return nil, engine.ErrInvalidTxCtx
 			}
-			// Get leader from block context proposer
+			// Get leader from block context proposer.
+			// Returns lowercase hex of the proposer's public key bytes.
+			// Returns an empty string if the proposer is not available.
 			if e.engineCtx.TxContext.BlockContext.Proposer == nil {
 				return makeText(""), nil
 			}
--- a/core/client/README.md
+++ b/core/client/README.md
@@ -345,15 +345,18 @@ Actions have access to several built-in contextual variables that provide inform
 - **`@caller`** - The address of the account executing the action
 - **`@txid`** - The unique transaction identifier
 - **`@height`** - The current block height
-- **`@leader`** - The address of the current block leader (validator)
+- **`@block_timestamp`** - The timestamp (Unix seconds) of the current block
+- **`@leader`** - The current block proposer's public key as a lowercase hex-encoded string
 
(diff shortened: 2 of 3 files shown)


node/engine/interpreter/context.go
```diff
@@ -550,6 +550,20 @@ func (e *executionContext) getVariable(name string) (Value, error) {
 				return nil, engine.ErrInvalidTxCtx
 			}
 			return makeText(e.engineCtx.TxContext.Authenticator), nil
+		case "leader":
+			if e.engineCtx.InvalidTxCtx {
+				return nil, engine.ErrInvalidTxCtx
+			}
+
+			// Get current leader from the validators system
```
outerlook, line 558: Hey @MicBun, maybe we could not rely on a new `getCurrentLeader` code if this information is extracted from the same source as `@height`, i.e. `e.engineCtx.TxContext.BlockContext.Proposer`, which already exists

From reading this code, I believe it's bug-free, but it's better if we need no extra code

It should be safe, as blocks are always proposed by the leader of a block (as much as height is deterministic)

Or did you find any reason not to keep only that?

outerlook: but for sure there could be things here that I'm just not perceiving correctly yet

PR #1614, opened 2025-09-26
chore: improve snapshotter & fix protocol

### LINKED ISSUE #1526: Problem: Snapshot process is inefficient
for current data, snapshots is already a process that holds our node for several minutes, and we can improve it a lot with some tweaks. But this should be well tested to prevent any issue with non determinism

### Problem Summary

The current snapshot creation process (`node/snapshotter/snapshotter.go`) has severe performance bottlenecks, primarily caused by inefficient disk I/O patterns. For large databases (e.g., 100GB+), creating a snapshot can take hours. This impacts node operations and recovery times significantly.

### Root Cause Analysis

There are three major inefficiencies:

1.  **Random Disk Access During Sorting (CRITICAL):**
    *   In `sanitizeDump` (STAGE2), the code sorts table data (`COPY` blocks) by reading the file once, storing file offsets, sorting the offsets, and then using `f.Seek()` to re-read *every single row* in the new order.
    *   **Impact:** This performs millions of random disk operations. Random I/O is orders of magnitude slower than sequential I/O and is the primary bottleneck.

2.  **Excessive Intermediate Disk Writes:**
    *   Each stage (dump, sanitize, compress, chunk) writes its full output to disk before the next stage reads it.
    *   `pg_dump` -> `stage1.sql` -> `stage2.sql` -> `stage3.sql.gz` -> `chunks`.
    *   **Impact:** A 100GB database results in over 400GB of sequential writes and reads, adding significant unnecessary time.

3.  **Double I/O for Chunk Hashing:**
    *   In `splitDumpIntoChunks` (STAGE4), the code writes a chunk to a file, closes it, and then immediately reads the entire file back just to calculate its hash.
    *   **Impact:** Doubles the amount of I/O needed for the final chunking stage.

### Proposed Solution

We need to eliminate random I/O and reduce total sequential I/O.

1.  **Implement External Merge Sort for Sanitization:**
    *   **Action:** Replace the `f.Seek()` logic in `sanitizeDump` with an external merge sort.
    *   **How:** Read table data into a fixed-size memory buffer (e.g., 256MB). If the table exceeds the buffer, write the sorted buffer to a temporary file and repeat. Finally, merge all sorted temporary files sequentially.
    *   **Benefit:** Eliminates all random I/O. Ensures constant memory usage, preventing Out-Of-Memory (OOM) errors even with huge tables.

2.  **Pipeline Stages with `io.Pipe`:**
    *   **Action:** Stream the output of one stage directly to the input of the next using Go's `io.Pipe` and goroutines (Sanitize -> Compress -> Chunk).
    *   **Benefit:** Removes the need for large intermediate files (`stage2output.sql`, `stage3output.sql.gz`), cutting total sequential I/O by more than half.

3.  **Hash While Writing Chunks:**
    *   **Action:** Use `io.MultiWriter` to write data to both the final chunk file on disk AND the hash function simultaneously.
    *   **Benefit:** Eliminates the need to re-read the chunk file, halving the I/O for the final stage.

### Impact & ROI

Implementing these changes will yield massive performance gains:

*   **Speed:** Estimated 10x-100x faster. A process that takes hours or days could be reduced to minutes.
*   **Reduced Disk Load:** Total data moved (I/O) will decrease by ~60% (e.g., from ~580GB to ~220GB for a 100GB source).
*   **Scalability & Stability:** The process will be able to handle arbitrarily large databases using a fixed amount of memory, preventing OOM crashes.

### Alternatives Considered

1.  **Simple In-Memory Sort:**
    *   **Idea:** Load the entire table's data into memory instead of using file offsets.
    *   **Problem:** Unsafe. If a single table is larger than available RAM (e.g., a 50GB table on a 32GB machine), the node will crash (OOM).

2.  **Do Nothing:**
    *   **Problem:** The current implementation does not scale and is unacceptably slow for large datasets due to the random I/O bottleneck.

outerlook:
## Description

This PR introduces several enhancements to the snapshot handling process, focusing on improving type safety and streamlining functionality. Key changes include:

- Replaced hardcoded snapshot format values with `snapshotter.DefaultSnapshotFormat` for better maintainability.
- Added error handling for unsupported snapshot formats in the `LoadSnapshotChunk` method.
- Updated the `snapshotDB` function to simplify the logic for determining if snapshots are due.
- Enhanced the `SnapshotChunkReq` struct to ensure correct binary marshaling and unmarshaling, including adjustments to the byte layout.
- Introduced a new test file for `SnapshotChunkReq` to validate binary round-trip functionality and ensure data integrity.
- fix the bug in the protocol that ignored the format and included that bit of information in the hash (note, this is breaking!)

## Related Issue

- fix #1526

### DIFF
--- a/node/block_processor/processor.go
+++ b/node/block_processor/processor.go
@@ -745,8 +745,7 @@ var (
 )
 
 func (bp *BlockProcessor) snapshotDB(ctx context.Context, height int64) error {
-	snapshotsDue := bp.snapshotter.Enabled() &&
-		(bp.snapshotter.IsSnapshotDue(uint64(height)) || len(bp.snapshotter.ListSnapshots()) == 0)
+	snapshotsDue := bp.snapshotter.Enabled() && bp.snapshotter.IsSnapshotDue(uint64(height))
 	// snapshotsDue = snapshotsDue && height > max(1, a.cfg.InitialHeight)
 
 	if snapshotsDue {
--- a/node/snapshotter/handler.go
+++ b/node/snapshotter/handler.go
@@ -26,7 +26,7 @@ const (
 	snapshotGetTimeout      = 45 * time.Second
 
 	ProtocolIDSnapshotCatalog protocol.ID = "/kwil/snapcat/1.0.0"
-	ProtocolIDSnapshotChunk   protocol.ID = "/kwil/snapchunk/1.0.0"
+	ProtocolIDSnapshotChunk   protocol.ID = "/kwil/snapchunk/1.1.0"
 	ProtocolIDSnapshotRange   protocol.ID = "/kwil/snaprange/1.0.0"
 	ProtocolIDSnapshotMeta    protocol.ID = "/kwil/snapmeta/1.0.0"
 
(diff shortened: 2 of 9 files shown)


MicBun (approved):
lgtm

PR #1629, opened 2025-10-22
chore: skip event votes on catchup

### LINKED ISSUE #1628: Bug: Node tries to vote during catch up
When the nodes try to vote on a listened event during catchup, it crashes.

That kind of situation surfaced now that we have listeners who need to vote on new events, and it's not normal because usually the listener pauses during catch-up, but that bug surfaces if the node is stopped with an event, and later it is started again

```
node stopped with error: error catching up: failed to apply block at height: 345834: error: error executing block: failed to broadcast the voteID transactions: node is catching up, cannot process transactions right now
```

outerlook:
## Description

This commit updates the block processor to prevent broadcasting of voteID transactions when the node is in catch-up or synchronization mode. The change ensures that the node does not attempt to process new transactions during these states, enhancing stability and preventing potential errors.

Changes:
- Added a condition to check the syncing state before broadcasting voteID transactions.

This improvement contributes to more reliable block execution during synchronization phases.

## Related Issue

- fix #1628 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where vote transactions were being broadcast during node synchronization. Transactions are now correctly withheld while the node is catching up with the network, preventing unnecessary network traffic and ensuring proper synchronization behavior.

### DIFF
--- a/node/block_processor/processor.go
+++ b/node/block_processor/processor.go
@@ -481,7 +481,8 @@ func (bp *BlockProcessor) ExecuteBlock(ctx context.Context, req *ktypes.BlockExe
 	bp.recordBlockExecEndTime()
 
 	// Broadcast any voteID events that have not been broadcasted yet
-	if bp.broadcastTxFn != nil {
+	// Skip broadcasting during catch-up/sync since the node cannot process new transactions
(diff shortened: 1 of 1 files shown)


MicBun (approved):
lgtm

PR #1705, opened 2026-06-02
chore: encode agent-wallet execution transactions

MicBun:
Partially toward letting an agent key operate its wallet within limits:

resolves: https://github.com/truflation/website/issues/4036

Adds the `MAAExec` transaction payload: the wire format for executing a single inner action as a Modular Agent Address (MAA). It is the foundation the execution route builds on. The route lands in a stacked follow-up PR that carries the closing keyword for #4036.

## Contents
- `PayloadTypeMAAExec` (`"maa_exec"`) and `MAAExec{ MAAAddress, Namespace, Action, Arguments }`, with little-endian `Marshal`/`UnmarshalBinary` modeled on `ActionExecution` (a single call, not batched).
- Registered in `payloadConcreteTypes`, `payloadTypes`, and `Valid()` so the type decodes and is recognized across the node.
- Golden-vector and round-trip tests that freeze the wire layout (the language SDKs mirror these exact bytes).

No route is registered in this PR, so a `maa_exec` transaction has no handler and is inert (rejected as an unroutable type) until the follow-up. No behavior change for existing transactions.

### DIFF
--- a/core/types/payloads.go
+++ b/core/types/payloads.go
@@ -51,6 +51,9 @@ const (
 	PayloadTypeCreateResolution    PayloadType = "create_resolution"
 	PayloadTypeApproveResolution   PayloadType = "approve_resolution"
 	PayloadTypeDeleteResolution    PayloadType = "delete_resolution"
+	// PayloadTypeMAAExec executes a single inner action as a Modular Agent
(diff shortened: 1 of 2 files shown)


MicBun:
@holdex pr submit-time 4h

PR #1706, opened 2026-06-03
feat: run allow-listed actions as an agent wallet

MicBun:
resolves: https://github.com/truflation/website/issues/4036

Builds on the merged payload PR:

- https://github.com/trufnetwork/kwil-db/pull/1705

Adds `maaExecRoute`, which lets an authorized signer run a single allow-listed action **as** a Modular Agent Address (MAA), with `@caller` rewritten to the MAA for the inner call. This is the mechanism that lets an agent key operate its wallet within limits.

## Behavior
- The MAA must be a known, joined instance (created by `maa_join`); an unknown address is rejected.
- The outer signer must be the rule's restricted key (the agent) **or** the instance's unrestricted key (the owner). The decision is by **raw-byte** address comparison, because engine identifiers are EIP-55 checksummed while the rule store emits lowercase hex — a string compare would be a consensus bug.
- The inner call is **top-level**, so SYSTEM-only ops (mint / issue / lock_admin / unlock) stay blocked even though `@caller` is now the MAA.
- Gas and nonce are paid by the outer signer; the MAA never pays gas.

## Consensus-breaking
This registers the active route for the `maa_exec` payload (the payload type itself merged in #1705). Network-wide activation is by binary deployment for now.

## Tests
Fake-engine unit tests cover every branch: restricted and unrestricted happy paths (asserting `@caller` is rewritten to the MAA and arguments pass through), empty-namespace defaulting to `main`, and the three rejections — unknown MAA, unauthorized signer, and non-allow-listed action (each asserting the inner action does not run).

### DIFF
--- a/node/txapp/maa_route.go
+++ b/node/txapp/maa_route.go
@@ -1,0 +1,207 @@
+package txapp
+
+import (
+	"bytes"
+	"context"
+	"encoding/hex"
+	"fmt"
+	"math/big"
+	"strings"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/trufnetwork/kwil-db/common"
+	"github.com/trufnetwork/kwil-db/core/types"
+	"github.com/trufnetwork/kwil-db/extensions/consensus"
+	"github.com/trufnetwork/kwil-db/node/engine"
+)
+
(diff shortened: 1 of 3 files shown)


MicBun:
@holdex pr submit-time 4h

PR #1710, opened 2026-06-11
chore: hold maa_exec until a network-scheduled activation height

MicBun:
resolves: https://github.com/truflation/website/issues/4117

Activating the `maa_exec` route on a running network is currently a flag-day: the route is live the moment a binary carrying it is deployed, so a mixed-version fleet forks — an old binary skips an in-block `MAAExec` with `CodeInvalidTxType` while a new binary executes it, diverging TxResults and changesets. This PR makes activation a network-agreed, height-scheduled event.

## What this does

- **`maa_activation_height` consensus parameter** (`core/types/params.go`). `0` (the default) means `maa_exec` is not activated; a non-zero `H` activates the route from block `H` on. New networks set it in genesis (`kwild setup genesis --maa-activation-height`); live networks schedule it with the stock resolution flow (`kwild consensus propose -u '{"maa_activation_height": H}'`).
- **The zero value is invisible to consensus.** The params hash is carried in every block header (`Header.NetworkParamsHash`, validated in `consensus/block.go`), and the params JSON is rewritten into `kwild_chain.params` inside every hashed block transaction. The field is therefore `omitempty` in the stored JSON and skipped in `Hash()` while zero, so a binary with this PR is byte-identical in consensus, block for block, to one without it — until the network explicitly schedules activation. The binary upgrade itself is hash-neutral; the scheduling resolution is the only consensus-visible event, and a straggler node still on an old binary halts cleanly on header validation instead of forking silently. `TestNetworkParametersZeroMAAActivationHeightIsInvisible` pins this with goldens captured from the pre-change code over mainnet's actual parameter values.
- **Dual gate.** Mempool admission (`applyTransaction`, beside the existing migration gates) keeps pre-activation `MAAExec` out of mempools and blocks entirely; `maaExecRoute.PreTx` (the activation chokepoint marked since the route landed) fails one deterministically on every node if a leader force-includes it anyway. Both wrap `ErrUnknownPayloadType`, so the rejection reads exactly like a pre-MAA binary's answer to this payload type.

## How it is verified

- Unit: gate matrix (fail-closed without context / never activated / below height / at and after height), mempool admission tests, and the existing route suite under an activated context.
- `test/integration/maa_activation_test.go` (`TestMAAExecActivation`) rehearses the full rollout on a real 3-validator network, driving the literal `maaExecRoute` with raw hand-signed `MAAExec` transactions: genesis without activation → broadcast rejected; activation scheduled via a `param_updates` resolution, every node converging on `H`; rejected below `H`; from `H` on the inner action executes as the MAA (`@caller` rewritten, identical state on all three nodes); a signer that is neither of the instance's keys fails deterministically; and the AppHash is asserted equal across nodes at every height through the activation fork.

## Summary by CodeRabbit

* **New Features**
  * Added a network parameter `maa-activation-height` to configure when MAA execution becomes active.
  * Implemented activation-height gating to enforce MAA transaction availability after the configured activation height.

* **Tests**
  * Added comprehensive test coverage for MAA activation behavior, including mempool admission gating and end-to-end integration testing.

* **Chores**
  * Updated test module dependencies to latest compatible versions.

### DIFF
--- a/app/setup/genesis.go
+++ b/app/setup/genesis.go
@@ -40,12 +40,13 @@ type genesisFlagConfig struct {
 }
 
 type networkParams struct {
-	withGas       bool
-	leader        string
-	dbOwner       string
-	maxBlockSize  int64
-	joinExpiry    time.Duration
-	maxVotesPerTx int64
+	withGas             bool
+	leader              string
+	dbOwner             string
+	maxBlockSize        int64
+	joinExpiry          time.Duration
+	maxVotesPerTx       int64
+	maaActivationHeight int64
 }
 
 func GenesisCmd() *cobra.Command {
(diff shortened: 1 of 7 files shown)


MicBun:
@holdex pr submit-time 4h

PR #1711, opened 2026-06-11
test: rehearse the maa activation rollout on a real network

MicBun:
resolves: https://github.com/truflation/website/issues/4117

Rehearses the coordinated MAA rollout end to end on a real 3-validator network, driving the literal `maaExecRoute` with raw hand-signed `MAAExec` transactions — the route's first end-to-end exercise over real consensus.

## What it proves

- A network genesised without activation (the parameter at its zero default) rejects `MAAExec` at broadcast, reading like an unknown payload type.
- Activation is scheduled with the production `param_updates` resolution flow (propose + approve), and every node converges on the scheduled height.
- Below the scheduled height the same transaction is still rejected; from the height on, the inner action executes as the MAA — `@caller` rewritten, with identical resulting state on all three nodes.
- A signer holding neither of the instance's keys passes the activation gate but fails authorization in the route, deterministically.
- The AppHash is asserted equal across all nodes at every height through the activation fork — the determinism sweep for the rollout.

The synthetic `maa_get_instance` / `maa_get_allowed_actions` actions mirror the production rule-store getters byte for byte in signature and projection (BYTEA params, `'0x'`-hex text columns). Verifies the activation gating shipped in:

- <https://github.com/trufnetwork/kwil-db/pull/1710>


MicBun:
@holdex pr submit-time 2h

PR #1713, opened 2026-06-19
chore: keep crediting bridge deposits when an RPC caps getLogs

MicBun:
The EVM bridge listener catches up by requesting eth_getLogs over [lastSeen, head] in chunks of block_sync_chuck_size (default 1,000,000). When that range exceeds what the RPC provider allows, the provider rejects every request (e.g. "exceed maximum block range", "Request contains invalid block params"); the catch-up loop returns the error, stops, and then retries the identical oversized request forever. The listener never advances, so bridge deposits and withdrawals stop being picked up.

Make catch-up adaptive: on a failed chunk, halve the chunk size and retry the same range instead of giving up. The retry is bounded -- once the chunk is a single block and still fails, the error is real and is surfaced, so a genuine outage still triggers the normal recovery/backoff. A mis-sized or unset block_sync_chuck_size can no longer wedge the listener.

The split is consensus-safe: sub-ranges are contiguous and processEvents chains via the persisted lastHeightWithEvent/lastSeenHeight, so the ordered resolution sequence is identical regardless of how the range is chunked.

Covered by TestSyncBlockChunks_AdaptiveChunking: halves on an oversized range, surfaces an error instead of looping forever, and respects the configured chunk size when there is no provider limit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved block synchronization reliability with adaptive retry logic that automatically reduces request chunk sizes on failures, better handling RPC provider limitations and preventing sync stalls.

* **New Features**
  * Added configurable delay parameter to control pause intervals during block catch-up operations.

### DIFF
--- a/node/exts/evm-sync/instance.go
+++ b/node/exts/evm-sync/instance.go
@@ -263,12 +263,13 @@ func (l *listenerInfo) listen(ctx context.Context, service *common.Service, even
 		}
 
 		indiv := &individualListener{
-			chain:            l.chain,
-			syncConf:         syncConf,
-			chainConf:        chainConf,
-			client:           ethClient,
-			orderedSyncTopic: l.uniqueName,
-			getLogsFunc:      l.getLogs,
+			chain:               l.chain,
+			syncConf:            syncConf,
+			chainConf:           chainConf,
+			client:              ethClient,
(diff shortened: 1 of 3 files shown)


MicBun:
@holdex pr submit-time 3h

PR #1714, opened 2026-07-08
build: raise postgres work_mem to keep sorts in memory

MicBun:
resolves: https://github.com/truflation/website/issues/4272

Adds `-c work_mem=64MB` to the kwil-postgres image CMD so large sorts and hash joins run in memory instead of spilling to disk (the Postgres default is 4MB). Already applied at runtime on the mainnet nodes; this makes it the default for all deployments that use the image.

## Summary by CodeRabbit

* **Chores**
  * Updated the PostgreSQL container startup settings to use a higher memory work setting, which may improve database performance for some workloads.


MicBun:
@holdex pr submit-time 2h

PR #1715, opened 2026-07-16
perf: keep top-holder queries fast at scale

MicBun:
## What

Adds a Postgres index `idx_balances_reward_balance` on `kwil_erc20_meta.balances(reward_id, balance)` and bumps the erc20 meta-schema version so existing nodes pick it up on upgrade.

## Why

Ordered balance lookups — `WHERE reward_id = $id [AND balance >= $min] ORDER BY balance LIMIT $n` — currently do a sequential scan + top-N sort on `balances`. This backs the upcoming node "ordered balances / richlist" action, and the index keeps those queries fast as a token's holder count grows.

Kuneiform's `CREATE INDEX` grammar has no per-column `DESC`, but a plain `(reward_id, balance)` btree is scanned backward for `ORDER BY balance DESC`, forward for `ASC`, and also serves the `balance >= threshold` range — so it covers every pattern without a separate sort.

## How it ships

Additive and `IF NOT EXISTS`. `currentVersion` 2 → 3 makes `OnStart` re-run the idempotent `createSchema` on existing nodes. Index creation is consensus-neutral (indexes don't affect app hash).

## Test

New `TestBalancesRewardBalanceIndex` asserts the index exists on `kwil_erc20_meta.balances` after schema creation. The full erc20 package suite passes.

## Related

- resolves: trufnetwork/trufscan#185

## Summary by CodeRabbit

- **Performance Improvements**
  - Improved performance for top-holder lookups and balance-threshold queries.
  - Ordered balance results can now be retrieved more efficiently, especially on larger datasets.
- **Bug Fixes**
  - Existing deployments automatically receive the performance enhancement during schema updates without affecting stored balances.

### DIFF
--- a/node/exts/erc20-bridge/erc20/meta_schema.sql
+++ b/node/exts/erc20-bridge/erc20/meta_schema.sql
@@ -32,6 +32,12 @@ CREATE TABLE IF NOT EXISTS balances (
     CONSTRAINT balance_must_be_positive CHECK (balance >= 0)
 );
 
+-- index for ordered top-holder lookups: WHERE reward_id = $id [AND balance >= $min]
+-- ORDER BY balance {ASC|DESC} LIMIT $n. Without it, ordering a token's balances is a
+-- sequential scan + top-N sort. A btree on (reward_id, balance) is scanned forward for
+-- ASC and backward for DESC, and also serves the balance-threshold range.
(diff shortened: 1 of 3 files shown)


MicBun:
@holdex pr submit-time 4h

PR #1729, opened 2026-09-17
fix: tell a peer with PEX off apart from a failing one

### LINKED ISSUE #1719: Bug: pex=false peers contribute no statesync snapshot providers
**Version:** kwild v2.5.7 (`kwild version` -> 0.10.1), chain `tn-v2.1`, sentry role (power 0)

### Question for maintainers

Is `pex = false` intentional on `node-1.mainnet.truf.network`? Either answer points somewhere useful, and I can't determine it from outside - it isn't exposed on any public endpoint (the `mode` field in `user.health` is the private-mode flag, not PEX).

### Presenting symptom

```
[WRN] PEERS: Failed to get peers from 4e0b5c952be7f26698dc1898ff3696ac30e990f25891aeaf88b0285eab4663e1#ed25519: failed to read and decode peers: EOF
```

Repeats every ~20s indefinitely. Peer exchange with node-2 (`0c830b69...`) works normally; only node-1 does this.

### What's ruled out

**Not transport.** TCP to `node-1.mainnet.truf.network:26656` is open, verified independently from a separate network. The libp2p connection succeeds moments earlier:

```
[INF] PEERS: Connected to peer (Outbound) 4e0b5c95...#ed25519 @ /ip4/3.134.167.133/tcp/26656
```

**Not a node-1 outage.** Probing from an unrelated host, node-1 is healthy and in consensus - `user.health` reports `healthy: true`, `syncing: false`, height 2554061 with app hash `94c751d1...`, identical to node-2.

**Not version skew.** `node/peers/pex.go`, which contains the entire wire format (`peersMsg`, `writePeers`, `recvPeersProto`), has exactly one commit in its history (`dc839a13`, 2025-01-07). `node/peers/` as a whole is unchanged since 2025-09-25. There is no wire-format difference between v2.5.4 and v2.5.7. A format change would also surface as a JSON type error, not a clean `EOF`.

**Not blacklisting or gating.** The connection gater intercepts only at connection level (`InterceptPeerDial`, `InterceptSecured`, `InterceptAccept` in `node/peers/gate.go`); there is no stream-level gating. A blacklisted peer is refused the connection outright, not accepted and then closed.

**Not a size limit, timeout, empty list, or rate limit.** An empty peer list still encodes valid JSON (`{"chain_id":...,"peers":null}`). The 1 MB `io.LimitReader` and the 4s write deadline would both yield `unexpected EOF`, not `EOF`. `DiscoveryStreamHandler` has no rate limiting.

### Root cause

`NewPeerMan` registers the discovery handler unconditionally, stubbing it out when PEX is off (`node/peers/peers.go:193`):

```go
if cfg.PEX || cfg.SeedMode {
    host.SetStreamHandler(ProtocolIDDiscover, pm.DiscoveryStreamHandler)
} else {
    host.SetStreamHandler(ProtocolIDDiscover, func(s network.Stream) {
        s.Close()          // zero bytes written
    })
}
```

Because the handler is registered either way, `/kwil/discovery/1.0.0` is advertised via identify, so the caller's `NewStream` succeeds. The peer then closes having written nothing, and `recvPeersProto`'s `json.Decoder.Decode` (`node/peers/pex.go:114`) returns a clean `EOF`. This is the only path in the codebase producing a zero-byte close on a successfully negotiated discovery stream.

### Why this matters beyond log noise

Statesync snapshot-provider discovery runs **entirely over PEX**. There is no DHT behind it despite the filename:

```
DiscoverSnapshots -> discoverProviders(ctx, snapshotCatalogNS, s.discoverer)   // node/statesyncer.go
discoverProviders -> d.FindPeers(ctx, namespace)                               // node/dht.go:58
```

`s.discoverer` is `P2PService.discovery` (`node/statesync.go:101`), which is the `PeerMan` - and `PeerMan.FindPeers` is the PEX request fan-out.

So **a peer with `pex = false` contributes zero snapshot providers to a bootstrapping node.** If that is node-1's configuration, then the node documented as the canonical statesync provider contributes nothing to snapshot-provider *discovery*. It still functions as a `trusted_provider`, since that's a static config entry dialed directly, and still serves catalogs and chunks once found - but a fresh node's discovery pool comes from node-2's PEX response alone.

That thinner pool interacts badly with the unbounded verification retry in #1718.

Secondary effect: the caller cannot distinguish "this peer doesn't do PEX" from "this peer is broken," so `startPex` -> `FindPeers` re-dials every connected peer every 20s forever and warns each time. That cost real diagnostic time here - a healthy, in-consensus node looks like a failing one.

### Reproduction

Start node A with `pex = true` and node B with `pex = false`, and connect A to B. A logs the warning every 20s indefinitely. With B as A's only peer, A's statesync provider discovery returns empty regardless of what snapshots B holds.

### Suggested fix

Don't register `ProtocolIDDiscover` when PEX is disabled, so `NewStream` fails fast with an unambiguous "protocol not supported" - and have the caller record non-participating peers rather than re-asking every cycle. Separately, it's worth deciding whether snapshot-provider discovery should depend on PEX at all, given that PEX is an operator-tunable setting and statesync bootstrapping is not.

MicBun:
resolves: https://github.com/trufnetwork/kwil-db/issues/1719

An operator reading the logs cannot tell a peer that has peer exchange switched off from one that is failing. Both produce the same line, every 20 seconds, for as long as the peer stays connected:

```
[WRN] PEERS: Failed to get peers from 4e0b5c95...: failed to read and decode peers: EOF
```

Our mainnet leader logged 1,080 of them in six hours, all from one healthy, in-consensus node. That cost real diagnostic time, which is what the issue was filed about.

## The suggested fix is not available

The issue proposes not registering `ProtocolIDDiscover` when PEX is off, so the caller gets an unambiguous "protocol not supported". That protocol is in `RequiredStreamProtocols`, and two places act on that list:

- `node/node.go:334` closes the connection to any connected peer missing a required protocol, at startup.
- `node/peers/peers.go:1125` tags such a peer `lame` so the connection manager prunes it first, and clears its addresses so it is neither advertised nor reconnected.

A node that stopped advertising peer exchange would be disconnected by every upgraded peer and dropped from their address books. On mainnet that is the node documented as the canonical statesync provider. Nothing about consensus changes, but that node stops hearing blocks over p2p until it is reachable again.

## What changed

**The serving side answers instead of hanging up.** A node with PEX off now writes a well-formed empty peer list. It keeps advertising the protocol, so the required-protocol check still passes, and the caller reads a valid answer rather than a zero-byte close.

**The receiving side tells the two cases apart.** A bare `EOF` becomes `ErrPeerServesNoPeers`, which both callers log at debug. This half needs no cooperation from the other operator, so it quiets a running network before anyone upgrades. A truncated message still arrives as `ErrUnexpectedEOF` and still warns.

Both halves are needed. The first fixes the peer that is misread; the second fixes the reader, against every peer still on an older release.

## Re-asking each cycle is deliberate

The issue also suggests recording non-participating peers so `startPex` stops re-asking. It costs one stream per peer per 20 seconds, and it is what picks a peer up if PEX is switched on later without a reconnect. A skip list trades that for state that has to be invalidated correctly. Left out.

## Tests

Each hunk was checked by breaking it and watching a named test fail.

| Broken | Result |
| --- | --- |
| PEX-off handler hangs up again | `TestDiscoveryStreamAnswersWhenPexIsOff` fails |
| zero-byte answer never classified | `TestRequestPeersClassifiesAPeerThatServesNothing` and the log test fail |
| every read error classified | `TestRequestPeersStillReportsATruncatedAnswer` fails |
| `FindPeers` warns anyway | `TestFindPeersDoesNotWarnAboutAPeerThatServesNothing` fails |
| `crawlPeer` warns anyway | `TestCrawlPeerDoesNotWarnAboutAPeerThatServesNothing` fails |

The log tests read the captured output and reproduce the exact line from the issue when the change is reverted. `./node/...` passes with a live postgres, `-race` is clean on `node/peers`, and `gofmt` and `go vet` are clean.

## Correcting the issue

The statesync half of the report does not hold, and the correction is already on the issue rather than in this diff. Snapshot-provider discovery runs over the Kademlia DHT, not peer exchange, so a peer with `pex = false` does contribute providers. What `pex = false` changes is one DHT option, `DisableAutoRefresh`. The log spam is the half that was real, and it is what this addresses.


MicBun:
@holdex pr submit-time 4h

MicBun:
@holdex bug commit https://github.com/trufnetwork/kwil-db/commit/41aa8672ce320573b303384f26d528e16f92f9ed && bug author @chappjc

PR #1737, opened 2026-09-17
chore: give a distant peer time to answer the first byte

### LINKED ISSUE #1731: Problem: block_sync settings never reach a catching-up node
An operator who edits `block_sync.*` in `kwild.toml` gets no effect on catch-up block sync. The values are read, parsed, and then dropped.

`node/config.go:20` declares `BlockSync *config.BlockSyncConfig`, and `NewNode` reads it at `node/node.go:250`. `buildNode` never sets it:

```go
nc := &node.Config{
    ChainID: ..., RootDir: ..., PrivKey: ..., DB: db,
    P2P:       &d.cfg.P2P,
    Statesync: &d.cfg.StateSync,
    DBConfig:  &d.cfg.DB,
    // no BlockSync
}
```

`app/node/build.go:579-594`

So `n.blockSyncCfg` is nil on every production node, and each `if n.blockSyncCfg != nil` falls through to a hardcoded constant at `node/block.go:523-530`: 2 s request, 20 s response, 500 ms idle.

The same struct does reach the state-sync service, at `app/node/build.go:247`. The knob works on the path that fetches a single block during statesync (`node/statesync.go:194`) and is dead on the path that fetches sixty-eight thousand.

Nothing looks broken today, because every shipped default is numerically identical to the constant it would replace. All nine values in `DefaultConfig()` (`config/config.go:366-376`) match `block.go:39-40`, `block.go:523-525`, and `protocol.go:94-102`. A node running shipped values behaves the same either way. A node whose operator changed a value is ignored without any signal.

## The 500 ms idle timeout is a whole-round-trip budget

`readAll` re-arms the deadline before every read, the first one included:

```go
for {
    s.SetReadDeadline(time.Now().Add(idleTimeout))
    n, err := s.Read(buf[len(buf):cap(buf)])
```

`node/block.go:437-450`

`requestBlockHeight` writes the request and enters `readAll` immediately (`node/block.go:391-395`). That first 500 ms therefore has to cover the request flight, the peer's block-store read, a decode, a re-encode, and the first byte coming back. At the 305 ms round trip measured in the parent Goal, the peer has under 195 ms left for its own work. When the deadline fires, `getBlkHeight` logs `unexpected error from %v: %v` (`node/block.go:562`) and moves to another peer, which costs a further round trip. How often it fires on a distant node has never been measured, because none of the three strings the catch-up path logs on failure appear in the greps run so far.

## What needs to change

- Set `BlockSync: &d.cfg.BlockSync` in the `node.Config` literal in `buildNode`.
- Give the first read of a response the `response_timeout` budget, and apply `idle_timeout` from the second read onward. The change belongs in `readAll` (`node/block.go:431`), which has one caller today.
- Before merging, check each fleet config for a `[block_sync]` value that differs from the shipped default, since those start taking effect.

Any experiment that tunes a `block_sync` value to test a hypothesis about catch-up is blocked until the wiring lands. Run one before that and it returns a false negative.

Blame:

- <https://github.com/trufnetwork/kwil-db/commit/ff963213> added `BlockSyncConfig`, added `node.Config.BlockSync`, wired the struct into the state-sync service, and did not wire it into `buildNode`.

Part of:

- <https://github.com/trufnetwork/kwil-db/issues/1730>

MicBun:
resolves: https://github.com/trufnetwork/kwil-db/issues/1731

A node far from its peers drops peers that were about to answer. `readAll` re-armed its deadline before every read, the first one included:

```go
for {
    s.SetReadDeadline(time.Now().Add(idleTimeout))
    n, err := r.Read(b[len(b):cap(b)])
```

`node/block.go:437-450`

`requestBlockHeight` writes the request and enters `readAll` immediately, so that first `idle_timeout` — 500 ms by default — had to cover the request's flight to the peer, the peer's block-store read, a decode, a re-encode, and the first byte coming back. At the 305 ms round trip reported in the parent Goal, the peer had under 195 ms left for its own work. When the deadline fired, `getBlkHeight` logged `unexpected error from`, moved to another peer, and spent a second round trip replacing one that was still in progress.

`idle_timeout` describes itself as "timeout between reading chunks of a block". Before the first byte there are no chunks to be between.

## What changed

The first read is bounded by the absolute deadline the caller already computes from `response_timeout`. `idle_timeout` applies from the second read onward, which is the stalled-transfer detector it was named for. One loop, in a function with a single caller.

## A silent peer now costs more

A peer that accepts a stream and then says nothing costs up to `response_timeout` instead of 500 ms. Bounded, and not the failover path that carries the weight: `blkGetHeightStreamHandler` runs under `defer s.Close()` with its own deadlines — 5 s to read the request, 45 s to write a block — so a peer that cannot serve closes the stream and the reader moves on immediately. The 500 ms only ever bit peers that were still working.

What is left is the true blackhole, where the connection is gone and neither side knows yet. That case already had the 20 s ceiling once any byte had arrived; this extends it to the case where none has.

## Tests

| Broken | Result |
| --- | --- |
| first read held to `idleTimeout` again | `a peer slower than idleTimeout is not abandoned` fails |
| first read left unbounded | `a peer that never answers still fails at the response deadline` fails |
| `idleTimeout` no longer applied mid-transfer | `a transfer that stalls mid-block still trips idleTimeout` fails |

The test peer answers on a schedule and honours whichever deadline the reader set for the read in progress, so the three cases differ only in when it speaks. The middle one passes before and after — it is there to show `idle_timeout` keeps its real job.

`./node/` passes, `-race` is clean over repeated runs, and so are `gofmt` and `go vet`.

## Sequencing

- <https://github.com/trufnetwork/kwil-db/pull/1736>

They merge in either order and do not touch the same lines. Until that one lands, the budget this switches to is the hardcoded 20 s rather than whatever the operator configured — which is the other half of the same Problem.


MicBun:
@coderabbitai review

MicBun:
@holdex pr submit-time 4h

PR #1742, opened 2026-09-22
chore: log the execution time the metric already records

MicBun:
Follow-up to a merged PR, from CodeRabbit findings that landed after the merge.

- <https://github.com/trufnetwork/kwil-db/pull/1741>

## The log field did not measure what the comment said

`#1741` added a comment stating that `duration` on the `Executed block` line covers `ExecuteBlock` and nothing after it. The code read `time.Since(now)` at the log call, which is after `mets.RecordExecuted`, the `blockResult` allocation and `catchupTicker.Reset`. The line was already like that; the comment claiming otherwise is what made it wrong.

`ce.state.tExecuted` is already read on the line after `ExecuteBlock` returns, and `mets.RecordExecuted` already derives the metric from it, so the log now uses that same subtraction:

```go
ce.state.tExecuted = time.Now()
execution := ce.state.tExecuted.Sub(t0)
mets.RecordExecuted(ctx, execution, ...)
...
ce.log.Info("Executed block", ..., "duration", execution)
```

`now` is gone, and no new timestamp replaces it.

The point is not the microseconds this removes. `duration` and `consensus.exec.latency` are now one number read once, rather than two measurements of nearly the same thing. The Problem behind `#1741` is an operator reconciling what the log says against what the histogram says, and those two could previously disagree.

## What `apply` measures, said properly

The second finding was that `syncProgress.applied` covers the wait for `ce.state.mtx` and the block decode as well as execute and commit, which its doc did not say.

The boundary stays where it is, and the doc now describes it. Decode is local CPU work on bytes already received; it runs on every block, and moving it out would push it into `elapsed - network - apply`, which `#1741` documents as time asleep between retries. That remainder is only readable if nothing else is hiding in it.

The lock wait is reachable but not realistic on this path. Every other writer of `ce.state.mtx` runs inside `consensusEventLoop`, the same goroutine that calls `doCatchup`, so none can contend with it. The exception is `resetBlockProp` on the `resetEventLoop` goroutine, which fires on a leader reset — something a node catching up is not taking part in rounds for. One of those inflates a single block in a single window.

## Tests

None added, and that is worth stating rather than glossing. Both changes are where a timestamp is read and what a comment claims. Proving the first would need injected delay inside `mets.RecordExecuted` or `catchupTicker.Reset`; proving the second would need lock contention around a full `processAndCommit`, which wants Postgres, validator signatures and a block store. Neither earns an API shaped for the test.

`./node/consensus/` and `./node/` pass against `ghcr.io/trufnetwork/kwil-postgres:16.8-2`, including all nine tests `#1741` added, and so do `gofmt`, `go vet` and `go build ./...`.

Part of:

- <https://github.com/trufnetwork/kwil-db/issues/1730>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved execution latency reporting for executed blocks and aligned log output with the recorded duration.

* **Documentation**
  * Clarified progress timing details, including the stages covered when applying a block during synchronization.


MicBun:
@holdex pr submit-time 1h

PR #1743, opened 2026-09-22
perf: serve a block without decoding and re-encoding it

MicBun:
Part of:

- <https://github.com/trufnetwork/kwil-db/issues/1733>

A node serving a block decoded it out of badger and encoded it straight back. `blkGetHeightStreamHandler` called `GetByHeight`, which runs `DecodeBlock` over every transaction in the block, and then ran `EncodeBlock` on the result to arrive back at the bytes badger was already holding. `blkGetStreamHandler` did the same for a request by hash.

`GetRawByHeight` has been on the `RawGetter` interface since [`1741858a`](https://github.com/trufnetwork/kwil-db/commit/1741858a) (Mar-2025), and `node/block.go` already used it for the JSON-RPC path. The serving path never moved over.

The response then left as five to seven separate `Write` calls, because `WriteCompactBytes` sends a length ahead of each variable piece, and every `Write` on a libp2p stream is a yamux frame of its own. The snapshot handler wraps its stream in a buffered writer; the block handler did not.

The decode and the re-encode are paid by whoever is serving, once per block per peer, which is what a public bootnode does all day. The framing costs both ends a handful of syscalls per block.

## What changed

**Both handlers serve the stored bytes.** `blkGetHeightStreamHandler` takes `GetRawByHeight`, `blkGetStreamHandler` takes `GetRaw`. `Store` writes what `EncodeBlock` produced, so these are the same bytes the old path arrived back at.

**`GetRaw` returns the height.** The lookup it already did to answer "do we have this" reads `map[Hash]int64`, so the height was sitting there being discarded. That is what `mets.BlockRetrieved` was passing `-1` for, and it is what a response to a request by hash has to write into the frame.

**One buffer per response.** A 16 KB `bufio.Writer` holds the framing and a block of ordinary size, so the peer reads one frame instead of seven. A block larger than that goes out in three whatever its size: a buffer's worth, the rest of the block straight to the stream, then the best height. The buffer bounds what gets copied, not what can be served.

**A failed read is no longer silent.** `GetRaw` checked for `badger.ErrKeyNotFound` and returned `nil` for every other error, handing the caller an empty block and no reason. It now returns the error. That was an RPC nuisance before; it decides what a peer receives now.

## What one block costs to serve

Against the real block store, block already hot in badger, so this is the encode and decode rather than the disk read:

| Block | Before | After |
| --- | --- | --- |
| 2.0 KB, 1 txn | 23.1 µs · 37.9 KB · 137 allocs | **2.5 µs · 7.7 KB · 33 allocs** |
| 9.4 KB, 5 txns | 56.3 µs · 190 KB · 480 allocs | **5.2 µs · 30.0 KB · 33 allocs** |
| 36.9 KB, 20 txns | 185.6 µs · 810 KB · 1,760 allocs | **22.9 µs · 124 KB · 33 allocs** |

Roughly 8–11× on time, and the allocation count stops tracking the number of transactions. Serving 68,000 blocks to one peer used to allocate around 9.3M objects at the mainnet mean block size; it now allocates 2.2M.

## Frames on the wire

| Response | Before | After |
| --- | --- | --- |
| By height, block fits the buffer | 7 | **1** |
| By height, any block at all | 7 | **≤ 3** |
| By hash, block fits the buffer | 5 | **1** |
| Not found, with our best height | 2 | **1** |

## Tests

| Broken | Result |
| --- | --- |
| `GetRaw` decodes and re-encodes again | `TestBlockStore_GetRawIsWhatWasStored` fails |
| `GetRaw` stops reporting the height | `TestBlockStore_GetRawIsWhatWasStored` and `TestBlockStore_GetRawByHeightAgreesWithGetRaw` fail |
| a read failure goes back to returning nil | `TestBlockStore_GetRawReportsAReadFailure` fails |
| the buffer is removed or shrunk | the three frame-count tests fail |
| the flush error is dropped | `TestWriteBlockReportsAFailedSend` fails |
| the frame order or contents change | `TestServeBlockOverTheWire` fails, over mocknet, against the real client parser |
| the not-found reply loses the best height | `TestServeBlockOverTheWire/by_height,_not_found` fails |

The verbatim test is the one worth explaining. Byte equality cannot tell "returned what was stored" from "decoded it and encoded it again", because the two agree on every block the store wrote. So it appends a trailing byte to the stored value: `DecodeBlock` stops after the last transaction and ignores what follows, so a round trip drops it and returns a block one byte short.

Each row above was run as an actual mutation against the suite. Eight of nine were caught. The ninth is below.

## Not covered

- `mets.ServedBlock` takes the height from `req.Height`, and no test can see a metric attribute without standing up an OTel reader. A wrong value there would have to be written deliberately rather than drift in.
- `blocks.retrieved.count` and `blocks.retrieved.bytes` now carry a real height where the raw path reported `-1`. Anyone reading those by height sees heights rather than one `-1` bucket.

`./node/...` passes with `-tags=ext_test,pglive` against the CI Postgres image, as does the rest of the module. `go build ./...`, `gofmt` and `go vet` are clean.

## Summary by CodeRabbit

- **Performance**
  - Improved block retrieval and serving efficiency by reducing response frames and write operations, especially for larger blocks.

- **Reliability**
  - Block lookups now preserve the exact stored block data.
  - Storage read errors are reported consistently instead of being silently ignored.

- **Protocol Updates**
  - Block responses now include the requested block hash and current best height.
  - Not-found responses continue to provide the best known height.


MicBun:
@holdex pr submit-time 4h

